PR #10, opened 2019-10-10
fix connected() & add port range check &  add @nonnull annotation & set maxInboundMessageSize

youny626:


JinHai-CN:
Add change log

JinHai-CN:
#10

youny626:
#14

PR #17, opened 2019-10-11
change IndexParam in DescribeIndexResponse to Index

youny626:
#18


JinHai-CN:
Please update changlog

PR #121, opened 2020-07-13
Determine the size of the array

dddddai:
1. Determine the size of the array to avoid expansion and memory waste

2. Change long to int.The size of the ArrayList cannot exceed Integer.MAX_VALUE


sahuang:
Thanks for contributing to Milvus!

Issue #116, opened 2020-07-08
[Feature] Stub should allow users to set Deadline


hdoer:
Java API does not allow users to set the global default Deadline and each requested Deadline. So if I send a request to MilvusServer, I can only wait. This is not the usual practice.

I want to add some API to allow users to do this.

First, create a super class called RequestParam. It contains some field like collectionName, readTimeout and so on.
Second, let SearchParam inherit and remove collectionName from SearchParam.
Last, Determine whether the search method should add a deadline to the stub.

Similarly, other methods do the same.

sahuang:
We plan to add timeout (deadlines) to each API in the future. It will allow you to set a deadline from execution of a method (e.g. 30 seconds). If no results returned from gRPC in 30s, `DEADLINE_EXCEEDED` will be thrown. There will be a default timeout (probably 24 hours).

This might be supported with Milvus v0.11.0 around mid-August. Let me know if this is exactly what your feature request asks about.

sahuang:
Please refer to [this issue](https://github.com/milvus-io/milvus/issues/2713) for more information. We do not plan to add deadline itself into SDK.

PR #128, opened 2020-08-24
GRPC timeout support

thirstycrow:
```java
// A 'root' MilvusClient instance
MilvusClient client = new MilvusGrpcClient(new ConnectParam.Builder().withHost(host).withPort(port).build());

// A MilvusClient instance with timeout support
MilvusClient timeoutClient = client.withTimeout(50, TimeUnit.MILLISECONDS);

// Close either of instance also close the other
timeoutClient.close();
```

### DIFF
--- a/src/main/java/io/milvus/client/MilvusGrpcClient.java
+++ b/src/main/java/io/milvus/client/MilvusGrpcClient.java
@@ -107,14 +107,22 @@ public class MilvusGrpcClient extends AbstractMilvusGrpcClient {
     channel.shutdown();
     long now = System.nanoTime();
     long deadline = now + TimeUnit.SECONDS.toNanos(maxWaitSeconds);
-    while (now < deadline && !channel.isTerminated()) {
-      try {
-        channel.awaitTermination(deadline - now, TimeUnit.NANOSECONDS);
-      } catch (InterruptedException ex) {
+    boolean interrupted = false;
+    try {
+      while (now < deadline && !channel.isTerminated()) {
+        try {
+          channel.awaitTermination(deadline - now, TimeUnit.NANOSECONDS);
+        } catch (InterruptedException ex) {
+          interrupted = true;
+        }
       }
-    }
-    if (!channel.isTerminated()) {
-      channel.shutdownNow();
+      if (!channel.isTerminated()) {
+        channel.shutdownNow();
+      }
+    } finally {
+      if (interrupted) {
+        Thread.currentThread().interrupt();
+      }
     }
   }
 
--- a/src/main/java/io/milvus/client/MilvusGrpcClient.java
+++ b/src/main/java/io/milvus/client/MilvusGrpcClient.java
@@ -24,10 +24,16 @@ import com.google.common.util.concurrent.Futures;
 import com.google.common.util.concurrent.ListenableFuture;
 import com.google.common.util.concurrent.MoreExecutors;
 import com.google.protobuf.ByteString;
-import io.grpc.ConnectivityState;
+import io.grpc.CallOptions;
+import io.grpc.Channel;
+import io.grpc.ClientCall;
+import io.grpc.ClientInterceptor;
 import io.grpc.ManagedChannel;
 import io.grpc.ManagedChannelBuilder;
+import io.grpc.MethodDescriptor;
 import io.grpc.StatusRuntimeException;
+import io.milvus.client.exception.InitializationException;
+import io.milvus.client.exception.UnsupportedServerVersion;
 import io.milvus.grpc.*;
 import java.nio.Buffer;
 import java.nio.ByteBuffer;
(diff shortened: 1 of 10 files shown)


sahuang:
Hi @thirstycrow 

Thanks for raising this pull request and contributing to Milvus! There are a couple of things that I'd like to discuss with you.

- It seems that `MilvusClient instance with timeout support` will set a deadline for each method call. There's an issue (#116) that discussed the necessity of adding a timeout mechanism for Java SDK. In short, we think it is more flexible to let users self-define deadlines rather than implementing them in SDK. Do let me know your thoughts.

- I noticed that you changed the grpc client test to using docker images. The client test is actually used for internal purposes (i.e. to test the latest features and compatibility with Milvus master branch). Therefore, a local Milvus server is needed to run the tests. Current test codes only rely on published docker images and do not suit our needs.

- Next month we plan to release Milvus 0.11.0 which supports a lot more features, including scalar field filtering and DSL queries. Therefore, Java SDK will see a dramatic change. I guess the safest approach is to create a separate version of Java SDK for Milvus v0.10 that supports timeout if we feel it's needed.

thirstycrow:
> * It seems that `MilvusClient instance with timeout support` will set a deadline for each method call. There's an issue (#116) that discussed the necessity of adding a timeout mechanism for Java SDK. In short, we think it is more flexible to let users self-define deadlines rather than implementing them in SDK. Do let me know your thoughts.

It's a compromise on lack of default argument support in Java. By this way, users are given the flexibility to set a timeout of not, without changes to the existing API. Multiple client instances will have to be created to support different timeouts, but it's still easier for the user, than to support timeout for individual API methods. Besides, GRPC Deadlines are propagated to the server side, and has a potential to improve responsiveness under load if respected by the server.

> * I noticed that you changed the grpc client test to using docker images. The client test is actually used for internal purposes (i.e. to test the latest features and compatibility with Milvus master branch). Therefore, a local Milvus server is needed to run the tests. Current test codes only rely on published docker images and do not suit our needs.

I will refactor the test code to support them both later.

> * Next month we plan to release Milvus 0.11.0 which supports a lot more features, including scalar field filtering and DSL queries. Therefore, Java SDK will see a dramatic change. I guess the safest approach is to create a separate version of Java SDK for Milvus v0.10 that supports timeout if we feel it's needed.

Yes, please make it this way.

thirstycrow:
Now, tests are running without containers by default, and containers can be enabled by setting the `with-containers` system property to `true`

thirstycrow:
Now, tests are running without containers by default, and containers can be enabled by setting the with-containers system property to true

sahuang:
Sounds cool, is it possible to set the with-containers system property in code? I added the property in setUp() but does not work.

thirstycrow:
It is possible but a little bit verbose.

```java
class TestContainerSwitch {
  public static final String KEY = "user.name";
  public static final String FALSE = "^$";
  public static final String TRUE = ".*";
  
  public static final String ENABLED = TRUE;
}

@Testcontainers
@EnabledIfSystemProperty(named = TestContainerSwitch.KEY, matches = TestContainerSwitch.ENABLED)
class ContainerMilvusClientTest extends MilvusClientTest {
}

@DisabledIfSystemProperty(named = TestContainerSwitch.KEY, matches = TestContainerSwitch.ENABLED)
class MilvusClientTest {
}
```

sahuang:
In `examples/MilvusClientExample.java`:

```
// Create Milvus client
MilvusClient client = new MilvusGrpcClient();

// Connect to Milvus server
ConnectParam connectParam = new ConnectParam.Builder().withHost(host).withPort(port).build();
try {
  Response connectResponse = client.connect(connectParam);
} catch (ConnectFailedException e) {
  System.out.println("Failed to connect to Milvus server: " + e.toString());
  throw e;
}
```

Should be changed to 

```
// Connect to Milvus server
ConnectParam connectParam = new ConnectParam.Builder().withHost(host).withPort(port).build();
MilvusClient client = new MilvusGrpcClient(connectParam);
```

And `Disconnect from Milvus server` in the last few lines should be replaced by `client.close();`

Another question, it seems like at this moment `ConnectFailedException` will never be thrown because creating MilvusClient only throws `InitializationException` now. Any difference in between (why add InitializationException instead of using the previous one?)

src/main/java/io/milvus/client/MilvusGrpcClient.java
```diff
@@ -40,106 +46,145 @@
...
-    try {
+  @Override
+  public void close(long maxWaitSeconds) {
+    channel.shutdown();
+    long now = System.nanoTime();
+    long deadline = now + TimeUnit.SECONDS.toNanos(maxWaitSeconds);
+    while (now < deadline && !channel.isTerminated()) {
+      try {
+        channel.awaitTermination(deadline - now, TimeUnit.NANOSECONDS);
+      } catch (InterruptedException ex) {
```
sahuang: Need to log exception if possible

thirstycrow: Interruptions are pushed off instead of swallowed in the latest code, inspired by guava's `Futures.getUninterruptibly`

PR #129, opened 2020-08-26
Support GRPC name resolver and load balancing

thirstycrow:
Support GRPC name resolver and load balancing, which are necessary for HA. Version check should be moved to somewhere   after a connection is created, but I did not find a way to do it with grpc-java. So I propose an alternative by sending the client version via GRPC header, and version check would then be performed by the server.

### DIFF
--- a/src/main/java/io/milvus/client/ConnectParam.java
+++ b/src/main/java/io/milvus/client/ConnectParam.java
@@ -19,13 +19,17 @@
 
 package io.milvus.client;
 
+import io.grpc.ManagedChannelBuilder;
+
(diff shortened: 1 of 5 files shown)


thirstycrow:
@sahuang Please take some time to review this PR.

sahuang:
Thanks for the contribution! I didn't realize you would add more pull requests after the last one, so v0.8.5 was actually released a few days ago. I guess either I will push a v0.8.6 SDK if more contributions are made by community, or I will try to pick this into v0.9.0 for Milvus v0.11.0.

PR #130, opened 2020-09-12
Fix shutdown of channel

dddddai:
Remove redundant loop and simplify the code

### DIFF
--- a/src/main/java/io/milvus/client/MilvusGrpcClient.java
+++ b/src/main/java/io/milvus/client/MilvusGrpcClient.java
@@ -110,24 +110,13 @@ public class MilvusGrpcClient extends AbstractMilvusGrpcClient {
   @Override
   public void close(long maxWaitSeconds) {
     channel.shutdown();
-    long now = System.nanoTime();
-    long deadline = now + TimeUnit.SECONDS.toNanos(maxWaitSeconds);
-    boolean interrupted = false;
     try {
-      while (now < deadline && !channel.isTerminated()) {
-        try {
-          channel.awaitTermination(deadline - now, TimeUnit.NANOSECONDS);
-        } catch (InterruptedException ex) {
-          interrupted = true;
-        }
-      }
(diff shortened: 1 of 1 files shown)


sahuang:
Hi, thanks for the pull request. I think the change you proposed is similar to what I did before v0.8.5. @thirstycrow did a few changes to the timeout and shutdown of Milvus server and used this method. The reason can be seen from [this commit message](https://github.com/milvus-io/milvus-sdk-java/pull/128/commits/0c36a9a3634f144ad2e5123fdde59d8ef5e464ff).

dddddai:
> Hi, thanks for the pull request. I think the change you proposed is similar to what I did before v0.8.5. @thirstycrow did a few changes to the timeout and shutdown of Milvus server and used this method. The reason can be seen from [this commit message](https://github.com/milvus-io/milvus-sdk-java/pull/128/commits/0c36a9a3634f144ad2e5123fdde59d8ef5e464ff).

Thanks for your review. I did not changed the logic but just simplify the code.

And the while loop doesn't make sense to me.
1.`now` never has never changed in the loop.
2.`awaitTermination` waits for a timeout, there is no need to use a while loop.

https://github.com/milvus-io/milvus-sdk-java/blob/master/src/main/java/io/milvus/client/MilvusGrpcClient.java#L111-L132

sahuang:
Thanks for the explanation, I think your point is valid and should be correct.

@thirstycrow do you mind explaining your rationale? I also feel that the while loop/now is not necessary.

thirstycrow:
My intention was to put off any interruption until the shutdown complete normally or timed out, because I am not sure whether it is possible for the method to receive an unintended interruption. The current timestamp (`now`) should have been updated at the end of the while block.

sahuang:
The simplified version seems alright to me, I don't feel it's really necessary to consider unintended interruption when it comes to shutting down the channel and server.

PR #134, opened 2020-10-12
Simplify the client code

thirstycrow:
1. API methods now return the actual data that was wrapped in the various response classes, and exceptional cases are expressed by throwing exceptions. Now it's possible for the users to handle exceptions in a global exception handler instead of checking response status on every call site.

2. Data structures for API params / return values are rewritten in the way wrapping the builder of the corresponding gRPC data structure. Code for type conversions is not required any more.

3. The blocking API method now just invoke the corresponding non-blocking version if there is one and wait the future to complete. So it's not necessary to write an individual test method for the non-blocking one.
 
4. Enums for index types and metrics types are reintroduced. It's not necessary to use string for API params just because their wire format changed from int to string. However some binary index types may need to be complemented.

5. Type for binary data is changed back to ByteBuffer from List<Byte>. The latter is bad in performance.

6. Status logs for individual requests are removed. My understanding is that, because exceptional cases were expressed by response statuses, when user forget to check the status, the application may behave uncertainly, and we need the log to diagnose. Since the API methods now throw concise exceptions, request log is not a necessity.

7. Log4j is used only for test. A library should not decide for the user which logging framework to use.

8. Dependencies are minimized.


sahuang:
> 4. Enums for index types and metrics types are reintroduced. It's not necessary to use string for API params just because their wire format changed from int to string. However some binary index types may need to be complemented.

Can you elaborate more on "binary index types may need to be complemented"? It is indeed more convenient to use enums in this scenario.

> 5. Type for binary data is changed back to ByteBuffer from List. The latter is bad in performance.

That's true, I did not find a way to convert from JSON strings back to ByteBuffer. Your solution seems to solve it well.

> 6. Status logs for individual requests are removed. My understanding is that, because exceptional cases were expressed by response statuses, when user forget to check the status, the application may behave uncertainly, and we need the log to diagnose. Since the API methods now throw concise exceptions, request log is not a necessity.

Logs were used for various concerns. As you mentioned, one purpose is to log exceptional cases so users can locate their program errors. We design the logs (especially the `INFO` part) to allow users to check whether their operations are as expected and whether the returned information is exactly the same as they wish. 

For example, suppose a user creates a collection `XYZ` and their `createCollection` call is successful. However, they have a typo in their create params code. They may not realize it throughout the program because no error is thrown. With the log information, they can inspect their parameters and input arguments to double-check the correctness.

In addition, some users also require logs to analyze other things such as `grep` specific keywords from a bunch of logs. Due to these reasons, I think logs are still useful apart from throwing concise exceptions. It would be nice if the status logs (especially `logInfo`) for each request are added back. They serve as extra information on top of the API calls.

Thanks for the contribution!

sahuang:
By the way, I ran the tests based on current Milvus v0.11.0 and encountered 1 failure in `getCollectionInfo`:

```
org.opentest4j.AssertionFailedError: 
Expected :COLLECTION_NOT_EXISTS
Actual   :UNEXPECTED_ERROR

	at org.junit.jupiter.api.AssertionUtils.fail(AssertionUtils.java:55)
	at org.junit.jupiter.api.AssertionUtils.failNotEqual(AssertionUtils.java:62)
	at org.junit.jupiter.api.AssertEquals.assertEquals(AssertEquals.java:182)
	at org.junit.jupiter.api.AssertEquals.assertEquals(AssertEquals.java:177)
	at org.junit.jupiter.api.Assertions.assertEquals(Assertions.java:1124)
	at io.milvus.client.MilvusClientTest.assertErrorCode(MilvusGrpcClientTest.java:121)
	at io.milvus.client.MilvusClientTest.getCollectionInfo(MilvusGrpcClientTest.java:610)
	at ......
```

thirstycrow:
> Can you elaborate more on "binary index types may need to be complemented"? It is indeed more convenient to use enums in this scenario.

It seems that a "BIN_" prefix have to be prepended to the index type now. I added BIN_IVF_FLAT as a enum value, because it appeared here in the test: https://github.com/milvus-io/milvus-sdk-java/blob/0.9.0/src/test/java/io/milvus/client/MilvusGrpcClientTest.java#L617, but I'm not sure whether there is any other valid index type for binary data.

> In addition, some users also require logs to analyze other things such as `grep` specific keywords from a bunch of logs. Due to these reasons, I think logs are still useful apart from throwing concise exceptions. It would be nice if the status logs (especially `logInfo`) for each request are added back. They serve as extra information on top of the API calls.

Request logging is now re-implemented using gRPC ClientInterceptor.

thirstycrow:
> By the way, I ran the tests based on current Milvus v0.11.0 and encountered 1 failure in `getCollectionInfo`:
> 
> ```
> org.opentest4j.AssertionFailedError: 
> Expected :COLLECTION_NOT_EXISTS
> Actual   :UNEXPECTED_ERROR
> 
> 	at org.junit.jupiter.api.AssertionUtils.fail(AssertionUtils.java:55)
> 	at org.junit.jupiter.api.AssertionUtils.failNotEqual(AssertionUtils.java:62)
> 	at org.junit.jupiter.api.AssertEquals.assertEquals(AssertEquals.java:182)
> 	at org.junit.jupiter.api.AssertEquals.assertEquals(AssertEquals.java:177)
> 	at org.junit.jupiter.api.Assertions.assertEquals(Assertions.java:1124)
> 	at io.milvus.client.MilvusClientTest.assertErrorCode(MilvusGrpcClientTest.java:121)
> 	at io.milvus.client.MilvusClientTest.getCollectionInfo(MilvusGrpcClientTest.java:610)
> 	at ......
> ```

I think it'd be better for the server to return an error code of COLLECTION_NOT_EXISTS in this case

sahuang:
> I think it'd be better for the server to return an error code of COLLECTION_NOT_EXISTS in this case

This will be fixed from server-side. Meanwhile, we are performing server tests with this pull request changes and it should be ready before the official release of Milvus v0.11.0.

thirstycrow:
> This will be fixed from server-side. Meanwhile, we are performing server tests with this pull request changes and it should be ready before the official release of Milvus v0.11.0.

Fixed.

sahuang:
> Fixed.

This was actually fixed in latest 0.11.0 (Server will return `COLLECTION_NOT_EXISTS` instead of `UNEXPECTED_ERROR`).

del-zhenwu:
The complete index type list is [here](https://github.com/milvus-io/milvus/blob/0.11.0/core/src/index/knowhere/knowhere/index/IndexType.cpp). Enum needs to be updated (e.g. missing `BIN_FLAT`).

sahuang:
It seems that `withConnectTimeout` in ConnectParam is not used when creating the channel.

```
channel = ManagedChannelBuilder
        .forTarget(connectParam.getTarget())
        .usePlaintext()
        .maxInboundMessageSize(Integer.MAX_VALUE)
        .defaultLoadBalancingPolicy(connectParam.getDefaultLoadBalancingPolicy())
        .keepAliveTime(connectParam.getKeepAliveTime(TimeUnit.NANOSECONDS), TimeUnit.NANOSECONDS)
        .keepAliveTimeout(connectParam.getKeepAliveTimeout(TimeUnit.NANOSECONDS), TimeUnit.NANOSECONDS)
        .keepAliveWithoutCalls(connectParam.isKeepAliveWithoutCalls())
        .idleTimeout(connectParam.getIdleTimeout(TimeUnit.NANOSECONDS), TimeUnit.NANOSECONDS)
        .build();
```

`keepAliveTimeout` and `idleTimeout` are used. However I checked methods for `ManagedChannelBuilder` and could not find any information about `connectTimeout` usage.

Do you know if `connectTimeout` is not needed in this case? I tested that when using `connectTimeout` in `ConnectParam` it did not take effect at all.

Thanks.

sahuang:
Currently `getCollectionInfo` will return a `CollectionMapping` in format such as

`fields = [{field=float, type=FLOAT, params=null}, {field=int64, type=INT64, params=null}`

But in order to be consistent with server side and other SDKs, this needs to be changed to 

`fields = [{**name**=float, type=FLOAT, params=null}, {**name**=int64, type=INT64, params=null}`

Can you fix it accordingly? Thanks.

sahuang:
Another thing: Currently `getEntityByID` will return a map with key=entity_id and value=field_values. If some ids to search are deleted or do not even exist in Milvus, the resulting map will not contain those ids.

Our intentional behavior for this API is that the result should contain all ids from users. If the id does not exist, value should be an exmpty map. (Which is why in the previous design API returns a list rather than a map. The list order corresponds to input ids.)

thirstycrow:
> I tested that when using `connectTimeout` in `ConnectParam` it did not take effect at all. Do you know what is the correct way of setting connect timeout?

It depends on the transport. For netty, it can be set like this:
```java
((NettyChannelBuilder) channelBuilder)
        .withOption(ChannelOption.CONNECT_TIMEOUT_MILLIS, connectionTimeouMillis)
```

For okhttp, a custom SocketFactory would have to be created:
```java
((OkHttpChannelBuilder) channelBuilder)
        .socketFactory(customeSocketFactory)
```

thirstycrow:
> Currently `getCollectionInfo` will return a `CollectionMapping` in format such as
> 
> `fields = [{field=float, type=FLOAT, params=null}, {field=int64, type=INT64, params=null}`
> 
> But in order to be consistent with server side and other SDKs, this needs to be changed to
> 
> `fields = [{**name**=float, type=FLOAT, params=null}, {**name**=int64, type=INT64, params=null}`
> 
> Can you fix it accordingly? Thanks.

Updated

thirstycrow:
> Our intentional expectation of this API is that the result should contain **all ids from users**. If the id does not exist, value should be an empty map. (Which is why in the previous design the API returns a list rather than a map. The list order corresponds to input ids.)

What's the benefit of using empty map for non-existency? Users would have to do an extra filtering for valid data.

sahuang:
> What's the benefit of using empty map for non-existency? Users would have to do an extra filtering for valid data.

You are right, I guess the current design is better.

src/test/java/io/milvus/client/MilvusGrpcClientTest.java
```diff
@@ -775,128 +556,99 @@ void searchBinary() {
...
-        randomCollectionName);
-
-    List<? extends Map<String, Object>> fields = getCollectionInfoResponse.getCollectionMapping()
-        .get().getFields();
-    for (Map<String, Object> field : fields) {
+    CollectionMapping collectionMapping = client.getCollectionInfo(randomCollectionName);
+    assertEquals(randomCollectionName, collectionMapping.getCollectionName());
+
+    for (Map<String, Object> field : collectionMapping.getFields()) {
       if (field.get("field").equals("float_vec")) {
```
sahuang: `field.get("name")`

thirstycrow: updated.

sahuang:
All server tests have passed in our dev environment. Thanks for your help.

PR #139, opened 2020-10-17
Add a dsl api for building search queries

thirstycrow:
With a predefined schema

```java
public class TestSchema extends Schema {
  public final Int32Field intField = new Int32Field("int32");
  public final Int64Field longField = new Int64Field("int64");
  public final FloatField floatField = new FloatField("float");
  public final DoubleField doubleField = new DoubleField("double");
  public final FloatVectorField floatVectorField = new FloatVectorField("float_vec", 64);
  public final BinaryVectorField binaryVectorField = new BinaryVectorField("binary_vec", 64);
}
```

query dsl can be constructed like this

```java
import io.milvus.client.dsl.Query;
import static io.milvus.client.dsl.Query.*;

Query floatVectorQuery(List<List<Float>> floatVectors) {
    return bool(
        must(
            schema.intField.gt(5).lte(10),
            schema.floatVectorField.query(vectors).param("nprobe", 16).top(1)));
}
```

### DIFF
--- a/src/main/java/io/milvus/client/dsl/VectorQuery.java
+++ b/src/main/java/io/milvus/client/dsl/VectorQuery.java
@@ -1,0 +1,98 @@
+package io.milvus.client.dsl;
+
+import com.google.protobuf.UnsafeByteOperations;
+import io.milvus.client.MetricType;
+import io.milvus.client.SearchParam;
+import io.milvus.grpc.VectorParam;
+import io.milvus.grpc.VectorRecord;
+import io.milvus.grpc.VectorRowRecord;
+import org.json.JSONObject;
+
+import java.nio.ByteBuffer;
+import java.util.List;
+import java.util.stream.Collectors;
+
+public class VectorQuery<T> extends Query {
+  private final Schema.VectorField<T> field;
+  private final List<T> queries;
+  private String placeholder;
+  private int topK = 10;
+  private float boost = 1.0f;
+  private MetricType metricType;
+  private JSONObject params = new JSONObject();
+
+  VectorQuery(Schema.VectorField<T> field, List<T> queries) {
+    this.field = field;
+    this.queries = queries;
+    this.placeholder = field.name;
+    this.metricType = field instanceof Schema.FloatVectorField ? MetricType.L2 : MetricType.JACCARD;
+  }
+
+  public VectorQuery<T> placeholder(String placeholder) {
+    this.placeholder = placeholder;
+    return this;
+  }
+
+  public VectorQuery<T> top(int topK) {
+    this.topK = topK;
+    return this;
+  }
+
+  public VectorQuery<T> boost(float value) {
+    this.boost = value;
+    return this;
+  }
+
+  public VectorQuery<T> metricType(MetricType metricType) {
+    this.metricType = metricType;
+    return this;
+  }
+
+  public VectorQuery<T> param(String key, Object value) {
+    params.put(key, value);
+    return this;
+  }
+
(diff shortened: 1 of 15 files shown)


sahuang:
Currently "eq" and "ne" operations are not supported by Milvus range query, and I am not sure if they will be added in v0.11.1. I have submitted a feature request https://github.com/milvus-io/milvus/issues/4017, and hopefully I can know the result soon.

Meanwhile, I will do more testing to see if the DSL API works fine. Thanks for your contribution.

sahuang:
The testFloatVectorQuery seems to generate the wrong DSL query from builder.

The dsl from it is `{"bool":{"must":[{"vector":"float_vec"}]}}`

I think the intentional query is something like `{"bool":{"must":[{"vector":{"float_vec":{"topk": 1, "query": [...], "type": "float"}}}]}}`

The problem seems to be that `query` in `service.buildSearchParam(query)` does not take any effect.

thirstycrow:
> I think the intentional query is something like `{"bool":{"must":[{"vector":{"float_vec":{"topk": 1, "query": [...], "type": "float"}}}]}}`

`{"bool":{"must":[{"vector":{"float_vec":{"topk": 1, "query": [...], "type": "float"}}}]}}` is preprocessed in `SearchParam.setDsl` to extract the vector query part, and `{"bool":{"must":[{"vector":{"placeholder"}]}}` is what's actually set to the `dsl` field of gRPC SearchParam. 

`MilvusService.buildSearchParam` returns `SearchParam` instead of JSON, so the query data will not have to be transformed into JSON and parsed back to what it already is.

I did not find any constraint in the design document, on the number of vector fields which can be queried in a single query. So I chose the vector field name as the default placeholder name in case where multiple vector fields involved. The placeholder can be set explicitly via `VectorQuery.placeholder`.

sahuang:
Thanks for your explanation.

> I did not find any constraint in the design document, on the number of vector fields which can be queried in a single query. 

Currently, Milvus only supports a single vector in the collection for insertion and query. I realized that `testFloatVectorQuery` failed because the Milvus server does not check for vector field count when creating a collection. The server confuses `binary_vec' with `float_vec` field names and raises an error. [This](https://github.com/milvus-io/milvus/issues/4029) will hopefully be fixed soon.

The placeholder name does not matter and the current design is okay.

sahuang:
Hi, just wanna know your current progress on this pull request? 0.9.1 was released yesterday with some bug fix, so this might go into 0.9.2 for Milvus v0.11.1.

thirstycrow:
No more update except the removal of EQ and NE from query range. Is there something else need to be done?

sahuang:
> No more update except the removal of EQ and NE from query range. Is there something else need to be done?

Can you refer to https://github.com/thirstycrow/milvus-sdk-java/pull/2 for more information?

thirstycrow:
This PR is now rebased to branch 0.9.2

sahuang:
Can you open a separate PR into 0.9.2?

src/main/java/io/milvus/client/dsl/VectorQuery.java
```diff
@@ -0,0 +1,98 @@
...
+          .build();
+    }
+
+    VectorParam vectorParam = VectorParam.newBuilder()
+        .setJson(new JSONObject()
+            .put(placeholder, new JSONObject()
+                .put(field.name, new JSONObject()
+                    .put("topk", topK)
+                    .put("metric_type", metricType.name())
+                    .put("boost", boost)
```
sahuang, line 84: `boost` is not available now.

Edit: Does not matter atm.

sahuang (approved):

PR #191, opened 2021-08-28
framework && simple connection

trovwu:


trovwu (commented):
init

src/main/java/io/milvus/param/R.java
```diff
@@ -0,0 +1,70 @@
...
+    public T getData() {
+        return data;
+    }
+
+    public void setData(T data) {
+        this.data = data;
+    }
+
+    public static <T> R<T> failed(){
+        R<T> r = new R<>();
```
xiaofan-luan, line 35: If failed, should we pass in the exception msg here?

PR #196, opened 2021-08-31
Add development.md for contributor to setup environment(#195)

xiaofan-luan:
see also #195
Signed-off-by: xiaofan-luan <xiaofan.luan@zilliz.com>

PR #197, opened 2021-08-31
Master

trovwu:
graft version of insert, search ,flush and delete function's implement.  next step is to consider and optimize more details.


xiaofan-luan:
why is this pull request named master? see https://github.com/milvus-io/milvus-sdk-java/pull/196 for detailed github flow

PR #198, opened 2021-09-01
graft version of insert, search ,flush and delete function's implement. next step is to consider and optimize more details.

trovwu:


trovwu:
add query and caldistance

src/main/java/io/milvus/client/AbstractMilvusGrpcClient.java
```diff
@@ -37,7 +38,325 @@
...
+        List<ByteString> byteStrings = vectors.stream().map(vector -> {
+            ByteBuffer allocate1 = ByteBuffer.allocate(16);
+            vector.forEach(allocate1::putFloat);
+            byte[] array = allocate1.array();
+            return ByteString.copyFrom(array);
+        }).collect(Collectors.toList());
+
+
+        PlaceholderValue.Builder pldBuilder = PlaceholderValue.newBuilder()
+                .setTag("$0")
```
xiaofan-luan: Where is this $0 come from?

src/main/java/io/milvus/param/DeleteParam.java
```diff
@@ -0,0 +1,60 @@
...
+
+    public static class Builder {
+        private String dbName = "";
+        private String collectionName;
+        private String partitionName ;
+
+        private Builder() {
+        }
+
+        public static Builder nweBuilder() {
```
xiaofan-luan: newBuilder

src/main/java/io/milvus/param/DeleteParam.java
```diff
@@ -0,0 +1,60 @@
+package io.milvus.param;
+
+import javax.annotation.Nonnull;
+
+public class DeleteParam {
+    private final String dbName;
```
xiaofan-luan: We don't have primary field for delete?

src/main/java/io/milvus/client/AbstractMilvusGrpcClient.java
```diff
@@ -37,7 +38,325 @@
...
+        String partitionName = insertParam.getPartitionName();
+        int fieldNum = insertParam.getFieldNum();
+        List<String> fieldNames = insertParam.getFieldNames();
+        List<DataType> dataTypes = insertParam.getDataTypes();
+        List<List<?>> fieldValues = insertParam.getFieldValues();
+        Integer filedNameSize = Optional.ofNullable(fieldNames).map(List::size).orElse(0);
+        Integer dTypeSize = Optional.ofNullable(dataTypes).map(List::size).orElse(0);
+        int fieldValueSize = Optional.ofNullable(fieldValues).map(List::size).orElse(0);
+        if (fieldNum != fieldValueSize || fieldNum != filedNameSize || fieldNum != dTypeSize) {
+            throw new ParamException("size is illegal");
```
xiaofan-luan: print more detailed information about what field is illegal, that helps user to do self debug

src/main/java/io/milvus/client/AbstractMilvusGrpcClient.java
```diff
@@ -37,7 +38,325 @@
...
+        SearchRequest.Builder builder = SearchRequest.newBuilder()
+                .setDbName(searchParam.getDbName())
+                .setCollectionName(searchParam.getCollectionName());
+        if (!searchParam.getPartitionNames().isEmpty()) {
+            searchParam.getPartitionNames().forEach(builder::addPartitionNames);
+        }
+
+        List<List<Float>> vectors = searchParam.getVectors();
+        List<ByteString> byteStrings = vectors.stream().map(vector -> {
+            ByteBuffer allocate1 = ByteBuffer.allocate(16);
```
xiaofan-luan: allocate a valid size based on vector size, this could avoid extra copy

src/main/java/io/milvus/client/AbstractMilvusGrpcClient.java
```diff
@@ -37,7 +38,325 @@
...
+        if (vectorDataType.contains(dataType)) {
+            if (dataType == DataType.FloatVector) {
+                List<Float> floats = new ArrayList<>();
+                // 每个object是个list
+                for (Object object : objects) {
+                    if (object instanceof List) {
+                        List list = (List) object;
+                        list.forEach(o -> floats.add((Float) o));
+                    } else {
+                        throw new ParamException("参数有问题");
```
xiaofan-luan: remove chinese error message

src/main/java/io/milvus/client/AbstractMilvusGrpcClient.java
trovwu: this value come from pymilvus client.   pymilvus.client.prepare.py line 602

src/main/java/io/milvus/param/DeleteParam.java
trovwu: modified in new commit

src/main/java/io/milvus/param/DeleteParam.java
trovwu: proto doesn't have primary filed,but proto has value expr. add value expr in new commit

src/main/java/io/milvus/client/AbstractMilvusGrpcClient.java
trovwu: this need DDL. this exception will add int next step. todo

src/main/java/io/milvus/client/AbstractMilvusGrpcClient.java
trovwu: this problem modified in new commit

src/main/java/io/milvus/client/AbstractMilvusGrpcClient.java
trovwu: this problem has modified in new commit

src/main/java/io/milvus/client/AbstractMilvusGrpcClient.java
xiaofan-luan: If this is a magic number can we put it into some constants?

trovwu: get it

xiaofan-luan:
name of the patch should be draft of xxx

PR #201, opened 2021-09-08
zhaoweilong control

Javajava1:
Three methods about control writed by zhao-weilong.

### DIFF
--- a/src/main/java/io/milvus/client/MilvusClient.java
+++ b/src/main/java/io/milvus/client/MilvusClient.java
@@ -19,206 +19,25 @@
 
 package io.milvus.client;
 
-import io.milvus.grpc.DescribeCollectionResponse;
-import io.milvus.grpc.DescribeIndexResponse;
-import io.milvus.grpc.GetCollectionStatisticsResponse;
-import io.milvus.grpc.GetIndexBuildProgressResponse;
-import io.milvus.grpc.GetIndexStateResponse;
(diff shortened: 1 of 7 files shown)


src/main/java/io/milvus/client/MilvusClient.java
```diff
@@ -19,206 +19,25 @@
...
-    R<GetIndexStateResponse> getIndexState(GetIndexStateParam requestParam);
-
-    /**
-     * Show index building progress.
-     *
-     * @param requestParam {@link GetIndexBuildProgressParam}
-     * @return {status:result code,data:GetIndexStateResponse{status,indexed_rows}}
-     */
-    R<GetIndexBuildProgressResponse> getIndexBuildProgress(GetIndexBuildProgressParam requestParam);
+  R<Boolean> hasCollection(String collectionName);
```
xiaofan-luan, line 39: Seems that there is a conflict with previous patch, need to rebase

src/main/java/io/milvus/client/MilvusClient.java
```diff
@@ -19,206 +19,25 @@
...
+  R<GetPersistentSegmentInfoResponse> getPersistentSegmentInfo(GetPersistentSegmentInfoParam getPersistentSegmentInfoParam);
```
xiaofan-luan, line 40: Please give some comments as previous call

xiaofan-luan:
Already merged by other branch, close for now

PR #208, opened 2021-09-15
Add Docker based test for java client

xiaofan-luan:
1) Add Docker based test for create and drop collection
2) Fix some error logic in create/drop collection path

Fix : #209
Signed-off-by: xiaofan-luan <xiaofan.luan@zilliz.com>

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,43 @@
+version: '3.5'
+
+services:
+  etcd:
+    container_name: milvus-test-etcd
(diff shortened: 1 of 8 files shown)


src/main/java/io/milvus/exception/ExceptionUtils.java
```diff
@@ -0,0 +1,938 @@
+  /*
```
yhmo: This file seems come from org.apache.commons.lang3.exception.ExceptionUtils. Is it necessary to duplicate it here?

PR #214, opened 2021-10-29
Add mock server for unittest

yhmo:

### DIFF
--- a/src/main/java/io/milvus/param/R.java
+++ b/src/main/java/io/milvus/param/R.java
@@ -37,6 +37,8 @@ public class R<T> {
         this.exception = exception;
     }
 
+    public String getMessage() { return exception.getMessage(); }
+
     public Integer getStatus() {
         return status;
     }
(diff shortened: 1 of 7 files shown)


src/main/java/io/milvus/param/collection/CreateCollectionParam.java
```diff
@@ -58,9 +58,9 @@ public String getDescription() {
 
 
     public static final class Builder {
-        private String collectionName;
-        private int shardsNum;
-        private String description;
+        private String collectionName = "";
```
xiaofan-luan: The default collectionName set to "" is not a good idea, user may create a collectionName with empty string. 
Better to throw exception if collection name is null when call build

src/main/java/io/milvus/param/R.java
```diff
@@ -60,17 +62,17 @@ public void setData(T data) {
         return r;
     }
 
-    public static <T> R<T> failed(ErrorCode errorCode) {
+    public static <T> R<T> failed(ErrorCode errorCode, String msg) {
```
xiaofan-luan, line 65: Why do we directly call failed(ErrorCode errorCode, Exception e)?

yhmo: No need to always construct an Exception object, easy to call.

src/main/java/io/milvus/param/collection/CreateCollectionParam.java
yhmo: make sense

xiaofan-luan (approved):

PR #220, opened 2021-11-08
Implement load and flush

yhmo:
1. sync load logic
2. sync flush logic
3. avoid compiler warning
4. implement getMetrics/getPersistentSegmentInfo/getQuerySegmentInfo
5. add unittest cases
6. improve usability of some param classes
7. add toString() method for each param class
8. refine error message and exception handling
9. use lambok to simplify code
10. add response wrapper for usability
11. fix search error
12. setup CI

### DIFF
--- a/.github/mergify.yml
+++ b/.github/mergify.yml
@@ -1,0 +1,18 @@
+pull_request_rules:
+  - name: Test passed for code changed
+    conditions:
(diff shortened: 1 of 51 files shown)


yhmo:
@Mergifyio refresh

PR #221, opened 2021-11-12
Implement load and flush

yhmo:

1. sync load logic
2. sync flush logic
3. avoid compiler warning
4. implement getMetrics/getPersistentSegmentInfo/getQuerySegmentInfo
5. add unittest cases
6. improve usability of some param classes
7. add toString() method for each param class
8. refine error message and exception handling
9. use lambok to simplify code
10. add response wrapper for usability
11. fix search error
12. setup CI

### DIFF
--- a/src/main/java/io/milvus/client/AbstractMilvusGrpcClient.java
+++ b/src/main/java/io/milvus/client/AbstractMilvusGrpcClient.java
@@ -22,28 +22,32 @@ package io.milvus.client;
 import com.google.protobuf.ByteString;
 import io.grpc.StatusRuntimeException;
 import io.milvus.exception.ClientNotConnectedException;
+import io.milvus.exception.IllegalResponseException;
 import io.milvus.exception.ParamException;
 import io.milvus.grpc.*;
 import io.milvus.param.Constant;
-import io.milvus.param.ParamUtils;
 import io.milvus.param.R;
 import io.milvus.param.RpcStatus;
 import io.milvus.param.alias.AlterAliasParam;
 import io.milvus.param.alias.CreateAliasParam;
 import io.milvus.param.alias.DropAliasParam;
 import io.milvus.param.collection.*;
+import io.milvus.param.control.GetMetricsParam;
+import io.milvus.param.control.GetPersistentSegmentInfoParam;
+import io.milvus.param.control.GetQuerySegmentInfoParam;
 import io.milvus.param.dml.*;
 import io.milvus.param.index.*;
 import io.milvus.param.partition.*;
+import lombok.NonNull;
 import org.apache.commons.collections4.CollectionUtils;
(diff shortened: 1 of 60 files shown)


src/main/java/io/milvus/client/AbstractMilvusGrpcClient.java
```diff
@@ -417,704 +164,1389 @@ private FieldData genFieldData(String fieldName, DataType dataType, List<?> obje
...
+        if (partitionNames == null || partitionNames.isEmpty()) {
+            ShowCollectionsRequest showCollectionRequest = ShowCollectionsRequest.newBuilder()
+                    .addCollectionNames(collectionName)
+                    .setType(ShowType.InMemory)
+                    .build();
+
+            // Use showCollection() to check loading percentages of the collection.
+            // If the inMemory percentage is 100, that means the collection has finished loading.
+            // Otherwise, this thread will sleep a small interval and check again.
+            while (true) {
```
xiaofan-luan, line 180: what about timeout? do we need a timeout for waiting to load?

src/main/java/io/milvus/client/AbstractMilvusGrpcClient.java
```diff
@@ -417,704 +164,1389 @@ private FieldData genFieldData(String fieldName, DataType dataType, List<?> obje
...
 
-        Status response;
-        try {
-            response = blockingStub().createCollection(createCollectionRequest);
+            Status response = blockingStub().createCollection(createCollectionRequest);
 
             if (response.getErrorCode() == ErrorCode.Success) {
-                logInfo("Created collection " + requestParam.getCollectionName() + " successfully!\n{}",
-                        requestParam.toString());
+                logInfo("CreateCollectionRequest successfully!");
```
xiaofan-luan: need at least the collection name here? same as the other ddl requests.

src/main/java/io/milvus/client/AbstractMilvusGrpcClient.java
xiaofan-luan: same as flush, we should have a time limit of blocking wait, for example 5 mins.

yhmo: Sounds good. I will add a parameter to control the time out.

src/main/java/io/milvus/client/AbstractMilvusGrpcClient.java
yhmo: In each api method I add a line "logInfo(requestParam.toString());", to print out the input prameters, including the collection name.

yhmo: done

xiaofan-luan (approved):
lgtm

PR #270, opened 2022-03-22
multi server failover huazaizai0322@163.com

huazaizai0322:


yhmo:
DCO not passed. Please add your email address to the commit.

yhmo:
@huazaizai0322 
Ok, I read the code change, very impressive. This failover machinery forwards DML requests to all milvus services, and check heartbeat for each service by 10 seconds interval, once the master service is down, the server monitor will choose another available service as master.
One concern is: milvus has no ability to sync data between milvus services. If a service is down, and missed some DML requests, query may failed or result is wrong when this service is elected as master again.

huazaizai0322:
> 

Hi @yhmo Thanks for your review. In my production, I publish a new collection(create collection, insert vector, create index and load collection) every day. I have a embedding collection table. it contain a 'version' field, you can understand version as milvus collection name. When successful publish vector into all cluster, version will be changed. Then search with the new version in milvus-sdk-java. So if insert data failed, version will be not changed.

yhmo:
@huazaizai0322 
Still one problem left, the DCO check. We require each pull request is signed off by a user email.
https://github.com/milvus-io/milvus-sdk-java/pull/270/checks?check_run_id=5648367922
`Signed-off-by: Author Name <authoremail@example.com>`

PR #324, opened 2022-07-08
Feature/2.1.0 beta

yelusion2:


.gitmodules
```diff
@@ -1,3 +1,6 @@
 [submodule "src/main/milvus-proto"]
 	path = src/main/milvus-proto
 	url = https://github.com/milvus-io/milvus-proto.git
+[submodule "milvus-proto"]
```
yhmo, line 4: Why here need another submodule? Duplicate?

PR #325, opened 2022-07-08
xun.huang@zilliz.com 1.milvus client impl https 2.delete account password log 3.query and search add consistency level.

yelusion2:
1.milvus client impl https 
2.delete account password log 
3.query and search add consistency level.

Signed-off-by: xun.huang@zilliz.com <xun.huang@zilliz.com>

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -25,7 +25,7 @@
 
     <groupId>io.milvus</groupId>
     <artifactId>milvus-sdk-java</artifactId>
-    <version>2.0.5</version>
+    <version>2.1.0-beta1</version>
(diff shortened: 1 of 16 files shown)


pom.xml
```diff
@@ -55,6 +55,8 @@
         <url>https://github.com/milvus-io/milvus-sdk-java</url>
     </scm>
 
+
+    <!-- 使用分发管理将本项目打成jar包，直接上传到指定服务器 -->
```
yhmo: We should use english to comment

yelusion2: yes, i how delete the unused comment.

PR #361, opened 2022-07-25
rollback CalcDistance method

bestRiven:
Signed-off-by: yu <958988867@qq.com>

rollback CalcDistance method, this method we used online service.


yhmo:
The calcDistance() is not good designed: it use proxy node to compute distance, this will affect the QPS in some conditions.
Currently, we want to do some refacture work to milvus, this function is bad designed, so we decide to remove it.

PR #367, opened 2022-08-10
Fix keepAliveTimeout timeunit error

psc0606:
fix error time unit which cause "io.grpc.netty.shaded.io.netty.channel.ChannelException: setsockopt() failed: Invalid argument"


yhmo:
good catch!

There is a test error in automation. In the `/milvus-sdk-java/src/test/java/io/milvus/client/MilvusServiceClientTest.java`, you need to replace all the ` .withKeepAliveTimeout(keepAliveTimeoutMs, TimeUnit.NANOSECONDS)` with ` .withKeepAliveTimeout(keepAliveTimeoutMs, TimeUnit.MILLISECONDS)`

psc0606:
@yhmo fix all unit test

PR #430, opened 2023-01-19
[cherry-pick] Add refresh parameter to load api

yhmo:


yhmo:
The ci failure is caused by grpc version(incompatible). Will submit another pr to fix.

PR #559, opened 2023-07-11
Support TLS

yhmo:


pom.xml
```diff
@@ -111,9 +111,8 @@
     <dependencies>
         <dependency>
             <groupId>io.grpc</groupId>
-            <artifactId>grpc-netty-shaded</artifactId>
```
yangshuai0711, line 114: What is the reason for the code change? In my environment, the Netty version conflicts with the Netty version used internally by the service. Due to the complex dependencies, it's not possible to resolve this issue. Can we revert to the version with shaded?

PR #584, opened 2023-08-21
Fix bug of highLevel-delete

lentitude2tk:


xiaofan-luan:
can we have some test cases for the API?

lentitude2tk:
> can we have some test cases for the API?

Sure, I will add corresponding test cases for this scene.

PR #644, opened 2023-09-22
[upsert] Fix NPE in case of non empty database name

eolivelli:
This PR solves this error:

```
java.lang.NullPointerException: Cannot invoke "io.milvus.grpc.InsertRequest$Builder.setDbName(String)" because "this.insertBuilder" is null
	at io.milvus.param.ParamUtils$InsertBuilderWrapper.<init>(ParamUtils.java:268)
	at io.milvus.client.AbstractMilvusGrpcClient.upsert(AbstractMilvusGrpcClient.java:1625)
	at ai.langstream.agents.vector.milvus.MilvusWriter$MilvusVectorDatabaseWriter.upsert(MilvusWriter.java:99)
	... 69 more
```


yhmo:
good catch!

PR #724, opened 2023-12-28
chore: logInfo is too high for recording request params

vsanna:
related issue: https://github.com/milvus-io/milvus-sdk-java/issues/725

## Context
- AbstractMilvusGrpcClient captures requestParam and (method names) in logs as INFO level, which is the default log level of this lib
- That causes applications that use this lib to generate a large amount of logs with default settings.
    - Of course, we can change log level to DEBUG by calling `setLogLevel` though.
- Also, in my opinion, capturing parameters is a little bit overkilling as INFO level as considering this is a library. That should be decided by the users of this lib.
- I'd like to ask if milvus team would change the log level to capture requestParam from INFO to DEBUG.

## What change is made in this PR
- I changed logInfo to logDebug to capture requestParam and two String value that are names of the invoked methods.


yhmo:
make sense

/lgtm
/approve

PR #742, opened 2024-01-10
Support without specifying metricType and indexType

lentitude2tk:


src/main/java/io/milvus/param/IndexType.java
```diff
@@ -27,6 +27,7 @@
  */
 public enum IndexType {
     INVALID,
+    None(0),
```
yhmo, line 29: Both INVALID and None are 0. No need to keep INVALID.

lentitude2tk: OK

PR #746, opened 2024-01-23
V2 Milvus Client

nianliuu:
## Description
Currently we have 5 different milvus SDK: python, java, node, go, restful. With the evolve of milvus, different api have different api format and usage. It's becoming harder to maintain for owner and harder to use for user. Based on this, we reformate unified all 5 SDKs. Providing this V2 API.
## Code Change
- all apis are under MilvusClientV2, user can call all apis under this class
- refactor api request and response to make them simple and easy to user
- use Req and Resp for request and response, not expose java class from grpc
- all files are under v2 path, not overlap with v2 api
- only add basic parameter for this version, some advance feature needs future development
## Test
- add unit tests for all v2 APIs
- may still need test team to test them thoroughly


src/main/java/io/milvus/v2/client/MilvusClientV2.java
```diff
@@ -0,0 +1,485 @@
...
+
+    }
+    /**
+     * Creates a Milvus client instance.
+     *
+     * @param connectConfig Milvus server connection configuration
+     */
+    public void connect(ConnectConfig connectConfig){
+        this.connectConfig = connectConfig;
+        channel = clientUtils.getChannel(connectConfig);
```
yhmo, line 81: If the "channel" is already connected, better to shutdown it before this line.

nianliuu: ok, will update
if(this.channel != null){
            this.channel.shutdownNow();
       }

PR #753, opened 2024-02-02
Fix the issue of default conversion to 0 when id is not of type long

p-moon:
Hello everyone, when I tried to set the primary key to varchar type, I found that after using embedding for retrieval, all returned result IDs were 0. After investigation, it was found that all IDs were forcibly converted to 0 here Work hard to merge the code in a timely manner to fix this issue, and recommend a MILVUS Mybatis style ORM library that integrates with Spring Boot using this SDK, which can help Java developers get started better:

https://github.com/JDK-Plus/spring-boot-starter-milvus

### DIFF
--- a/src/main/java/io/milvus/response/SearchResultsWrapper.java
+++ b/src/main/java/io/milvus/response/SearchResultsWrapper.java
@@ -61,7 +61,6 @@ public class SearchResultsWrapper extends RowRecordWrapper {
         List<IDScore> idScore = getIDScore(indexOfTarget);
         for (int i = 0; i < topK; ++i) {
             QueryResultsWrapper.RowRecord record = new QueryResultsWrapper.RowRecord();
-            record.put("id", idScore.get(i).getLongID());
             record.put("distance", idScore.get(i).getScore());
             buildRowRecord(record, i);
             records.add(record);


p-moon:
@yhmo

src/main/java/io/milvus/response/SearchResultsWrapper.java
```diff
@@ -61,7 +61,6 @@ public List<QueryResultsWrapper.RowRecord> getRowRecords(int indexOfTarget) {
         List<IDScore> idScore = getIDScore(indexOfTarget);
         for (int i = 0; i < topK; ++i) {
             QueryResultsWrapper.RowRecord record = new QueryResultsWrapper.RowRecord();
-            record.put("id", idScore.get(i).getLongID());
```
xiaofan-luan, line 64: I thought we should not remove this info.
but we should get ID based on the schema of the collection.

SomeLike
if useStringPK {
  record.put("id", idScore.get(i).GetStringID());
} else {
  record.put("id", idScore.get(i).getLongID());

}

yhmo:
I have submitted another pr for this fix: https://github.com/milvus-io/milvus-sdk-java/pull/757
Append different IDs by the type of primary key:
```
IDScore score = idScore.get(i);
if (score.getStrID().isEmpty()) {
    record.put("id", score.getLongID());
} else {
    record.put("id", score.getStrID());
}
```

Close this pr as it is replcaed by #757

PR #819, opened 2024-03-27
add retry count for search, if retry for more than five times, throw exception.

nianliuu:


xiaofan-luan:
why not follow timeout?

PR #846, opened 2024-04-09
support searchIterator && queryIterator

lentitude2tk:


examples/main/java/io/milvus/IteratorExample.java
```diff
@@ -0,0 +1,361 @@
...
+        List<List<Float>> vectors = new ArrayList<>();
+        for (int n = 0; n < count; ++n) {
+            List<Float> vector = getFloatVector((float) n);
+            vectors.add(vector);
+        }
+
+        return vectors;
+    }
+
+    private static List<Float> getFloatVector(Float init) {
```
yhmo: There is already a common util method that can replace this method: https://github.com/milvus-io/milvus-sdk-java/blob/ab46ffcfdb17521998086f5fed7f76837677bbec/examples/main/java/io/milvus/CommonUtils.java#L36

examples/main/java/io/milvus/IteratorExample.java
```diff
@@ -0,0 +1,361 @@
...
+            searchIteratorParamBuilder.withTopK(topK);
+        }
+
+        R<SearchIterator> response = milvusClient.searchIterator(searchIteratorParamBuilder.build());
+        handleResponseStatus(response);
+        return response.getData();
+    }
+
+
+    private List<List<Float>> generateFloatVectors(int count) {
```
yhmo: There is already a common util method that can replace this method:
https://github.com/milvus-io/milvus-sdk-java/blob/ab46ffcfdb17521998086f5fed7f76837677bbec/examples/main/java/io/milvus/CommonUtils.java#L45

examples/main/java/io/milvus/IteratorExample.java
```diff
@@ -0,0 +1,361 @@
...
+    private static final Integer VECTOR_DIM = 8;
+    private static final String AGE_FIELD = "userAge";
+
+    private static final String INDEX_NAME = "userFaceIndex";
+    private static final IndexType INDEX_TYPE = IndexType.IVF_FLAT;
+    private static final String INDEX_PARAM = "{\"nlist\":128}";
+    private static final boolean CLEAR_EXIST = false;
+    private static final Integer NUM_ENTITIES = 1000;
+
+    private void handleResponseStatus(R<?> r) {
```
yhmo: There is a common util method can replace this method:
https://github.com/milvus-io/milvus-sdk-java/blob/ab46ffcfdb17521998086f5fed7f76837677bbec/examples/main/java/io/milvus/CommonUtils.java#L30

examples/main/java/io/milvus/IteratorExample.java
lentitude2tk: After the subsequent synchronization of CommonUtils, then proceed with code simplification.

examples/main/java/io/milvus/IteratorExample.java
lentitude2tk: After the subsequent synchronization of CommonUtils, then proceed with code simplification.

examples/main/java/io/milvus/IteratorExample.java
lentitude2tk: After the subsequent synchronization of CommonUtils, then proceed with code simplification.

PR #903, opened 2024-05-21
Replace FastJson by Gson

yhmo:


ll382:
Why is the beta version being put on the main branch? The class conversion hasn't been handled properly.

PR #1046, opened 2024-08-21
add fluent api for AlterCollectionReq & CreateCollectionReq

any35:
as title:

I found that each time an object is created, a lot of fields need to be passed in. 

To make this process more convenient, I added some code to implement a Fluent API, and I hope it will be considered for adoption.

/assign @yelusion2

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,6 +26,7 @@ hs_err_pid*
 .DS_Store
 
 .idea/
+.vscode/
 target/
 volumes/
 *.iml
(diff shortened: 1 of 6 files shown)


yhmo:
@any35 
The code change looks good to me.
I suppose the README.md is only for versions matrix and installation, no need to add develop information in it.
The DEVELOPMENT.md is for developers. "git submodule update --init" is already mentioned in this doc.

PR #1067, opened 2024-09-13
Adding ability to use javax SSLContext directly when setting up connectConfig.

ncover21:
Additionally fixing check in schema converter: 
- The describe collection method tries to get the max_length of the field, but for the field type `JSON` it is an empty value causing an issue when converting the schema.

### DIFF
--- a/src/main/java/io/milvus/v2/client/ConnectConfig.java
+++ b/src/main/java/io/milvus/v2/client/ConnectConfig.java
@@ -24,6 +24,7 @@ import lombok.Data;
 import lombok.NonNull;
 import lombok.experimental.SuperBuilder;
 
+import javax.net.ssl.SSLContext;
 import java.net.URI;
 import java.util.concurrent.TimeUnit;
 
@@ -56,6 +57,8 @@ public class ConnectConfig {
     private Boolean secure = false;
     @Builder.Default
     private long idleTimeoutMs = TimeUnit.MILLISECONDS.convert(24, TimeUnit.HOURS);
+
+    private SSLContext sslContext;
 
     public String getHost() {
         URI uri = URI.create(this.uri);
--- a/src/main/java/io/milvus/v2/utils/ClientUtils.java
+++ b/src/main/java/io/milvus/v2/utils/ClientUtils.java
@@ -24,12 +24,17 @@ import io.grpc.ManagedChannelBuilder;
 import io.grpc.Metadata;
 import io.grpc.netty.shaded.io.grpc.netty.GrpcSslContexts;
 import io.grpc.netty.shaded.io.grpc.netty.NettyChannelBuilder;
+import io.grpc.netty.shaded.io.netty.handler.ssl.ApplicationProtocolConfig;
+import io.grpc.netty.shaded.io.netty.handler.ssl.ClientAuth;
+import io.grpc.netty.shaded.io.netty.handler.ssl.IdentityCipherSuiteFilter;
+import io.grpc.netty.shaded.io.netty.handler.ssl.JdkSslContext;
 import io.grpc.netty.shaded.io.netty.handler.ssl.SslContext;
 import io.grpc.stub.MetadataUtils;
 import io.milvus.client.MilvusServiceClient;
 import io.milvus.grpc.*;
 import io.milvus.v2.client.ConnectConfig;
 import org.apache.commons.lang3.StringUtils;
+import org.jetbrains.annotations.NotNull;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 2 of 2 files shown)


yhmo:
@ncover21 

I have two questions:
1. What does your customized SslContext look like? trustManager and keyManager?
2. Ideally, JSON field's max_length is 65535 returned by the server. I just tested different milvus versions 2.4.4/2.3.5, both returned the correct value max_length=65535. Which version of your milvus?

ncover21:
> @ncover21
> 
> I have two questions:
> 
> 1. What does your customized SslContext look like? trustManager and keyManager?
> 2. Ideally, JSON field's max_length is 65535 returned by the server. I just tested different milvus versions 2.4.4/2.3.5, both returned the correct value max_length=65535. Which version of your milvus?

Hi @yhmo, the nifi implementation uses a trustStore and implements the class [SSLContextService.java](https://github.com/apache/nifi/blob/main/nifi-extension-bundles/nifi-standard-services/nifi-ssl-context-service-api/src/main/java/org/apache/nifi/ssl/SSLContextService.java) and an implementation can be seen here [StandardSSLContextService.java](https://github.com/apache/nifi/blob/9b7b246dcc1f57c2d94d9816c3615f89bb7fe8bf/nifi-extension-bundles/nifi-standard-services/nifi-ssl-context-bundle/nifi-ssl-context-service/src/main/java/org/apache/nifi/ssl/StandardSSLContextService.java). The issue is that we maintain a ssl context from the library `javax.net.ssl.SSLContext `and the milvus sdk uses `io.grpc.netty.shaded.io.netty.handler.ssl.SslContext` so I am having to create a method for translation as well.

Using Zilliz to deploy a milvus DB I have set up a db and you can see the response from zilliz for request `https://cloud-service.cloud.zilliz.com/cloud/v1/instance/collection/describe` and the max_length missing from type `JSON` 
```json 
{
    "Code": 0,
    "Message": "success",
    "RequestId": "f93b4e7b-69b2-4124-b835-f5cdc14a747f",
    "Data": {
        "Description": "",
        "NumOfEntities": 0,
        "CreateTime": 1726453271647,
        "Schemas": [
            {
                "FieldName": "primary_keyd",
                "DataType": 21,
                "ElementType": 0,
                "TypeParams": {
                    "max_length": "256"
                },
                "IndexParams": null,
                "PrimaryKey": true,
                "Description": "The Primary Key",
                "Dimension": null,
                "AutoId": false,
                "IsPartitionKey": false
            },
            {
                "FieldName": "vector",
                "DataType": 101,
                "ElementType": 0,
                "TypeParams": {
                    "dim": "1536"
                },
                "IndexParams": null,
                "PrimaryKey": false,
                "Description": "",
                "Dimension": 1536,
                "AutoId": false,
                "IsPartitionKey": false
            },
            {
                "FieldName": "text",
                "DataType": 21,
                "ElementType": 0,
                "TypeParams": {
                    "max_length": "32000"
                },
                "IndexParams": null,
                "PrimaryKey": false,
                "Description": "",
                "Dimension": null,
                "AutoId": false,
                "IsPartitionKey": false
            },
            {
                "FieldName": "metadata",
                "DataType": 23,
                "ElementType": 0,
                "TypeParams": {},
                "IndexParams": null,
                "PrimaryKey": false,
                "Description": "",
                "Dimension": null,
                "AutoId": false,
                "IsPartitionKey": false
            }
        ],
        "EnableDynamic": true
    }
}
```
A describe of this collection through the sdk causing an exception to occur

ncover21:
Hi @yhmo, looks like there might have been a one off issue with the metadata field in my zilliz db. I tried to replicate it but could not. I have removed the change to the schema class

yhmo:
@ncover21 

The code change loogs good to me.
Could you cherry-pick this pr to 2.4 branch?

/lgtm

ncover21:
> @ncover21
> 
> The code change looks good to me. Could you cherry-pick this pr to 2.4 branch?
> 
> /lgtm

Thanks @yhmo here is the PR: https://github.com/milvus-io/milvus-sdk-java/pull/1068

src/main/java/io/milvus/v2/client/ConnectConfig.java
```diff
@@ -57,6 +58,8 @@ public class ConnectConfig {
     @Builder.Default
     private long idleTimeoutMs = TimeUnit.MILLISECONDS.convert(24, TimeUnit.HOURS);
 
+    private SslContext sslContext;
```
exceptionfactory: Instead of injecting the Netty `SslContext`, it seems worth injecting the `javax.net.ssl.SSLContext` and then converting it to the Netty `SslContext` inside of the `ClientUtils` class. That way it avoids introducing the shaded Netty class reference into the public `ConnectConfig` class.

exceptionfactory: For reference, the [JdkSslContext](https://netty.io/4.1/api/io/netty/handler/ssl/JdkSslContext.html) provides the translation from the Java `SSLContext` to the Netty version.

exceptionfactory (commented):
Thanks for the quick update @ncover21, the adjusted approach using the java `SSLContext` in `ConnectConfig` looks good from my perspective!

src/main/java/io/milvus/v2/client/ConnectConfig.java
ncover21: Addressed suggestion above, changed to SSLContext on ConnectConfig and added conversion in the ClientUtils method

ncover21:
Addressed comments and updated forked PR for branch 2.4 https://github.com/milvus-io/milvus-sdk-java/pull/1068

PR #1116, opened 2024-10-17
Log enhancement: Use slf4j.Logger to replace System.out and print the exception stack trace.

xiaojunxiang2023:
Log enhancement: Use slf4j.Logger to replace System.out and print the exception stack trace.


yhmo:
@xiaojunxiang2023 
The DCO is not passed. Could you use "git commit -s" to add the commit?

PR #1231, opened 2024-12-13
enhance: support hints param

chasingegg:
issue: https://github.com/milvus-io/milvus/issues/37360


chasingegg:
@yhmo will help on search_params refactoring, after that we do not need to change anything in sdk to support new params like `hints`.
/hold

yhmo:
close this pr. I will submit another pr to refine the searchParams

Issue #1413, opened 2025-05-21
Clarify and implement `fieldName` filtering behavior in MilvusClientV2.listIndexes


jamm-king:
**Describe the problem**  
I’m using `MilvusClientV2.listIndexes(ListIndexesReq)` to check and auto-create indexes on specific vector fields before loading a collection into memory. According to the official documentation, `listIndexes` accepts both `collectionName` and `fieldName`, so I expected it to return *only* the indexes defined on that field. In practice, however, it ignores `fieldName` and returns *all* index names in the collection.

**What I found in the code**  
- `listIndexes` builds a `DescribeIndexRequest` with the provided `fieldName`, calls the gRPC stub’s `describeIndex(...)`, but then simply maps **every** returned `IndexDescription` to its `indexName` without any client-side filtering.  
- The SDK’s separate `describeIndex` wrapper *does* filter by `fieldName` (or `indexName`) after the RPC, but `listIndexes` does not.

**Expected behavior**  
- If `fieldName` is supplied, `listIndexes` should return *only* the index names that correspond to that field.  
- If the intention is to always return all indexes, then the `fieldName` parameter should be removed from `ListIndexesReq` to avoid confusion.

**Suggested fix**  
In `GrpcMilvusClient.listIndexes`, after receiving the `DescribeIndexResponse`, add a filter step such as:
```java
response.getIndexDescriptionsList().stream()
    .filter(desc -> request.getFieldName() == null
        || desc.getFieldName().equals(request.getFieldName()))
    .map(IndexDescription::getIndexName)
    .collect(Collectors.toList());

yhmo:
Good catch! This pr will fix this bug: https://github.com/milvus-io/milvus-sdk-java/pull/1415

PR #1414, opened 2025-05-21
fix(listIndexes): apply fieldName filter when provided #1413

jamm-king:
### What’s changed
- In `IndexService.listIndexes()`, add client-side filtering so that when `ListIndexesReq.fieldName` is non-null, only indexes on that field are returned.
- Preserve existing behavior: if `fieldName` is null, return all index names.

### Why
- According to documentation, `listIndexes` takes `fieldName` as a parameter, so users expect per-field index lists. Without this fix, it always returns every index in the collection, causing confusion and downstream load failures.

### How
1. After calling `blockingStub.describeIndex(...)`, stream-filter the `IndexDescription` list by `desc.getFieldName().equals(fieldName)` if provided.
2. Map the filtered list to `getIndexName()`.
3. Add unit tests covering:
   - No `fieldName` → all indexes
   - Valid `fieldName` → only that index

### Testing
- Added unit test to verify filedName specified case

### DIFF
--- a/sdk-core/src/main/java/io/milvus/v2/service/index/IndexService.java
+++ b/sdk-core/src/main/java/io/milvus/v2/service/index/IndexService.java
@@ -178,11 +178,10 @@ public class IndexService extends BaseService {
             return new ArrayList<>();
         }
         rpcUtils.handleResponse(title, response.getStatus());
-        List<String> indexNames = new ArrayList<>();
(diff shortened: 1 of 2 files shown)


yhmo:
Thank you for the effort. I intend to add the test case in the MilvusClientV2DockerTest.java instead of IndexTest.java, so I just created a new pr to fix it. Will close this pr.

Issue #1434, opened 2025-06-27
Error when set limit only without topK when search


topikachu:
I meet error 
io.milvus.v2.exception.MilvusClientException: topk [0] is invalid, it should be in range [1, 16384], but got 0

when set limit only wihout topK

``` java

import com.google.gson.Gson;
import com.google.gson.JsonObject;
import io.milvus.v2.client.ConnectConfig;
import io.milvus.v2.client.MilvusClientV2;
import io.milvus.v2.service.collection.request.CreateCollectionReq;
import io.milvus.v2.service.collection.request.DropCollectionReq;
import io.milvus.v2.service.vector.request.InsertReq;
import io.milvus.v2.service.vector.request.SearchReq;
import io.milvus.v2.service.vector.request.data.FloatVec;
import io.milvus.v2.service.vector.response.InsertResp;
import org.junit.jupiter.api.Test;
import org.testcontainers.junit.jupiter.Container;
import org.testcontainers.junit.jupiter.Testcontainers;
import org.testcontainers.milvus.MilvusContainer;

import java.sql.SQLException;
import java.util.Arrays;
import java.util.Collections;
import java.util.List;
import java.util.Map;

import static org.assertj.core.api.Assertions.assertThat;


@Testcontainers
public class LimitTest {
    @Container
    static MilvusContainer milvus = new MilvusContainer("milvusdb/milvus:v2.5.11");


    @Test
    void testLimit() throws SQLException {
        // 1. Connect to Milvus server
        ConnectConfig connectConfig = ConnectConfig.builder()
                .uri(milvus.getEndpoint())
//                .token(TOKEN)
                .build();

        MilvusClientV2 client = new MilvusClientV2(connectConfig);

        String collectionName = "quick_setup";

        var delReq = DropCollectionReq.builder()
                .collectionName(collectionName)
                .build();
        client.dropCollection(delReq);


// 2. Create a collection in quick setup mode
        CreateCollectionReq quickSetupReq = CreateCollectionReq.builder()
                .collectionName(collectionName)
                .dimension(5)
                .build();

        client.createCollection(quickSetupReq);


        Gson gson = new Gson();
        List<JsonObject> data = Arrays.asList(
                gson.fromJson("{\"id\": 0, \"vector\": [0.3580376395471989f, -0.6023495712049978f, 0.18414012509913835f, -0.26286205330961354f, 0.9029438446296592f], \"color\": \"pink_8682\"}", JsonObject.class),
[...]

PR #1435, opened 2025-06-27
fix milvus-io/milvus-sdk-java#1434

topikachu:
Uses limit as topK value if topK is not set.


yhmo:
The issue has been fixed by the two prs: https://github.com/milvus-io/milvus-sdk-java/pull/1439
  https://github.com/milvus-io/milvus-sdk-java/pull/1438

PR #1475, opened 2025-07-15
Fix indexParams$value NPE

yeyinglang:
fix issue #1470  bug


yhmo:
The fix is in this pr https://github.com/milvus-io/milvus-sdk-java/pull/1465

Issue #1614, opened 2025-09-19
[bug] Unable to load collection with v2 client (databaseName not sent to server?)


ac101m:
When using client v2 to load a collection sync like so:

```kotlin
val loadCollectionReq = LoadCollectionReq.builder()
    .databaseName(testDatabaseName)
    .collectionName(testCollectionName)
    .sync(true)
    .build()

clientV2.loadCollection(loadCollectionReq)
```

The request times out and there are a lot of lines like follows in the client logs:

```
14:48:06.129 [grpc-default-worker-ELG-1-8] DEBUG io.grpc.netty.shaded.io.grpc.netty.NettyClientHandler -- [id: 0x7caf9efe, L:/127.0.0.1:46076 - R:localhost/127.0.0.1:19530] OUTBOUND HEADERS: streamId=21 headers=GrpcHttp2OutboundHeaders[:authority: localhost:19530, :path: /milvus.proto.milvus.MilvusService/GetLoadState, :method: POST, :scheme: http, content-type: application/grpc, te: trailers, user-agent: grpc-java-netty/1.59.1, grpc-accept-encoding: gzip] streamDependency=0 weight=16 exclusive=false padding=0 endStream=false
14:48:06.130 [grpc-default-worker-ELG-1-8] DEBUG io.grpc.netty.shaded.io.grpc.netty.NettyClientHandler -- [id: 0x7caf9efe, L:/127.0.0.1:46076 - R:localhost/127.0.0.1:19530] OUTBOUND DATA: streamId=21 padding=0 endStream=true length=41 bytes=00000000241222746573745f636f6c6c656374696f6e5f383730313337373831353833303734393532
14:48:06.130 [grpc-default-worker-ELG-1-8] DEBUG io.grpc.netty.shaded.io.grpc.netty.NettyClientHandler -- [id: 0x7caf9efe, L:/127.0.0.1:46076 - R:localhost/127.0.0.1:19530] INBOUND WINDOW_UPDATE: streamId=0 windowSizeIncrement=41
14:48:06.130 [grpc-default-worker-ELG-1-8] DEBUG io.grpc.netty.shaded.io.grpc.netty.NettyClientHandler -- [id: 0x7caf9efe, L:/127.0.0.1:46076 - R:localhost/127.0.0.1:19530] INBOUND PING: ack=false bytes=145258749040133895
[...]

PR #1616, opened 2025-09-22
Support cross-database ability for more interfaces

yhmo:
https://github.com/milvus-io/milvus-sdk-java/issues/1614

Add parameter "databaseName" for the following requests:
- GetCollectionStatsReq
- GetLoadStateReq
- HasCollectionReq
- ListCollectionsReq
- ReleaseCollectionReq
- RenameCollectionReq
- DropIndexReq
- ListIndexesReq
- CreatePartitionReq
- DropPartitionReq
- GetPartitinStatsReq
- HasPartitionReq
- ListPartitionsReq
- LoadPartitionsReq
- ReleasePartitionsReq
- CompactReq
- FlushReq
- GetPErsistentSegmentInfoReq
- GetQuerySegmentInfoReq

### DIFF
--- a/sdk-core/src/main/java/io/milvus/v2/service/collection/request/ListCollectionsReq.java
+++ b/sdk-core/src/main/java/io/milvus/v2/service/collection/request/ListCollectionsReq.java
@@ -1,0 +1,85 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package io.milvus.v2.service.collection.request;
+
+import org.apache.commons.lang3.builder.EqualsBuilder;
+
+import java.util.Objects;
+
+public class ListCollectionsReq {
+    private String databaseName;
+
+    private ListCollectionsReq(Builder builder) {
+        this.databaseName = builder.databaseName;
+    }
+
+    public static Builder builder() {
+        return new Builder();
+    }
+
+    // Getters
+    public String getDatabaseName() {
+        return databaseName;
+    }
+
+    // Setters
+    public void setDatabaseName(String databaseName) {
+        this.databaseName = databaseName;
+    }
+
+    @Override
+    public boolean equals(Object obj) {
+        if (this == obj) return true;
+        if (obj == null || getClass() != obj.getClass()) return false;
+
+        ListCollectionsReq that = (ListCollectionsReq) obj;
+
+        return new EqualsBuilder()
+                .append(databaseName, that.databaseName)
+                .isEquals();
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(databaseName);
+    }
+
+    @Override
+    public String toString() {
+        return "ListCollectionsReq{" +
+                "databaseName='" + databaseName +
+                '}';
+    }
+
+    public static class Builder {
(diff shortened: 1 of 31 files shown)


sdk-core/src/main/java/io/milvus/v2/service/collection/request/ListCollectionsReq.java
```diff
@@ -0,0 +1,29 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package io.milvus.v2.service.collection.request;
+
+import lombok.Data;
+import lombok.experimental.SuperBuilder;
+
+@Data
```
yeyinglang: If we plan to remove lombok, can we avoid introducing it here ?
ps: 
To ensure SDK stability , we do not recommand to use experimental API in the core module

sdk-core/src/main/java/io/milvus/v2/service/database/DatabaseService.java
```diff
@@ -32,7 +32,7 @@
 
 public class DatabaseService extends BaseService {
     public Void createDatabase(MilvusServiceGrpc.MilvusServiceBlockingStub blockingStub, CreateDatabaseReq request) {
-        String title = String.format("CreateDatabaseRequest databaseName:%s", request.getDatabaseName());
+        String title = String.format("Create database: '%s'", request.getDatabaseName());
```
yeyinglang, line 35: String.format() uses regular expressions, which is quite CPU-intensive, is three an alternative? eg: "Create databases" + request.getDatabaseName()

sdk-core/src/main/java/io/milvus/v2/service/index/IndexService.java
```diff
@@ -192,46 +211,45 @@ public List<String> listIndexes(MilvusServiceGrpc.MilvusServiceBlockingStub bloc
...
                 throw new MilvusClientException(ErrorCode.SERVER_ERROR, msg);
             }
 
             boolean allIndexBuildCompleted = true;
-            for (DescribeIndexResp.IndexDesc index : indices) {
-                if (index.getIndexState() == IndexBuildState.Failed) {
-                    String msg = "Index is failed, reason: " + index.getIndexFailedReason();
+            for (DescribeIndexResp.IndexDesc desc : descs) {
+                if (desc.getIndexState() == IndexBuildState.Failed) {
+                    String msg = "Index is failed, reason: " + desc.getIndexFailedReason();
```
yeyinglang: can we do this way ?  To reduce the creation of one variable and simplify the logic? 

eg : throw new MilvusClientException(ErrorCode.SERVER_ERROR, "Index is failed, reason: " + desc.getIndexFailedReason());

sdk-core/src/main/java/io/milvus/v2/service/collection/request/ListCollectionsReq.java
yhmo: I see, will rewrite this file.

sdk-core/src/main/java/io/milvus/v2/service/database/DatabaseService.java
yhmo: I think it is fine. In most cases, the performance bottleneck is not on the client-side. And this interface createDatabase() is not expected to be called frequently.
If we feel performance is affected by String.format() someday, we can rewrite them in one commit.

sdk-core/src/main/java/io/milvus/v2/service/index/IndexService.java
yhmo: ok

yhmo: done

sdk-core/src/main/java/io/milvus/v2/service/collection/request/ListCollectionsReq.java
yhmo: done

sdk-core/src/main/java/io/milvus/v2/service/database/DatabaseService.java
yeyinglang: I agree, **the performance bottleneck is not on the client-side**
But in daily work , Milvus SDK may only be used for a specific case.  and we often use flame graph to identify where the client performance bottleneck , and we will optimize it to reduce unnecessary CPU (for cost optimization)

yeyinglang: **If we feel performance is affected by String.format() someday, we can rewrite them in one commit.**

I think it's ok

PR #1653, opened 2025-10-18
Remove Lombok from milvus sdk-core 

yeyinglang:
[1469](https://github.com/milvus-io/milvus-sdk-java/issues/1469)

tips: compatible with the previous version

## Remove Lombok will be implemented in 5 steps 

1. Remove Lombok at one package which function has not yet been widely popularized (eg:io.milvus.v2.service.resourcegroup )
2. Remove Lombok at onepackage which functions have been popularized
3. Remove Lombok at large package (eg io.milvus.v2.service)
4. Remove Lombok at Milvus V2 Package (io.milvus.v2)
5. Remove Lombok at all src places (io.milvus)

Remove Lombok from milvus sdk-core
image

## Test Case Results 
image

### DIFF
--- a/sdk-core/src/main/java/io/milvus/pool/PoolConfig.java
+++ b/sdk-core/src/main/java/io/milvus/pool/PoolConfig.java
@@ -170,10 +170,10 @@ public class PoolConfig {
     }
 
     public static class Builder {
-        private int maxIdlePerKey = 5;
+        private int maxIdlePerKey = 10;
         private int minIdlePerKey = 0;
-        private int maxTotalPerKey = 10;
-        private int maxTotal = 50;
+        private int maxTotalPerKey = 50;
+        private int maxTotal = 1000;
         private boolean blockWhenExhausted = true;
         private Duration maxBlockWaitDuration = Duration.ofSeconds(3L);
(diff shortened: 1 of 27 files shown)


sdk-core/pom.xml
```diff
@@ -154,7 +154,7 @@
             <groupId>org.projectlombok</groupId>
             <artifactId>lombok</artifactId>
             <version>${lombok.version}</version>
-            <scope>provided</scope>
+            <scope>test</scope>
```
yhmo: Is this still necessary? Can we remove this dependency?

sdk-core/src/main/java/io/milvus/v2/service/collection/request/AddCollectionFieldReq.java
```diff
@@ -19,12 +19,203 @@
...
+    public String getDatabaseName() {
+        return databaseName;
+    }
+
+    public void setDatabaseName(String databaseName) {
+        this.databaseName = databaseName;
+    }
+
+    @Override
+    public boolean equals(Object obj) {
```
yhmo: Is this method necessary? We don't have special comparison logic for request classes.

Issue #1716, opened 2025-12-10
VectorUtils#vector result is null ,leading to NPE


yeyinglang:
# question:

<img width="2222" height="380" alt="image" src="https://github.com/user-attachments/assets/1e41159b-2334-436c-be2a-d4451da992a8" />

# Reproduction Code

```java
            SearchReq searchReq = SearchReq.builder()
                    .collectionName(Constant.COLLECTION_NAME)
                    .filter("id ==4627727850274838")
                    .outputFields(Arrays.asList("text", "id"))
                    .limit(1)
                    .build();

            SearchResp searchResp = milvusClient.search(searchReq);
```
*notes:  in this case,we should use query. *

yhmo:
The fix pr is mergerd.

PR #1717, opened 2025-12-10
fix vector result is null ,leading to NPE

yeyinglang:
https://github.com/milvus-io/milvus-sdk-java/issues/1716


sdk-core/src/main/java/io/milvus/v2/utils/VectorUtils.java
```diff
@@ -159,7 +159,7 @@ public SearchRequest ConvertToGrpcSearchRequest(SearchReq request) {
 
         // prepare target, the input could be vectors or string list for doc-in-doc-out
         List<BaseVector> vectors = request.getData();
-        if (vectors.isEmpty()) {
+        if (vectors == null || vectors.isEmpty()) {
```
yhmo, line 162: We can set a default list for the SearchReqBuilder:
```
public static class SearchReqBuilder {
        ......
        private List<BaseVector> data = new ArrayList<>(); // default value
```

PR #1797, opened 2026-04-07
feat(mol): add MOL type/function and PATTERN index

862103595:
related: 

- add `DataType.Mol`
- add `FunctionType.MOL_FINGERPRINT`
- add `PATTERN` index enum (v1/v2)
- support `mol_smiles_data` serialization and SMILES validation

### DIFF
--- a/sdk-core/src/test/java/io/milvus/client/MilvusServiceClientTest.java
+++ b/sdk-core/src/test/java/io/milvus/client/MilvusServiceClientTest.java
@@ -2866,6 +2866,13 @@ class MilvusServiceClientTest {
         }
         testScalarField(ScalarField.newBuilder().setStringData(strBuilder).build(),
                 DataType.VarChar, dim);
+
+        MolSmilesArray.Builder molBuilder = MolSmilesArray.newBuilder();
+        for (long i = 0; i < dim; ++i) {
+            molBuilder.addData("C" + i);
+        }
+        testScalarField(ScalarField.newBuilder().setMolSmilesData(molBuilder).build(),
+                DataType.Mol, dim);
     }
 
     @Test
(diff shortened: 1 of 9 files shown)


liliu-z:
**sdk-bulkwriter/src/main/java/io/milvus/bulkwriter/BulkWriter.java:1** The sdk-bulkwriter module is missing Mol type support entirely. `verifyByDatatype` in BulkWriter.java, schema construction in ParquetUtils.java, and the string serialization branch in ParquetFileWriter.java all need `case Mol:` branches added — without them, bulk import of Mol data will throw or silently fail.

yhmo (changes requested):
Found one correctness/test-coverage gap in the new MOL support that should be fixed before merging.

yhmo (commented):
Adding the inline note that failed to attach in the earlier review.

sdk-core/src/test/java/io/milvus/client/MilvusServiceClientTest.java
```diff
@@ -2866,6 +2866,13 @@ void testFieldDataWrapper() {
         }
         testScalarField(ScalarField.newBuilder().setStringData(strBuilder).build(),
                 DataType.VarChar, dim);
+
+        MolSmilesArray.Builder molBuilder = MolSmilesArray.newBuilder();
+        for (long i = 0; i < dim; ++i) {
+            molBuilder.addData("C" + i);
+        }
+        testScalarField(ScalarField.newBuilder().setMolSmilesData(molBuilder).build(),
+                DataType.Mol, dim);
```
yhmo, line 2875: The added test only covers `FieldDataWrapper` readback for `mol_smiles_data`. This PR also changes the write-side conversion paths in `ParamUtils` (`checkFieldData`, `checkFieldValue`, `genScalarField`, `objectToValueField`, `valueFieldToObject`) plus the V2 enum mapping for `DataType.Mol(27)`, but none of those new paths are exercised here. Since MOL support is mostly serialization/deserialization plumbing, it would be safer to add at least one request-side test that inserts/upserts a MOL scalar value and one enum mapping assertion for code `27`, otherwise a regression on the write path could still pass these tests.

PR #1810, opened 2026-04-24
Add option parameter to ConnectConfig and ConnectParam

lentitude2tk:
## Summary
- Add `Map<String, String> option` field to `ConnectConfig` (V2 client) and `ConnectParam` (V1 client)
- The option map gets forwarded to `ClientInfo.reserved` when connecting to the server
- Aligns with pymilvus client which recently added the same capability

## Test plan
- [x] Compile verification passed
- [ ] Manual test with Milvus server to verify option is forwarded to ClientInfo.reserved


liliu-z:
**:0** No tests added for the new option parameter functionality. Need coverage for builder get/set, null safety, defensive copying, and serialization into ClientInfo.reserved.

PR #1811, opened 2026-04-24
Add identifier interceptor to align with pymilvus behavior

lentitude2tk:
## Summary
- Extract identifier from `ConnectResponse` after successful Connect RPC
- Inject identifier into every subsequent gRPC request's metadata via a `ClientInterceptor`
- Required for kite-coordinator to resolve cluster_id from connection bindings
- Both V1 (`MilvusServiceClient`) and V2 (`MilvusClientV2`) clients are updated

## Test plan
- [x] Compile verification passed
- [x] Unit tests for `IdentifierInterceptor` added
- [ ] Manual test with Milvus server to verify identifier propagation

### DIFF
--- a/sdk-core/src/main/java/io/milvus/v2/client/MilvusClientV2.java
+++ b/sdk-core/src/main/java/io/milvus/v2/client/MilvusClientV2.java
@@ -19,7 +19,10 @@
 
 package io.milvus.v2.client;
 
+import io.grpc.Channel;
+import io.grpc.ClientInterceptors;
 import io.grpc.ManagedChannel;
+import io.milvus.common.interceptor.IdentifierInterceptor;
 import io.milvus.grpc.ClientInfo;
 import io.milvus.grpc.ConnectRequest;
 import io.milvus.grpc.ConnectResponse;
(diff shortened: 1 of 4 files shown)


sdk-core/src/main/java/io/milvus/v2/client/MilvusClientV2.java
```diff
@@ -168,7 +171,13 @@ private void connect(ConnectConfig connectConfig) {
...
         try {
             blockingStub = MilvusServiceGrpc.newBlockingStub(channel).withWaitForReady();
-            connect(connectConfig, blockingStub);
+            long identifier = connect(connectConfig, blockingStub);
+
+            // Wrap channel with identifier interceptor so that every subsequent RPC
+            // carries the identifier in gRPC metadata, aligning with pymilvus behavior.
+            Channel interceptedChannel = ClientInterceptors.intercept(channel,
+                    new IdentifierInterceptor(identifier));
+            blockingStub = MilvusServiceGrpc.newBlockingStub(interceptedChannel).withWaitForReady();
```
liliu-z, line 180: flush() creates a stub directly from the raw ManagedChannel, bypassing the IdentifierInterceptor attached to blockingStub. Polling RPCs via getFlushState() will not carry the identifier header. Derive the stub from this.blockingStub instead.

PR #1820, opened 2026-05-09
Support Snapshot interfaces

yhmo:


liliu-z:
**src/main/java/io/milvus/v2/service/snapshot/request/RestoreSnapshotReq.java:22** The proto `RestoreSnapshotRequest` (milvus.proto:2773) defines `bool rewrite_data = 5` to enable import-rewrite restore mode, but `RestoreSnapshotReq` does not expose this field and `SnapshotService.restoreSnapshot()` never calls `builder.setRewriteData()`. Java SDK users cannot use the server-side import-rewrite restore mode. Add a `Boolean rewriteData` field (defaulting to `false`) to the request DTO and set it on the proto builder when `true`.

liliu-z:
**src/main/java/io/milvus/v2/service/snapshot/response/RestoreSnapshotJobInfo.java:27** The restore state is converted via `info.getState().name()` (SnapshotService.java:225) to a raw string, and example code hardcodes proto enum names like `"RestoreSnapshotCompleted"` / `"RestoreSnapshotFailed"` for comparison (SnapshotExample.java:178,181). This is fragile and error-prone for users. While consistent with existing patterns in `UtilityService.java`, consider at minimum providing string constants (e.g. `STATE_COMPLETED`) to reduce typo risk.

PR #1826, opened 2026-05-09
Use milvus 3.0 container to test

yhmo:


sdk-core/src/test/java/io/milvus/docker-compose.yml
```diff
@@ -0,0 +1,67 @@
+version: '3.5'
+
+services:
+  etcd:
+    container_name: milvus-etcd
```
liliu-z: The new compose file uses global fixed container names (`milvus-standalone`, `milvus-etcd`, `milvus-minio`) and network name (`milvus`) that collide with `tests/milvustest*` compose files. No `-p` project name or top-level `name:` is set. Combined with the `alreadyHealthy` reuse logic in `TestUtils.startMilvusStandalone()` (L41), `sdk-core` tests can silently connect to a Milvus instance started by a different test suite — potentially a wrong image version. The old compose used `milvus-javasdk-*` prefixes which avoided this. Restore unique prefixes or pass `-p <unique-project>` in `runDockerCompose`.

sdk-core/src/test/java/io/milvus/TestUtils.java
```diff
@@ -4,14 +4,196 @@
...
+
+    public static void startMilvusStandalone(File composeFile, File volumeDirectory, List<String> containerNames) {
+        if (!composeFile.isFile()) {
+            throw new IllegalStateException("Milvus docker compose file not found: " + composeFile);
+        }
+        if (containerNames == null || containerNames.isEmpty()) {
+            throw new IllegalStateException("Milvus docker compose container names are empty");
+        }
+
+        boolean alreadyHealthy = allContainersHealthy(containerNames);
```
liliu-z: The `alreadyHealthy` logic silently reuses a residual container from a prior abnormal exit. This is an isolation regression vs. Testcontainers (which always started a fresh container). Concrete failure path: `testCacheCollectionSchema()` (L2963) creates `test_database` but never drops it; on re-run the residual DB causes `testDatabase()` L2877 `createDatabase("test_database")` to fail. At minimum, log a warning when reusing an existing healthy container, or provide a force-rebuild option.

sdk-core/src/test/java/io/milvus/client/MilvusClientDockerTest.java
```diff
@@ -91,16 +91,9 @@ public DescribeCollectionResponse getCollectionInfo(String databaseName, String
         }
     }
 
-    @Container
-    private static final MilvusContainer milvus = new MilvusContainer(TestUtils.MilvusDockerImageID)
-            .withEnv("DEPLOY_MODE", "STANDALONE");
-
     @BeforeAll
```
liliu-z, line 94: Both `MilvusClientDockerTest` (L94) and `MilvusClientV2DockerTest` (L101) removed `@Testcontainers(disabledWithoutDocker=true)`. Previously, `mvn verify` in a Docker-less environment would gracefully skip these tests; now it fails outright. Add a Docker availability check via `Assumptions.assumeTrue` at the start of `startMilvusStandalone` to restore graceful skip behavior.

PR #1837, opened 2026-05-15
Support field-level operations for upsert

yhmo:


sdk-core/src/main/java/io/milvus/v2/utils/DataUtils.java
```diff
@@ -66,11 +66,13 @@ public UpsertRequest convertGrpcUpsertRequest(UpsertReq requestParam, DescribeCo
             String collectionName = requestParam.getCollectionName();
 
             // generate upsert request builder
+            List<FieldPartialUpdateOp> fieldOps = convertFieldOps(requestParam.getFieldOps());
             MsgBase msgBase = MsgBase.newBuilder().setMsgType(MsgType.Upsert).build();
             upsertBuilder = UpsertRequest.newBuilder()
                     .setCollectionName(collectionName)
                     .setBase(msgBase)
-                    .setPartialUpdate(requestParam.isPartialUpdate())
+                    .setPartialUpdate(requestParam.isPartialUpdate() || hasNonReplaceFieldOps(fieldOps))
```
liliu-z, line 74: `convertGrpcUpsertRequest()` correctly computes `effectivePartialUpdate = isPartialUpdate() || hasNonReplaceFieldOps(fieldOps)` at line 74 and sets it on the gRPC builder. However, `fillFieldsData()` at line 172 still reads `requestParam.isPartialUpdate()` — the original `false` value. Then `checkAndSetRowData()` at line 310 rejects missing required fields because `partialUpdate` is false. Users who set `fieldOps` with ARRAY_APPEND/ARRAY_REMOVE without explicit `.partialUpdate(true)` will get a spurious `"The field: X is not provided"` exception from the SDK before the request ever reaches the server. Fix: pass `effectivePartialUpdate` (computed at line 74) into `fillFieldsData`/`checkAndSetRowData` instead of `requestParam.isPartialUpdate()`.

sdk-core/src/test/java/io/milvus/v2/service/vector/VectorTest.java
```diff
@@ -75,6 +80,34 @@ void testUpsert() {
         logger.info(statusR.toString());
     }
 
+    @Test
```
liliu-z, line 83: The new test uses a schema with only `id` and `vector`, and the request provides both fields. This cannot catch the high-severity validation bug because no required fields are missing. A test case is needed with: (1) a schema containing an additional non-nullable field (e.g., an Array field), (2) a data payload that omits it, and (3) `fieldOps = [ARRAY_APPEND("arrayField")]` without `.partialUpdate(true)`. Secondary gaps: no coverage for REPLACE, ARRAY_REMOVE, mixed ops, or error paths.

PR #1888, opened 2026-06-10
Struct field supports BINARY_VECTOR/FLOAT16_VECTOR/BFLOAT16_VECTOR/INT8_VECTOR

yhmo:


yhmo (commented):
Found one correctness issue in the new struct-vector test coverage: the binary struct-field test relies on AUTOINDEX but later asserts a specific concrete index type, which makes the test brittle against server-side auto-index selection changes.

examples/src/main/java/io/milvus/v2/StructExample.java
```diff
@@ -271,6 +338,75 @@ private static void search(String annsField, List<BaseVector> searchData) {
...
+        );
+    }
+
+    private static void searchBinaryVectorField(String filter) {
+        searchStructField(
+                BINARY_VECTOR_FIELD,
+                BINARY_VECTOR_METRIC,
+                filter,
+                struct -> new BinaryVec((ByteBuffer) struct.get(BINARY_VECTOR_FIELD)),
+                "Search on binary field '" + BINARY_VECTOR_FIELD + "' in struct '" + STRUCT_FIELD + "'"
```
yhmo, line 373: `examples/src/main/java/io/milvus/v2/StructExample.java:376`

`searchStructField(...)` takes a `metricType`, but before this line the helper never forwards it into the `SearchReq`, so all of the non-float struct-vector searches end up relying on whatever metric the server/index picks by default. That makes the example misleading and can produce wrong behavior if the default does not match the index that was just created. Please pass the supplied metric through to the `SearchReq`.

sdk-core/src/test/java/io/milvus/v2/client/MilvusClientV2DockerTest.java
```diff
@@ -1387,41 +1498,93 @@ void testStruct() {
...
+                .build());
+        Assertions.assertEquals(0L, (long) float16SearchResp.getSearchResults().get(0).get(0).getId());
+        Assertions.assertEquals(9L, (long) float16SearchResp.getSearchResults().get(1).get(0).getId());
+
+        SearchResp bfloat16SearchResp = client.search(SearchReq.builder()
+                .collectionName(randomCollectionName)
+                .annsField("st1[bfloat16_vector]")
+                .data(Arrays.asList(bfloat16EmbList0, bfloat16EmbList1))
+                .limit(topK)
+                .metricType(IndexParam.MetricType.MAX_SIM_IP)
```
yhmo: `sdk-core/src/test/java/io/milvus/v2/client/MilvusClientV2DockerTest.java:1570`

The bfloat16 search uses `MAX_SIM_IP` here, but the index for `st1[bfloat16_vector]` above was created with `MAX_SIM_COSINE`. That metric mismatch makes the test brittle and is already showing up in CI as a wrong top-1 ID. Please make the search metric match the index metric (or omit it and rely on the index configuration consistently).

examples/src/main/java/io/milvus/v2/SnapshotExample.java
```diff
@@ -54,7 +54,7 @@
 import java.util.List;
 
 public class SnapshotExample {
-    private static final String COLLECTION_NAME = "java_sdk_example_snapshot_v2";
+    private static final String COLLECTION_NAME = "CPP_V2_SNAPSHOT";
```
yhmo: Middle ---- `examples/src/main/java/io/milvus/v2/SnapshotExample.java:57`

This example now uses the fixed collection name `CPP_V2_SNAPSHOT`, which conflicts with the Java example naming pattern and can also interfere with other SDK examples or any existing collection with the same name on a shared Milvus instance. Please switch back to a Java-specific example name so the sample remains self-contained and rerunnable.

examples/src/main/java/io/milvus/v2/SnapshotExample.java
```diff
@@ -161,11 +161,11 @@ public static void main(String[] args) throws InterruptedException {
                         job.getJobId(), job.getSnapshotName(), job.getCollectionName(), job.getState(), job.getProgress());
             }
 
-            client.dropSnapshot(DropSnapshotReq.builder()
-                    .snapshotName(SNAPSHOT_NAME)
-                    .collectionName(COLLECTION_NAME)
-                    .build());
-            System.out.printf("Snapshot '%s' dropped%n", SNAPSHOT_NAME);
+//            client.dropSnapshot(DropSnapshotReq.builder()
```
yhmo: Middle ---- `examples/src/main/java/io/milvus/v2/SnapshotExample.java:164`

The snapshot cleanup is commented out here, so repeated runs leave `java_sdk_example_snapshot_backup` behind on the server. That makes the example non-idempotent and can cause later runs to fail or accumulate server-side garbage state. Please either restore the cleanup or make the retained snapshot behavior explicit and safe.

sdk-core/src/main/java/io/milvus/v2/service/utility/UtilityService.java
```diff
@@ -133,13 +133,18 @@ public Void waitFlush(MilvusServiceGrpc.MilvusServiceBlockingStub blockingStub,
                 Long flushTs = collectionFlushTs.get(collectionName);
                 boolean flushed = false;
                 while (!flushed) {
+                    System.out.println("Flush wait check: collection=" + collectionName
```
yhmo: Middle ---- `waitFlush()` now writes directly to stdout on every 1s poll for every collection. Because this loop has no timeout, slow flushes can spam application or CI logs and bypass the SDK logger configuration entirely. Please remove the `System.out.println(...)` calls or switch this to normal debug logging with an explicit opt-in level.

PR #1895, opened 2026-06-16
Support dumpMessages interface

yhmo:


sdk-core/src/main/java/io/milvus/v2/client/MilvusClientV2.java
```diff
@@ -1459,6 +1461,16 @@ public ListFileResourcesResp listFileResources(ListFileResourcesReq request) {
...
     }
 
+    /**
+     * Dump messages from a WAL range for data salvage.
+     *
+     * @param request dump messages request
+     * @return DumpMessagesResp containing a lazy stream of dumped messages
+     */
+    public DumpMessagesResp dumpMessages(DumpMessagesReq request) {
+        return cdcService.dumpMessages(this.getRpcStub(), request);
```
yhmo, line 1471: Medium ---- Should this go through `rpcUtils.retry(...)` like the other CDC entrypoints in this class? `getReplicateInfo()`, `getReplicateConfiguration()`, and `updateReplicateConfiguration()` all wrap the service call, and the iterator-style APIs (`queryIterator`/`searchIterator`) do the same for their initial RPC. As written, an `UNAVAILABLE` or `RESOURCE_EXHAUSTED` raised while opening the dump stream bypasses the SDK retry/topology-refresh path and leaks the transport failure directly to callers.

sdk-core/src/main/java/io/milvus/v2/service/cdc/CDCService.java
```diff
@@ -82,4 +94,92 @@ public UpdateReplicateConfigurationResp updateReplicateConfiguration(MilvusServi
...
+            private DumpMessageInfo nextMessage;
+            private boolean nextReady;
+
+            @Override
+            public boolean hasNext() {
+                if (nextReady) {
+                    return true;
+                }
+                while (responseIterator.hasNext()) {
+                    DumpMessagesResponse response = responseIterator.next();
```
yhmo, line 143: Medium ---- Could we normalize transport failures inside this iterator wrapper as well? `responseIterator.hasNext()` / `next()` run outside `rpcUtils.retry(...)` and without catching `StatusRuntimeException`, so a network error after the stream has been created will surface as a raw gRPC exception instead of the usual `MilvusClientException`. The new tests cover `Status` frames, but they do not exercise this transport-error path.

sdk-core/src/main/java/io/milvus/v2/client/MilvusClientV2.java
yhmo: no need

PR #1905, opened 2026-06-18
Improve volume upload concurrency and progress

lentitude2tk:
## Summary
- move Volume upload options to UploadFilesRequest and reduce default file upload concurrency to 5
- add byte-level upload progress for large files
- make VolumeFileManager upload state request-local so one client instance can run concurrent uploads safely
- keep existing public APIs source-compatible

## Tests
- mvn -pl sdk-bulkwriter -DskipTests compile
- mvn -pl sdk-bulkwriter -Dtest=VolumeManagerTest,ProgressInputStreamTest test

### DIFF
--- a/sdk-bulkwriter/src/main/java/io/milvus/bulkwriter/VolumeFileManager.java
+++ b/sdk-bulkwriter/src/main/java/io/milvus/bulkwriter/VolumeFileManager.java
@@ -23,6 +23,7 @@ import com.google.gson.Gson;
 import io.milvus.bulkwriter.common.clientenum.ConnectType;
 import io.milvus.bulkwriter.common.utils.FileUtils;
 import io.milvus.bulkwriter.model.UploadFilesResult;
+import io.milvus.bulkwriter.model.UploadProgress;
 import io.milvus.bulkwriter.request.volume.ApplyVolumeRequest;
 import io.milvus.bulkwriter.request.volume.UploadFilesRequest;
 import io.milvus.bulkwriter.resolver.EndpointResolver;
@@ -31,38 +32,41 @@ import io.milvus.bulkwriter.restful.DataVolumeUtils;
 import io.milvus.bulkwriter.storage.StorageClient;
 import io.milvus.bulkwriter.storage.client.MinioStorageClient;
 import io.milvus.exception.ParamException;
+import io.minio.errors.ErrorResponseException;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.lang3.tuple.Pair;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.io.File;
+import java.io.IOException;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.time.Instant;
-import java.util.Date;
+import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
+import java.util.Locale;
+import java.util.Map;
+import java.util.Set;
 import java.util.concurrent.*;
-import java.util.concurrent.atomic.AtomicInteger;
-import java.util.concurrent.atomic.AtomicLong;
+import java.util.concurrent.atomic.AtomicReference;
 
 public class VolumeFileManager {
     private static final Logger logger = LoggerFactory.getLogger(VolumeFileManager.class);
+    private static final long DEFAULT_CREDENTIAL_REFRESH_MARGIN_SECONDS = 300L;
     private final String cloudEndpoint;
     private final String apiKey;
     private final String volumeName;
     private final ConnectType connectType;
-    private final ExecutorService executor;
 
-    private StorageClient storageClient;
-    private ApplyVolumeResponse applyVolumeResponse;
+    private volatile VolumeSession lastVolumeSession;
 
     public VolumeFileManager(VolumeFileManagerParam volumeFileManagerParam) {
         this.cloudEndpoint = volumeFileManagerParam.getCloudEndpoint();
         this.apiKey = volumeFileManagerParam.getApiKey();
         this.volumeName = volumeFileManagerParam.getVolumeName();
-        this.connectType = volumeFileManagerParam.getConnectType();
-        this.executor = Executors.newFixedThreadPool(10);
+        this.connectType = volumeFileManagerParam.getConnectType() == null ? ConnectType.AUTO : volumeFileManagerParam.getConnectType();
     }
 
     /**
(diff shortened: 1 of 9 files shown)


sdk-bulkwriter/src/main/java/io/milvus/bulkwriter/VolumeFileManager.java
```diff
@@ -78,77 +80,136 @@ public CompletableFuture<UploadFilesResult> uploadFilesAsync(UploadFilesRequest
...
         Pair<List<String>, Long> localPathPair = FileUtils.processLocalPath(localDirOrFilePath);
         String volumePath = convertDirPath(request.getTargetVolumePath());
-
-        refreshVolumeAndClient(volumePath);
-        initValidator(localPathPair);
-
-        AtomicInteger currentFileCount = new AtomicInteger(0);
-        AtomicLong processedBytes = new AtomicLong(0);
+        int uploadConcurrency = Math.max(1, request.getUploadConcurrency());
+        int maxRetries = Math.max(1, request.getMaxRetries());
```
yhmo: sdk-bulkwriter/src/main/java/io/milvus/bulkwriter/VolumeFileManager.java line:84
Medium ---- `UploadFilesRequest.maxRetries` is documented as the caller's retry count, but this clamp turns `0` into `1`, so there is no way to disable retries. That changes the public contract for a malformed-but-type-valid input and means a caller who explicitly asks for "try once" still gets an extra retry. Could we preserve `0` here (for no retries) and only reject negative values if we want validation?

sdk-bulkwriter/src/main/java/io/milvus/bulkwriter/storage/client/MinioStorageClient.java
```diff
@@ -130,14 +137,61 @@ public Long getObjectEntity(String bucketName, String objectKey) throws Exceptio
...
+                httpClient.cache().close();
+            } catch (IOException e) {
+                logger.warn("Failed to close MinIO HTTP client cache", e);
+            }
+        }
+    }
+
+    static long calculateUploadPartSize(long fileSize, long requestedPartSizeBytes) {
+        if (requestedPartSizeBytes > 0) {
+            return requestedPartSizeBytes;
```
yhmo, line 186: sdk-bulkwriter/src/main/java/io/milvus/bulkwriter/storage/client/MinioStorageClient.java line:182
Medium ---- `partSizeBytes` is exposed as a caller-controlled upload option, but this branch forwards any positive value directly to MinIO. Values below S3's 5 MiB multipart minimum are still type-valid here, so a caller can now trigger a runtime failure just by setting `partSizeBytes=1`. Could we clamp explicit values to the same minimum (or validate and throw a `ParamException`) before building `PutObjectArgs`?

sdk-bulkwriter/src/main/java/io/milvus/bulkwriter/VolumeFileManager.java
```diff
@@ -217,65 +307,337 @@ private void uploadLocalFileToVolume(String localFilePath, String rootPath, Stri
...
     }
 
-    private void putObjectWithRetry(File file, String remoteFilePath, String volumePath) {
-        refreshIfExpire(volumePath);
+    private void putObjectWithRetry(File file, String remoteFilePath, String volumePath,
+                                    int maxRetries, long retryIntervalMillis,
+                                    UploadProgressTracker progressTracker,
+                                    UploadContext uploadContext, long partSizeBytes) {
         String msg = "upload " + file.getAbsolutePath();
+        FileUploadProgress progress = new FileUploadProgress(progressTracker, file.getAbsolutePath(), file.length());
```
liliu-z, line 323: The byte-progress map is keyed inconsistently (the reviewer has the direction reversed). Per-file streaming updates and reset use file.getAbsolutePath() (FileUploadProgress at line 320, driving updateFile/resetFile), but finishFile is called at line 120 with the raw localFilePath from the file list. FileUtils.processLocalPath returns path.toString() for a single regular file (FileUtils.java:21), so when the source is a single file given as a relative path, localFilePath is relative while new File(localFilePath).getAbsolutePath() is absolute; the two keys diverge and finishFile re-adds the whole file size to uploadedBytes (double counting). Directory inputs are unaffected because FileUtils returns absolute paths there. There is no NPE and the file is still marked complete; the only effect is inflated uploadedBytes/speed in the progress logs and listener (percent is clamped to 100). Use one canonical absolute key in both the streaming and finish paths.

sdk-bulkwriter/src/main/java/io/milvus/bulkwriter/VolumeFileManager.java
```diff
@@ -217,65 +307,337 @@ private void uploadLocalFileToVolume(String localFilePath, String rootPath, Stri
...
+    private <T> T withRetry(String actionName, Callable<T> callable, String volumePath,
+                            int maxRetries, long retryIntervalMillis,
+                            UploadContext uploadContext) {
         int attempt = 0;
         while (attempt < maxRetries) {
             try {
                 return callable.call();
-            } catch (RuntimeException e) {
-                throw e;
             } catch (Exception e) {
```
liliu-z, line 369: withRetry catches every Exception (except a progress-callback failure) and retries it up to maxRetries, recreating the volume session/credentials and sleeping retryIntervalMillis on each attempt, with no transient-vs-terminal classification. Deterministic failures (the out-of-range part-size IllegalArgumentException above, auth/permission denied, 4xx, invalid argument) are therefore retried the full budget and refresh credentials pointlessly before failing. This is a behavior change introduced by this PR: the previous code rethrew RuntimeExceptions immediately. The root cause is still attached to the final RuntimeException (so it is not 'masked behind a timeout'), but the failure is delayed by maxRetries × (credential refresh + retryIntervalMillis) and costs extra apply-volume calls. Classify errors and retry only transient/retryable ones.

sdk-bulkwriter/src/main/java/io/milvus/bulkwriter/VolumeFileManager.java
```diff
@@ -172,40 +233,69 @@ private void initValidator(Pair<List<String>, Long> localPathPair) {
...
-        logger.info("storage client refreshed");
     }
 
     private String convertDirPath(String inputPath) {
         if (StringUtils.isEmpty(inputPath) || inputPath.equals("/")) {
             return "";
         }
-        if (inputPath.endsWith("/")) {
-            return inputPath;
+        String[] parts = inputPath.replace("\\", "/").split("/");
```
liliu-z, line 278: convertDirPath operates on the target volume (remote) path, not the local directory as the reviewer states. The rewrite changes how some target paths map to the remote prefix: leading slashes, '.' segments, and redundant or backslash separators are stripped/collapsed (e.g. '/data' and './data' both become 'data/' rather than the old '/data/' and './data/'), and '..' now throws ParamException instead of being passed through. Callers who relied on the old mapping will see uploaded objects land under a different remote key. The new normalization is reasonable, but since the PR states the public APIs remain source-compatible, this runtime behavior change should be called out in the release notes and kept consistent with the single-file relative-path handling.

sdk-bulkwriter/src/main/java/io/milvus/bulkwriter/VolumeFileManager.java
lentitude2tk: Fixed in 0644b73f. `maxRetries` now preserves `0`, so callers can disable retries and get only the initial upload attempt. Negative values are still normalized to `0`.

sdk-bulkwriter/src/main/java/io/milvus/bulkwriter/storage/client/MinioStorageClient.java
lentitude2tk: Fixed in 0644b73f. Explicit `partSizeBytes` below the S3 multipart minimum now fails fast with `ParamException` before building MinIO `PutObjectArgs`.

sdk-bulkwriter/src/main/java/io/milvus/bulkwriter/VolumeFileManager.java
lentitude2tk: Fixed in 0644b73f. `finishFile` now uses the same canonical absolute path key as streaming progress/reset, and I added a relative single-file regression test to ensure uploaded bytes are not double counted.

sdk-bulkwriter/src/main/java/io/milvus/bulkwriter/VolumeFileManager.java
lentitude2tk: Fixed in 0644b73f. Retry now classifies failures: progress callback failures, `ParamException`/`IllegalArgumentException`, and non-retryable S3 4xx errors fail immediately; IO/timeout and retryable S3 throttling/5xx errors still use the retry budget. Added tests for `maxRetries=0` and non-retryable failures.

sdk-bulkwriter/src/main/java/io/milvus/bulkwriter/VolumeFileManager.java
lentitude2tk: Leaving this behavior unchanged. `convertDirPath` is intentionally normalizing the target Volume path, not the local source path: leading slashes, duplicate separators, and `.` segments collapse to a stable remote prefix, and `..` is rejected to avoid escaping the Volume root. This matches the PyMilvus behavior in the paired PR.

PR #1909, opened 2026-06-23
Add highlighter example

yhmo:


examples/src/main/java/io/milvus/v2/HighlighterExample.java
```diff
@@ -0,0 +1,183 @@
...
+                .addPreTag("<em>")
+                .addPostTag("</em>")
+                .fragmentSize(40)
+                .numOfFragments(10)
+                .build();
+
+        SearchResp searchResp = client.search(SearchReq.builder()
+                .collectionName(COLLECTION_NAME)
+                .data(Collections.singletonList(new EmbeddedText(queryText)))
+                .filter(String.format("TEXT_MATCH(%s, \"%s\")", TEXT_FIELD, queryText))
```
liliu-z: The example inserts rows and immediately runs searchWithHighlighter(), whose filter is TEXT_MATCH(text, "Milvus users"), without ever flushing the collection. The sibling TextMatchExample flushes right before its TEXT_MATCH queries with the comment "TEXT_MATCH requires the data is persisted" and does so even after a STRONG-consistency query, so STRONG consistency alone won't make the just-inserted rows matchable; the filter can exclude every row, leaving the search empty and the highlighter printing no fragments or scores. Add a client.flush(...) on the collection after insert and before the search to mirror TextMatchExample.

yhmo: examples/src/main/java/io/milvus/v2/HighlighterExample.java line:146
Medium ---- This example uses TEXT_MATCH(...) immediately after insertData(), but nothing flushes the collection first. The existing TextMatchExample flushes before any TEXT_MATCH query/search because that operator only works on persisted data, so this sample can return empty results or miss highlights even though the inserted rows are visible to a STRONG count query.

PR #1918, opened 2026-06-23
Align search response with pymilvus

yhmo:


sdk-core/src/main/java/io/milvus/v2/service/vector/VectorService.java
```diff
@@ -265,11 +265,24 @@ public SearchResp search(MilvusServiceGrpc.MilvusServiceBlockingStub blockingStu
         SearchResults response = blockingStub.search(searchRequest);
         rpcUtils.handleResponse(title, response.getStatus());
 
-        return SearchResp.builder()
+        java.util.Map<String, String> extraInfo = response.getStatus().getExtraInfoMap();
```
yhmo: sdk-core/src/main/java/io/milvus/v2/service/vector/VectorService.java line:268
Medium ---- This populates the new `SearchResp` metrics for `search()`, but `hybridSearch()` still returns the same `SearchResp` type without copying any of these fields from `response.getStatus().getExtraInfoMap()`. After this PR, callers will get `cost`/`scanned*`/`cacheHitRatio` for ordinary search and silently lose them for hybrid search, which makes the shared response contract inconsistent with the PR's "align search response" goal.

sdk-core/src/main/java/io/milvus/v2/service/vector/VectorService.java
```diff
@@ -286,10 +287,26 @@ public SearchResp hybridSearch(MilvusServiceGrpc.MilvusServiceBlockingStub block
         SearchResults response = blockingStub.hybridSearch(searchRequest);
         rpcUtils.handleResponse(title, response.getStatus());
 
-        return SearchResp.builder()
+        SearchResp.SearchRespBuilder respBuilder = SearchResp.builder()
```
yhmo, line 290: sdk-core/src/main/java/io/milvus/v2/service/vector/VectorService.java line:290
Medium ---- Should `hybridSearch()` also copy `response.getSessionTs()` into the shared `SearchResp` here? `search()` still sets `sessionTs`, `SearchResp` still exposes `getSessionTs()`, and the builder default is `1L`, so hybrid search callers now silently lose the server timestamp and always see the fallback value instead.

sdk-core/src/main/java/io/milvus/v2/service/vector/VectorService.java
```diff
@@ -286,10 +287,27 @@ public SearchResp hybridSearch(MilvusServiceGrpc.MilvusServiceBlockingStub block
...
-                .build();
+                .sessionTs(response.getSessionTs())
+                .recalls(response.getResults().getRecallsList());
+        fillSearchRespFromExtraInfo(respBuilder, response.getStatus().getExtraInfoMap());
+        return respBuilder.build();
+    }
+
+    private void fillSearchRespFromExtraInfo(SearchResp.SearchRespBuilder respBuilder, java.util.Map<String, String> extraInfo) {
+        if (extraInfo.containsKey("report_value")) {
+            respBuilder.cost(Long.parseLong(extraInfo.get("report_value")));
```
liliu-z, line 300: fillSearchRespFromExtraInfo calls Long.parseLong/Float.parseFloat directly on server-supplied extra_info values with no guard. If the server returns a non-integer string (e.g. "123.0" for a Long field across an SDK/server version skew, or an empty value), the parse throws NumberFormatException, which propagates out of search()/hybridSearch() and fails an otherwise-successful search, discarding valid results. Wrap each parse in try/catch (or validate first) and skip/default the field instead of failing the whole response.

sdk-core/src/test/java/io/milvus/v2/BaseTest.java
```diff
@@ -138,6 +138,12 @@ public void setUp() {
         when(blockingStub.query(any())).thenReturn(QueryResults.newBuilder().build());
         when(blockingStub.delete(any())).thenReturn(MutationResult.newBuilder().setDeleteCnt(2L).build());
         SearchResults searchResults = SearchResults.newBuilder()
+                .setStatus(Status.newBuilder().setCode(0)
+                        .putExtraInfo("report_value", "123")
```
liliu-z, line 142: The test setup now injects report_value/scanned_remote_bytes/scanned_total_bytes/cache_hit_ratio into the mocked search status, but no unit test asserts the parsed results — VectorTest.testSearch only calls statusR.toString(). A regression in fillSearchRespFromExtraInfo (wrong key, wrong type, dropped value) would pass CI unnoticed. Assert SearchResp.getCost()/getScannedRemoteBytes()/getScannedTotalBytes()/getCacheHitRatio() against the injected values.

sdk-core/src/main/java/io/milvus/v2/service/vector/VectorService.java
yhmo: no need to fix

PR #1923, opened 2026-06-24
Support AddFunctionField/DropCollectionField/DropFunctionField

yhmo:


sdk-core/src/main/java/io/milvus/v2/service/collection/CollectionService.java
```diff
@@ -697,6 +731,40 @@ public Void addCollectionFunction(MilvusServiceGrpc.MilvusServiceBlockingStub bl
...
+        String title = String.format("Add function field to collection: '%s' in database: '%s'", collectionName, dbName);
+
+        CreateCollectionReq.FieldSchema fieldSchema = SchemaUtils.convertFieldReqToFieldSchema(request);
+        FieldSchema grpcFieldSchema = SchemaUtils.convertToGrpcFieldSchema(fieldSchema, true);
+        AlterCollectionSchemaRequest.FieldInfo fieldInfo = AlterCollectionSchemaRequest.FieldInfo.newBuilder()
+                .setFieldSchema(grpcFieldSchema)
+                .build();
+        AlterCollectionSchemaRequest.AddRequest addRequest = AlterCollectionSchemaRequest.AddRequest.newBuilder()
+                .addFieldInfos(fieldInfo)
+                .addFuncSchema(SchemaUtils.convertToGrpcFunction(request.getFunction()))
```
yhmo, line 760: sdk-core/src/main/java/io/milvus/v2/service/collection/CollectionService.java line:751
Medium ---- `addFunctionField` never validates that `request.getFieldName()` matches `request.getFunction().getOutputFieldNames()`. Because `CreateCollectionReq.Function` still allows any output-field list, a caller can pass `fieldName("sparse")` together with `outputFieldNames(["other"])` or even multiple outputs, and this method will serialize an inconsistent `AlterCollectionSchemaRequest` instead of rejecting it up front. Please enforce the single-output-field contract here (exactly one output name, equal to the field being added) and add a test for the mismatch case.

sdk-core/src/main/java/io/milvus/v2/service/collection/CollectionService.java
```diff
@@ -697,6 +731,40 @@ public Void addCollectionFunction(MilvusServiceGrpc.MilvusServiceBlockingStub bl
...
+        if (request.getFunction() == null) {
+            throw new MilvusClientException(ErrorCode.INVALID_PARAMS, "Function cannot be null.");
+        }
+
+        String dbName = request.getDatabaseName();
+        String collectionName = request.getCollectionName();
+        String title = String.format("Add function field to collection: '%s' in database: '%s'", collectionName, dbName);
+
+        CreateCollectionReq.FieldSchema fieldSchema = SchemaUtils.convertFieldReqToFieldSchema(request);
+        FieldSchema grpcFieldSchema = SchemaUtils.convertToGrpcFieldSchema(fieldSchema, true);
```
liliu-z: addFunctionField converts the output field with convertToGrpcFieldSchema(fieldSchema, true); the forAddField flag makes the shared converter reject non-nullable vector fields (SchemaUtils.java:64-67), but Milvus rejects nullable function-output fields ('function output field cannot be nullable'). A BM25 SparseFloatVector output therefore has no valid request shape — non-nullable is blocked by the SDK before the RPC and nullable is blocked by the server — so the headline use case cannot be added. testFunctionFieldLifecycle sends isNullable(true) and would fail against a real server. Use a function-output conversion path that permits non-nullable vectors, as createCollectionWithSchema already does (it converts output fields with the 1-arg, non-forAddField, converter).

sdk-core/src/main/java/io/milvus/v2/service/collection/CollectionService.java
```diff
@@ -740,6 +808,32 @@ public Void dropCollectionFunction(MilvusServiceGrpc.MilvusServiceBlockingStub b
...
+
+        String dbName = request.getDatabaseName();
+        String collectionName = request.getCollectionName();
+        String title = String.format("Drop function field of collection: '%s' in database: '%s'", collectionName, dbName);
+        AlterCollectionSchemaRequest.Builder builder = AlterCollectionSchemaRequest.newBuilder()
+                .setCollectionName(collectionName)
+                .setAction(AlterCollectionSchemaRequest.Action.newBuilder()
+                        .setDropRequest(AlterCollectionSchemaRequest.DropRequest.newBuilder()
+                                .setFunctionName(request.getFunctionName())
+                                .setDropFunctionOutputFields(true)
```
liliu-z, line 840: dropFunctionField sends DropRequest.setDropFunctionOutputFields(true) (line 825), destructively dropping the function's output field(s), whereas dropCollectionFunction (lines 790-809) calls the separate dropCollectionFunction RPC with no such flag and leaves output fields intact. This divergence is not noted in the Javadoc, so a caller expecting dropCollectionFunction semantics can unexpectedly lose columns. Document the destructive behavior, or expose dropFunctionOutputFields as a parameter.

sdk-core/src/main/java/io/milvus/v2/service/collection/CollectionService.java
liliu-z: addFunctionField builds the output field via convertToGrpcFieldSchema (line 745) and sends it as-is, never calling setIsFunctionOutput(true). createCollectionWithSchema deliberately marks every function-output field with setIsFunctionOutput(true) (line 150), keyed off the function's output field names — meaning the server relies on that flag to recognize a field as function-generated. Without it, the field added here is transmitted as an ordinary vector field, so the server may reject the request or treat the column as user-supplied rather than function-populated. Set is_function_output=true on the field, as createCollection does.

PR #1927, opened 2026-06-25
Add code coverage check

yhmo:


.github/workflows/maven.yml
```diff
@@ -35,4 +35,11 @@ jobs:
       - name: Update submodule
         run: git submodule update --init
       - name: Build with Maven
-        run: mvn --batch-mode --update-snapshots verify
\ No newline at end of file
+        run: mvn -Pcoverage --batch-mode --update-snapshots verify
+      - name: Upload coverage
```
yhmo, line 39: High ---- This still does not make coverage a required check for the PR. The workflow only uploads reports here, and codecov-action@v5 leaves fail_ci_if_error at its default false, so a broken or missing upload would not fail Build and test. More importantly, this repo's merge automation is still keyed off the Build and test check rather than the new codecov/project or codecov/patch statuses, so the PR can still get the normal CI-passed path even if Codecov later reports below-threshold coverage. In other words, the change adds coverage reporting, but it does not actually enforce the coverage check promised by the PR title. Please either wire the Codecov statuses into the required merge checks or make this workflow fail when the coverage gate is not satisfied.

PR #1932, opened 2026-06-26
BulkWriter support binary/float16/int8 vectors in Struct field

yhmo:


sdk-bulkwriter/src/main/java/io/milvus/bulkwriter/writer/CSVFileWriter.java
```diff
@@ -40,9 +39,36 @@ private void initWriter() throws IOException {
...
+            byte[] bytes = byteBuffer.array();
+            List<Integer> result = new ArrayList<>(bytes.length);
+            for (byte b : bytes) {
+                result.add((int) b);
+            }
+            return result;
+        }
+        if (value instanceof List) {
+            List<?> values = (List<?>) value;
+            List<Object> normalized = new ArrayList<>(values.size());
```
liliu-z: normalizeValue's `instanceof List` branch unconditionally allocates a new ArrayList and re-adds every element on every row, so a plain FloatVector (List<Float>, whose elements are returned unchanged) is now shallow-copied per appendRow where before the CSV writer passed the list straight to gson. The copy is dominated by the subsequent gson.toJson, so the impact is minor, but it is an avoidable per-row allocation: initialize `normalized` lazily and only once a child element actually differs, letting unchanged vectors pass through without copying.

sdk-bulkwriter/src/main/java/io/milvus/bulkwriter/writer/JSONFileWriter.java
```diff
@@ -34,11 +36,37 @@ private void initWriter() throws IOException {
...
+            byte[] bytes = byteBuffer.array();
+            List<Integer> result = new ArrayList<>(bytes.length);
+            for (byte b : bytes) {
+                result.add((int) b);
+            }
+            return result;
+        }
+        if (value instanceof List) {
+            List<?> values = (List<?>) value;
+            List<Object> normalized = new ArrayList<>(values.size());
```
liliu-z: Same avoidable per-row copy as the CSV writer: normalizeValue's `instanceof List` branch always builds a new ArrayList and re-adds each element, so float/float16 vectors (List<Float>) and array fields are shallow-copied on every appendRow. The previous `rowValues.replaceAll(... value instanceof ByteBuffer ? ((ByteBuffer) value).array() : value)` left lists untouched. Allocate `normalized` lazily only after a child element actually changes so unchanged lists pass through.

sdk-bulkwriter/src/main/java/io/milvus/bulkwriter/writer/CSVFileWriter.java
```diff
@@ -40,9 +39,36 @@ private void initWriter() throws IOException {
         this.writer = new BufferedWriter(new java.io.FileWriter(filePath));
     }
 
+    private static Object normalizeValue(Object value) {
```
liliu-z: The new ~25-line `normalizeValue` helper is duplicated byte-for-byte in CSVFileWriter (lines 42-66) and JSONFileWriter (lines 39-63). Extract it to a shared utility (e.g. under bulkwriter/common/utils) so the two writers cannot diverge; concretely, the lazy-allocation fix for the per-row copy flagged above would otherwise have to be made identically in two places.

PR #1933, opened 2026-06-26
BulkWriter support binary/float16/int8 vectors in Struct field

yhmo:


sdk-bulkwriter/src/main/java/io/milvus/bulkwriter/storage/client/MinioStorageClient.java
```diff
@@ -167,8 +169,18 @@ public void close() {
         if (!closeHttpClient || httpClient == null) {
             return;
         }
-        httpClient.dispatcher().executorService().shutdown();
+        ExecutorService executorService = httpClient.dispatcher().executorService();
+        executorService.shutdown();
         httpClient.connectionPool().evictAll();
+        try {
+            if (!executorService.awaitTermination(5, TimeUnit.SECONDS)) {
```
yhmo, line 176: This adds meaningful shutdown behavior (`awaitTermination`, `shutdownNow`, interrupt restoration, cache close) but the PR doesn't add any coverage for it. Since this path only runs during cleanup, regressions here are easy to miss until users start leaking executor threads or hanging on close. Could we add a focused test around `MinioStorageClient.close()` (and ideally the `RemoteBulkWriter.close()` finally-path) to prove both the owned-client and injected-client branches behave as intended?

PR #1936, opened 2026-07-08
Support aggregation for search()

yhmo:


sdk-core/src/main/java/io/milvus/v2/service/vector/response/aggregation/AggregationHit.java
```diff
@@ -0,0 +1,109 @@
...
+        this.score = builder.score;
+        this.fields = builder.fields;
+        this.fieldIds = builder.fieldIds;
+    }
+
+    public static Builder builder() {
+        return new Builder();
+    }
+
+    public Object getPrimaryKey() {
```
yhmo: `SearchResp.SearchResult` uses `getPrimaryKey()` for the primary-key *field name* and `getId()` for the returned value, but this new type flips that meaning: `getPrimaryKey()` now returns the value itself. Because there are no examples/docs for the aggregation response yet, callers will mostly learn it from these accessors, and this inconsistency is likely to produce wrong integrations or awkward adapter code. Could we align this with the existing search result contract (for example, expose the hit value as `getId()` / `getPrimaryKeyValue()` and reserve `getPrimaryKey()` for the field name)?

sdk-core/src/main/java/io/milvus/v2/utils/ConvertUtils.java
```diff
@@ -141,6 +143,140 @@ public List<List<SearchResp.SearchResult>> getEntities(SearchResults response) {
...
     }
 
+    public List<List<AggregationBucket>> getAggregationBuckets(SearchResults response) {
+        List<AggregationBucket> buckets = new ArrayList<>();
+        for (AggBucket bucket : response.getResults().getAggBucketsList()) {
+            buckets.add(convertAggregationBucket(bucket));
+        }
+
+        List<Long> aggTopks = response.getResults().getAggTopksList();
+        if (aggTopks.isEmpty()) {
```
yhmo, line 157: sdk-core/src/main/java/io/milvus/v2/utils/ConvertUtils.java line:153
High ---- When aggTopks is empty this still returns a single outer entry containing the flat buckets list, so a normal search with no aggregation payload comes back as `[[]]` instead of `[]`. That silently changes the default `SearchResp.getAggregationBuckets()` contract for existing callers: code that currently treats `isEmpty()` as “no aggregation data” will start seeing a non-empty result on ordinary searches.

sdk-core/src/main/java/io/milvus/v2/service/vector/request/aggregation/MetricSpec.java
```diff
@@ -0,0 +1,125 @@
...
+        }
+        this.op = builder.op;
+        this.fieldName = builder.fieldName;
+    }
+
+    private MetricSpec(LegacyBuilder builder) {
+        this(new Builder().op(MetricOps.fromValue(builder.op)).fieldName(builder.fieldName));
+    }
+
+    public static LegacyBuilder builder() {
```
liliu-z: `MetricSpec.builder()` returns the string-based `LegacyBuilder`, whereas the sibling factories `SearchAggregation.builder()` and `TopHitsSpec.builder()` return the type-safe enum `Builder`. This is brand-new public API, so the inconsistent entry point gets locked in and can only be corrected later with a return-type-breaking change. Point `builder()` at the enum `Builder` and give the string variant a distinct factory name.

sdk-core/src/main/java/io/milvus/v2/service/vector/request/aggregation/OrderSpec.java
```diff
@@ -0,0 +1,148 @@
...
+        if (builder.key == null || builder.key.isEmpty()) {
+            throw new MilvusClientException(ErrorCode.INVALID_PARAMS,
+                    "SearchAggregation.order key must not be empty.");
+        }
+        this.key = builder.key;
+        this.direction = AggDirection.fromValue(builder.direction, "SearchAggregation.order direction");
+        this.nullFirst = builder.nullFirst;
+    }
+
+    public static LegacyBuilder builder() {
```
liliu-z: `OrderSpec.builder()` returns the string-based `LegacyBuilder`, inconsistent with the sibling aggregation specs whose `builder()` returns the enum `Builder` (`SearchAggregation`, `TopHitsSpec`). As new public API this locks in the divergence and is a breaking return-type change to fix after release; have `builder()` return the enum `Builder`.

sdk-core/src/main/java/io/milvus/v2/service/vector/request/aggregation/SortSpec.java
```diff
@@ -0,0 +1,148 @@
...
+        if (builder.fieldName == null || builder.fieldName.isEmpty()) {
+            throw new MilvusClientException(ErrorCode.INVALID_PARAMS,
+                    "TopHitsSpec.sort fieldName cannot be empty.");
+        }
+        this.fieldName = builder.fieldName;
+        this.direction = AggDirection.fromValue(builder.direction, "TopHitsSpec.sort direction");
+        this.nullFirst = builder.nullFirst;
+    }
+
+    public static LegacyBuilder builder() {
```
liliu-z: `SortSpec.builder()` returns the string-based `LegacyBuilder`, unlike the sibling aggregation specs whose `builder()` returns the enum `Builder`. Shipping this locks the inconsistency into public API and makes a later correction a breaking return-type change; return the enum `Builder` from `builder()`.

sdk-core/src/main/java/io/milvus/v2/utils/ConvertUtils.java
```diff
@@ -141,6 +143,144 @@ public List<List<SearchResp.SearchResult>> getEntities(SearchResults response) {
...
+        for (AggBucket bucket : response.getResults().getAggBucketsList()) {
+            buckets.add(convertAggregationBucket(bucket));
+        }
+
+        if (buckets.isEmpty()) {
+            return new ArrayList<>();
+        }
+
+        List<Long> aggTopks = response.getResults().getAggTopksList();
+        if (aggTopks.isEmpty()) {
```
yhmo, line 157: sdk-core/src/main/java/io/milvus/v2/utils/ConvertUtils.java line:157
Medium ---- When the server omits aggTopks for a multi-query search, this fallback collapses every bucket into a single outer entry instead of preserving the query count. That makes SearchResp.getAggregationBuckets() disagree with the sibling PyMilvus contract for the same wire payload and can silently mis-assign buckets to the wrong query vector for callers that correlate results by nq.

sdk-core/src/main/java/io/milvus/v2/utils/ConvertUtils.java
```diff
@@ -141,6 +143,153 @@ public List<List<SearchResp.SearchResult>> getEntities(SearchResults response) {
...
+            default:
+                return null;
+        }
+    }
+
+    private Object getAggHitFieldValue(AggHitField field) {
+        switch (field.getValueCase()) {
+            case INT_VAL:
+                return field.getIntVal();
+            case BOOL_VAL:
```
liliu-z, line 243: getAggHitFieldValue returns field.getBytesVal() directly, so a bytes-typed aggregation hit field lands in the user-facing AggregationHit.fields map as a com.google.protobuf.ByteString. Everywhere else the SDK converts protobuf bytes to a plain Java type before exposing it — FieldDataWrapper maps JSON bytes with ByteString.toStringUtf8() and binary vectors with toByteArray()/ByteBuffer — so a caller reading a bytes-valued hit field here gets an opaque protobuf type and will hit a ClassCastException if they cast to byte[] or String. Convert with toByteArray() (or toStringUtf8()) before storing.

sdk-core/src/main/java/io/milvus/v2/utils/ConvertUtils.java
```diff
@@ -141,6 +143,152 @@ public List<List<SearchResp.SearchResult>> getEntities(SearchResults response) {
...
+        for (AggBucket bucket : response.getResults().getAggBucketsList()) {
+            buckets.add(convertAggregationBucket(bucket));
+        }
+
+        if (buckets.isEmpty()) {
+            return new ArrayList<>();
+        }
+
+        List<Long> aggTopks = response.getResults().getAggTopksList();
+        if (aggTopks.isEmpty()) {
```
yhmo, line 157: sdk-core/src/main/java/io/milvus/v2/utils/ConvertUtils.java line:157
Medium ---- When the server omits aggTopks but still returns agg_buckets for a multi-query search, this fallback replaces the real bucket payload with one empty list per query. schema.proto defines agg_buckets and agg_topks as independent fields, so callers can legitimately receive buckets without per-query counts; in that case SearchResp.getAggregationBuckets() silently drops valid aggregation data instead of preserving it or surfacing the ambiguity.

examples/src/main/java/io/milvus/v2/SearchAggregationExample.java
```diff
@@ -0,0 +1,483 @@
...
+                .uri("http://localhost:19530")
+                .build();
+        MilvusClientV2 client = new MilvusClientV2(config);
+        try {
+            buildCollection(client);
+
+            case1SingleField(client);
+            case2CompositeKeyWithMetrics(client);
+            System.out.println("\n=== Case 3: JSON path grouping ===");
+            System.out.println("Skipped. The Java SDK currently does not allow bracketed JSON-path fields in SearchAggregation.fields.");
```
yhmo: examples/src/main/java/io/milvus/v2/SearchAggregationExample.java line:101
Low ---- This example says the Java SDK cannot use bracketed JSON-path fields in SearchAggregation.fields, but the shipped request path just forwards each field string into SearchAggregationSpec (common.proto even documents JSON path expression OK there). Leaving the example in a skipped state teaches callers that a supported contract is unavailable, so they may never try the JSON-path form or may cargo-cult the limitation into their own docs.

sdk-core/src/main/java/io/milvus/v2/utils/ConvertUtils.java
yhmo, line 157: sdk-core/src/main/java/io/milvus/v2/utils/ConvertUtils.java line:157
Medium ---- Could we avoid treating missing aggTopks as a single-query result here? SearchResp.getAggregationBuckets() is shaped per query, so when nq > 1 and the server omits aggTopks this fallback silently collapses every bucket into one outer entry and callers can no longer tell which query vector each bucket belongs to.

sdk-core/src/main/java/io/milvus/v2/utils/VectorUtils.java
```diff
@@ -319,7 +320,12 @@ public SearchRequest ConvertToGrpcSearchRequest(SearchReq request) {
                             .build());
         }
 
+        SearchAggregation searchAggregation = request.getSearchAggregation();
         if (request.getGroupByFieldName() != null && !request.getGroupByFieldName().isEmpty()) {
```
yhmo, line 324: sdk-core/src/main/java/io/milvus/v2/utils/VectorUtils.java line:324
Medium ---- Could we also guard the legacy raw searchParams path here? compatibleSearchParams(...) has already copied any caller-supplied group_by_field/group_size/strict_group_size keys into the request before this branch runs, so a request that sets searchAggregation plus those legacy params still sends both grouping APIs together even though the typed groupByFieldName setter is rejected.

PR #1937, opened 2026-07-10
Support OrderBy for search/query

yhmo:


sdk-core/src/main/java/io/milvus/v2/utils/VectorUtils.java
```diff
@@ -155,6 +164,12 @@ private static long getGuaranteeTimestamp(ConsistencyLevel consistencyLevel, Str
         }
     }
 
+    private static String formatOrderByFields(List<OrderByField> orderByFields) {
+        return orderByFields.stream()
+                .map(item -> item.getFieldName() + ":" + item.getDirection().getValue())
```
liliu-z, line 169: formatOrderByFields serializes each entry as `fieldName:direction` and joins entries with a comma, escaping neither delimiter. Ordinary collection field names cannot contain `,` or `:`, but this PR also supports ordering by JSON/dynamic-field paths (the example uses metadata["age"]); a JSON key that itself contains a comma or colon produces an ambiguous order-by string the server can misparse or reject. Escape the delimiters, or use a separator that cannot appear in a JSON-path key.

PR #1940, opened 2026-07-10
Support minhash function

yhmo:


sdk-core/src/test/java/io/milvus/docker-compose.yml
```diff
@@ -34,7 +34,7 @@ services:
 
   standalone:
     container_name: milvus-javasdk-standalone
-    image: milvusdb/milvus:master-20260609-1853a148
+    image: milvusdb/milvus:master-20260622-a6568ac2-amd64
```
yhmo, line 37: sdk-core/src/test/java/io/milvus/docker-compose.yml line:37
Medium ---- Pinning the test container to the `-amd64` image makes this test setup fail on arm64 runners and developer machines even though the previous tag was multi-arch. That turns the MinHash coverage on this PR into an architecture-specific path; could we keep a multi-arch tag here (or gate the platform explicitly) so the Docker tests remain runnable across supported environments?

sdk-core/src/main/java/io/milvus/v2/utils/SchemaUtils.java
```diff
@@ -321,7 +321,7 @@ public static CreateCollectionReq.Function convertFromGrpcFunction(FunctionSchem
         CreateCollectionReq.Function.FunctionBuilder builder = CreateCollectionReq.Function.builder()
                 .name(functionSchema.getName())
                 .description(functionSchema.getDescription())
-                .functionType(io.milvus.common.clientenum.FunctionType.fromName(functionSchema.getType().name()))
+                .functionType(io.milvus.common.clientenum.FunctionType.fromCode(functionSchema.getType().getNumber()))
```
liliu-z: For a function code this client's pinned proto doesn't know (a newer server), `functionSchema.getType()` returns the proto3 `UNRECOGNIZED` enum constant, and calling `.getNumber()` on that constant throws `IllegalArgumentException` — so `describeCollection` now fails outright for the whole collection. The previous code used `getType().name()` → `fromName(...)`, where `name()` returns "UNRECOGNIZED" and `fromName` returns `null`, degrading gracefully. Fix by reading the raw int with `functionSchema.getTypeValue()` and passing it to `fromCode(...)`, which already returns `null` for unknown codes. (Note: the throw is on `.getNumber()`, not on `getType()` itself.)

sdk-core/src/main/java/io/milvus/common/clientenum/FunctionType.java
```diff
@@ -21,9 +21,11 @@
 
 public enum FunctionType {
     UNKNOWN("Unknown", 0), // in milvus-proto, the name is "Unknown"
-    BM25("BM25", 1), // Added missing name parameter
+    BM25("BM25", 1),
     TEXTEMBEDDING("TextEmbedding", 2), // in milvus-proto, the name is "TextEmbedding"
-    RERANK("RERANK", 3); // Added missing name parameter
+    RERANK("Rerank", 3),
```
yhmo, line 26: sdk-core/src/main/java/io/milvus/common/clientenum/FunctionType.java line:26
Medium ---- Switching `RERANK` from `"RERANK"` to `"Rerank"` and making `fromName()` coerce unknown strings to `UNKNOWN` changes the public `FunctionType` contract for existing callers, even though the MinHash fix only needs the new `fromCode()` path in `SchemaUtils`. Any caller that serialized `getName()`, looked up `fromName("RERANK")`, or used `null` to detect unsupported names will now behave differently; could we keep the old `getName()/fromName()` semantics and limit this PR to adding `MINHASH`, `MOLFINGERPRINT`, and `fromCode()`?

PR #1942, opened 2026-07-13
Support commitImport/abortImport

yhmo:


sdk-bulkwriter/src/main/java/io/milvus/bulkwriter/restful/BulkImportUtils.java
```diff
@@ -29,40 +29,40 @@
...
+    }
 
-        Map<String, Object> params = JsonUtils.fromJson(JsonUtils.toJson(request), new TypeToken<Map<String, Object>>() {
-        }.getType());
-        String body = postRequest(requestURL, request.getApiKey(), params, 60 * 1000);
-        RestfulResponse<Object> response = JsonUtils.fromJson(body, new TypeToken<RestfulResponse<Object>>() {
-        }.getType());
-        handleResponse(requestURL, response);
-        return body;
+    public static String commitImport(String url, BaseDescribeImportRequest request) {
```
yhmo, line 59: sdk-bulkwriter/src/main/java/io/milvus/bulkwriter/restful/BulkImportUtils.java line:59
Low ---- This adds public commit/abort wrappers for import jobs, but there is no Java-side test or example that exercises either endpoint or asserts the serialized jobId/cloud identifiers. A small test similar to the existing get-import-progress coverage would make endpoint typos or request-shape regressions visible before release.

PR #1956, opened 2026-07-16
Support Text data type

lentitude2tk:
## What
- expose `DataType.Text` in the V2 Java SDK
- align scalar Text value handling with PyMilvus: use `string_data` without VARCHAR `max_length` validation
- preserve PyMilvus-compatible protocol encoding/decoding for Text array and struct sub-fields while leaving schema capability validation to Milvus
- strictly validate every JSON `Array<Text>` element as a non-null string
- support scalar Text defaults, query/search decoding, and BulkWriter JSON/CSV/Parquet flows

## Tests
- `mvn -pl sdk-core -Dtest=TextDataTypeTest test`
- `mvn -pl sdk-bulkwriter -am -Dtest=BulkWriterTest -Dsurefire.failIfNoSpecifiedTests=false test`

### DIFF
--- a/sdk-bulkwriter/src/main/java/io/milvus/bulkwriter/BulkWriter.java
+++ b/sdk-bulkwriter/src/main/java/io/milvus/bulkwriter/BulkWriter.java
@@ -371,6 +371,7 @@ public abstract class BulkWriter implements AutoCloseable {
                 return verifyVector(obj, field);
             }
             case VarChar:
+            case Text:
             case Geometry:
             case Timestamptz: {
                 return verifyVarchar(obj, field);
--- a/sdk-bulkwriter/src/test/java/io/milvus/bulkwriter/BulkWriterTest.java
+++ b/sdk-bulkwriter/src/test/java/io/milvus/bulkwriter/BulkWriterTest.java
@@ -544,6 +544,109 @@ public class BulkWriterTest {
         }
     }
 
+    @Test
+    void testWriteTextFields() {
+        CreateCollectionReq.CollectionSchema schema = CreateCollectionReq.CollectionSchema.builder().build();
+        schema.addField(AddFieldReq.builder()
+                .fieldName("id")
+                .dataType(DataType.Int64)
+                .isPrimaryKey(true)
+                .build());
+        schema.addField(AddFieldReq.builder()
+                .fieldName("body")
+                .dataType(DataType.Text)
+                .maxLength(1)
+                .build());
+        String longText = String.join("", Collections.nCopies(1024, "text"));
+        JsonObject row = new JsonObject();
+        row.addProperty("id", 1L);
+        row.addProperty("body", longText);
+
+        for (BulkFileType fileType : Arrays.asList(BulkFileType.JSON, BulkFileType.CSV, BulkFileType.PARQUET)) {
+            LocalBulkWriterParam param = LocalBulkWriterParam.newBuilder()
+                    .withCollectionSchema(schema)
+                    .withLocalPath("/tmp/bulk_writer_text")
+                    .withFileType(fileType)
+                    .build();
+            try (LocalBulkWriter writer = new LocalBulkWriter(param)) {
+                writer.appendRow(row);
+                writer.commit(false);
+                List<List<String>> batchFiles = writer.getBatchFiles();
+                Assertions.assertFalse(batchFiles.isEmpty());
+                String filePath = batchFiles.get(0).get(0);
+                if (fileType == BulkFileType.JSON) {
+                    try (Reader reader = Files.newBufferedReader(Paths.get(filePath))) {
+                        List<JsonObject> rows = new Gson().fromJson(reader,
+                                new TypeToken<List<JsonObject>>() {
+                                }.getType());
+                        Assertions.assertEquals(1, rows.size());
+                        Assertions.assertEquals(longText, rows.get(0).get("body").getAsString());
+                    }
+                } else if (fileType == BulkFileType.CSV) {
+                    List<String> lines = Files.readAllLines(Paths.get(filePath), StandardCharsets.UTF_8);
+                    Assertions.assertEquals(2, lines.size());
+                    List<String> header = Arrays.asList(lines.get(0).split(",", -1));
+                    String[] values = lines.get(1).split(",", -1);
+                    String body = values[header.indexOf("body")];
+                    Assertions.assertEquals(longText, body.substring(1, body.length() - 1));
+                } else {
+                    final int[] rowsRead = {0};
+                    new ParquetReaderUtils() {
+                        @Override
+                        public void readRecord(GenericData.Record record) {
+                            rowsRead[0]++;
+                            Assertions.assertEquals(longText, record.get("body").toString());
+                        }
+                    }.readParquet(filePath);
+                    Assertions.assertEquals(1, rowsRead[0]);
+                }
+            } catch (Exception e) {
+                Assertions.fail(e.getMessage());
+            }
+        }
+    }
+
+    @Test
+    void testWriteTextFieldsWithV1Schema() {
+        CollectionSchemaParam schema = CollectionSchemaParam.newBuilder()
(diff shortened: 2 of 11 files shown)


sdk-core/src/main/java/io/milvus/param/ParamUtils.java
```diff
@@ -474,6 +491,7 @@ public static List<Object> convertJsonArray(JsonArray jsonArray, DataType elemen
                     return JsonUtils.fromJson(jsonArray, new TypeToken<List<Double>>() {
                     }.getType());
                 case VarChar:
+                case Text:
```
yhmo: sdk-core/src/main/java/io/milvus/param/ParamUtils.java line:494
Medium ---- `Array<Text>` currently delegates to Gson's `List<String>` adapter, which converts numeric and boolean JSON elements to strings (and lets null reach a later NPE) instead of rejecting them. Scalar `Text` rejects non-string primitives, and PyMilvus validates every Text array element as `str`, so row inserts/BulkWriter can silently change malformed caller data. Please validate each element is a non-null JSON string before conversion and add malformed-array coverage.

sdk-core/src/main/java/io/milvus/response/FieldDataWrapper.java
```diff
@@ -264,6 +265,7 @@ private List<?> getFieldDataInternal() throws IllegalResponseException {
             case Double:
             case VarChar:
             case String:
+            case Text:
```
yhmo, line 269: sdk-core/src/main/java/io/milvus/response/FieldDataWrapper.java line:268
Low ---- This new public decoding branch is missing from `getFieldData()`'s Javadoc above, which still documents only VarChar as returning `List<String>`. Please include Text so callers can discover the return contract without reading the switch.

sdk-bulkwriter/src/test/java/io/milvus/bulkwriter/BulkWriterTest.java
```diff
@@ -544,6 +544,127 @@ void testAppend() {
...
+                    .withCollectionSchema(schema)
+                    .withLocalPath("/tmp/bulk_writer_text")
+                    .withFileType(fileType)
+                    .build();
+            try (LocalBulkWriter writer = new LocalBulkWriter(param)) {
+                writer.appendRow(row);
+                writer.commit(false);
+                List<List<String>> batchFiles = writer.getBatchFiles();
+                Assertions.assertFalse(batchFiles.isEmpty());
+                if (fileType == BulkFileType.PARQUET) {
```
yhmo: sdk-bulkwriter/src/test/java/io/milvus/bulkwriter/BulkWriterTest.java line:599
Low ---- The JSON and CSV iterations only prove that a file was created; only Parquet reads back `body`, `paragraphs`, and `chunks`. Since this PR advertises all three BulkWriter formats and each has a different writer path, please deserialize the JSON/CSV output and assert the Text values too so omissions or coercion cannot pass unnoticed.

sdk-core/src/main/java/io/milvus/param/ParamUtils.java
liliu-z: The SDK now lets callers declare Array<Text> fields and Text struct sub-fields — DataType.Text plus SchemaUtils passing the element type straight through, and convertJsonArray/BulkWriter/ParquetUtils all accepting Text as an array element — but every current server build rejects them at collection creation. validateElementType in internal/proxy/util.go permits only Bool/Int*/Float/Double/VarChar, so a top-level Array<Text> field (array-field path) and a Text struct sub-field (the SDK converts it to Array+elementType=Text, then ValidateFieldsInStruct → validateElementType) both fall through to "element type Text is not supported"; verified identical on the 2026-06-06, 2026-07-20, and 2026-07-28 server checkouts. The new unit tests only assert the grpc schema is built and the live Docker test only exercises scalar Text, so this non-functional path ships with green tests and false confidence. Reject Text as an array element / struct sub-field type in the SDK (same as the BulkWriter array path) until the server adds support.

sdk-bulkwriter/src/test/java/io/milvus/bulkwriter/BulkWriterTest.java
lentitude2tk: Fixed in 5231420e. The BulkWriter test now focuses on server-supported scalar Text and reads back JSON, CSV, and Parquet output, asserting the complete long `body` value for each format. The unsupported Text array/struct BulkWriter cases were removed.

sdk-core/src/main/java/io/milvus/param/ParamUtils.java
lentitude2tk: Fixed in 5231420e. `convertJsonArray()` no longer delegates Text arrays to Gson's `List<String>` adapter. It now validates every element as a non-null JSON string and throws `ParamException` for numeric, boolean, or null elements. Added malformed-array coverage for all three cases.

lentitude2tk: Confirmed the current server-side restriction. We also checked PyMilvus master: it does not hard-reject `Array<Text>` or Text struct schemas client-side and keeps protocol encoding/decoding, leaving schema capability validation to Milvus. Java core now follows that model, with the PR description clarified so it no longer claims end-to-end server support. Unsupported Text array/struct paths were removed from BulkWriter, the live Docker test covers scalar Text only, and JSON Text arrays now have strict per-element string validation.

sdk-core/src/main/java/io/milvus/response/FieldDataWrapper.java
lentitude2tk: Fixed in 5231420e. The public `getFieldData()` Javadoc now documents that Text fields return `List<String>`.

sdk-bulkwriter/src/main/java/io/milvus/bulkwriter/BulkWriter.java
```diff
@@ -457,7 +458,7 @@ private Pair<Object, Integer> verifyArray(JsonElement object, CreateCollectionRe
         DataType elementType = field.getElementType();
         if (TypeSize.contains(elementType)) {
             rowSize = TypeSize.getSize(elementType) * ((List<?>) array).size();
-        } else if (elementType == DataType.VarChar) {
+        } else if (elementType == DataType.VarChar || elementType == DataType.Text) {
             for (String str : (List<String>) array) {
                 rowSize += str.length();
             }
```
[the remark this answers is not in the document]
lentitude2tk: This is no longer applicable after 5231420e: Text arrays are no longer accepted by BulkWriter, and the added `elementType == Text` row-size branch has been removed. Core JSON `Array<Text>` conversion now rejects null elements explicitly.

sdk-core/src/test/java/io/milvus/v2/client/MilvusClientV2DockerTest.java
```diff
@@ -2020,6 +2020,106 @@ void testHybridSearch() {
...
+                .data(Collections.singletonList(new FloatVec(vector)))
+                .limit(1)
+                .outputFields(Collections.singletonList("body"))
+                .consistencyLevel(ConsistencyLevel.STRONG)
+                .build());
+        Map<String, Object> searchEntity = searchResp.getSearchResults().get(0).get(0).getEntity();
+        Assertions.assertEquals(longText, searchEntity.get("body"));
+
+        client.dropCollection(DropCollectionReq.builder().collectionName(collectionName).build());
+    }
```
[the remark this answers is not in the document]
yhmo: no need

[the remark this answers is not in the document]
lentitude2tk: Thanks. This test follows the existing cleanup convention used throughout `MilvusClientV2DockerTest`, where collections are dropped at the end of the happy path. Keeping that convention here avoids introducing a one-off structure unrelated to Text support, so no change is planned in this PR.

PR #1958, opened 2026-07-20
Align insert/upsert validation with pymilvus

yhmo:


sdk-core/src/main/java/io/milvus/v2/utils/DataUtils.java
```diff
@@ -203,16 +203,38 @@ private static String combineStructFieldName(String structName, String subFieldN
...
-            List<String> allFieldNames = new ArrayList<>();
-            normalFields.forEach((schema) -> allFieldNames.add(schema.getName()));
-            structFields.forEach((schema) -> allFieldNames.add(schema.getName()));
+            boolean isUpsert = upsertBuilder != null;
+            List<CreateCollectionReq.FieldSchema> inputFields = new ArrayList<>();
+            Set<String> inputFieldNames = new HashSet<>();
+            Set<String> structFieldNames = new HashSet<>();
+            structFields.forEach((schema) -> structFieldNames.add(schema.getName()));
+
+            JsonObject firstRow = rows.isEmpty() ? null : rows.get(0);
```
yhmo: sdk-core/src/main/java/io/milvus/v2/utils/DataUtils.java line:219
Medium ---- This derives whether the auto-ID primary key is an input column solely from row 0. For a dynamic-enabled collection, if the first row omits `id` but a later row supplies it, `id` is excluded from `inputFieldNames` and that later value is packed into dynamic JSON; reversing the row order instead requires `id` on every row. Please validate auto-ID presence across all rows and either reject mixed batches or handle them consistently so row order cannot change the wire semantics.

sdk-core/src/main/java/io/milvus/v2/utils/DataUtils.java
```diff
@@ -236,35 +258,84 @@ private void checkAndSetRowData(DescribeCollectionResp descColl, List<JsonObject
...
+                        .map(insertDataInfo -> insertDataInfo.data.size())
+                        .filter(count -> count > 0)
+                        .forEach(fieldCounts::add);
+                structFieldCounts.values().stream()
+                        .filter(count -> count > 0)
+                        .forEach(fieldCounts::add);
+                if (collectionSchema.isEnableDynamicField() && !insertDynamicDataInfo.data.isEmpty()) {
+                    fieldCounts.add(insertDynamicDataInfo.data.size());
+                }
+                if (fieldCounts.size() > 1) {
```
yhmo: sdk-core/src/main/java/io/milvus/v2/utils/DataUtils.java line:333
Medium ---- This only detects different nonzero field counts; it does not ensure that the sole count equals `rows.size()`. In a dynamic-disabled partial upsert such as a row with `id`/`vector` followed by an empty row, both encoded fields have count 1, `fieldCounts.size()` is 1, and the request is still declared with `numRows=2`. Please reject such rows or compare every emitted field count with the batch size before sending the request.

sdk-core/src/main/java/io/milvus/v2/utils/DataUtils.java
```diff
@@ -236,35 +259,80 @@ private void checkAndSetRowData(DescribeCollectionResp descColl, List<JsonObject
...
+                            : null;
+                    if (structFieldName != null) {
+                        String message = partialUpdate
+                                ? String.format("Partial struct update is unsupported for struct sub-field `%s`; "
+                                + "update the whole struct field `%s` instead", rowFieldName, structFieldName)
+                                : String.format("Struct sub-field `%s` cannot be used as a top-level field; "
+                                + "write the whole struct field `%s` instead", rowFieldName, structFieldName);
+                        throw new MilvusClientException(ErrorCode.INVALID_PARAMS, message);
+                    }
+                    if (!inputFieldNames.contains(rowFieldName)
```
liliu-z, line 251: This new unknown-field rejection throws when a row carries a key absent from the schema and dynamic fields are disabled. The existing testCacheCollectionSchema (MilvusClientV2DockerTest) reuses a single `row` object across two different collection schemas without clearing old keys, so at line 2751 the row still holds a stale `pk` key while the active schema's primary key is `aaa`. That upsert is expected to succeed (assertEquals(1L, upsertResp.getUpsertCnt())) but now throws "The field: pk is not defined in the collection schema.", breaking the test. Reset/rebuild the row (or remove stale keys) in the fixture, or reconsider rejecting unknown fields this aggressively.

sdk-core/src/main/java/io/milvus/v2/utils/DataUtils.java
```diff
@@ -295,24 +366,59 @@ private void checkAndSetRowData(DescribeCollectionResp descColl, List<JsonObject
...
-                    return;
-                }
-
-                // if the field is an output field of doc-in-doc-out, no need to provide value
-                if (outputFieldNames.contains(fieldName)) {
-                    return;
-                }
-
                 // in v2.6.1 support partial update, user can input partial fields
                 if (partialUpdate) {
```
liliu-z, line 417: During partial update this returns early for any field absent from the row, including the primary key, so a partial upsert that omits the PK still builds a request with no PK. A partial update must identify the existing row by its primary key, so a request without one cannot be applied by the server. Every other partial-update path supplies the PK (UpsertTest.partialUpdateSingleField line 273, MilvusClientV2DockerTest line 1930, and this PR's own docker testInsertUpsertValidation line 2088), yet the new unit test testPartialUpsertAllowsMissingAutoIdField enshrines the PK-less request as valid. Continue to require the primary key even when partialUpdate is true.

sdk-core/src/main/java/io/milvus/v2/utils/DataUtils.java
```diff
@@ -515,7 +678,11 @@ public static Object checkFieldValue(CreateCollectionReq.FieldSchema field, Json
         int dim = field.getDimension() == null ? 0 : field.getDimension();
         int maxLength = field.getMaxLength() == null ? 0 : field.getMaxLength();
         int maxCapacity = field.getMaxCapacity() == null ? 0 : field.getMaxCapacity();
-        return ParamUtils.checkFieldValue(field.getName(), dataType, elementType, dim,
-                maxLength, maxCapacity, field.getIsNullable(), field.getDefaultValue(), fieldData);
+        try {
+            return ParamUtils.checkFieldValue(field.getName(), dataType, elementType, dim,
+                    maxLength, maxCapacity, field.getIsNullable(), field.getDefaultValue(), fieldData);
+        } catch (ParamException e) {
```
yhmo: sdk-core/src/main/java/io/milvus/v2/utils/DataUtils.java line:684
High ---- `ParamUtils.checkFieldValue()` can also fail with runtime conversion exceptions that are not `ParamException` (for example, Gson throws `NumberFormatException` when cached `Int64` schema parses `"abc"`). If the collection was recreated and that field is now `VarChar`, the row is valid against the fresh schema, but this exception bypasses `DataNotMatchException`, so `VectorService` no longer refreshes its cached schema and every subsequent insert/upsert keeps failing locally. Please translate all schema-dependent conversion failures into `DataNotMatchException` (while leaving request-level validation outside it) and cover a numeric-to-string schema recreation case.

sdk-core/src/test/java/io/milvus/v2/utils/DataUtilsTest.java
```diff
@@ -0,0 +1,506 @@
...
+
+        UpsertRequest request = new DataUtils.InsertBuilderWrapper().convertGrpcUpsertRequest(
+                UpsertReq.builder()
+                        .collectionName("test")
+                        .data(Collections.singletonList(row))
+                        .partialUpdate(true)
+                        .build(),
+                collection);
+
+        Assertions.assertEquals(Arrays.asList("id", ""), fieldNames(request.getFieldsDataList()));
```
yhmo: sdk-core/src/test/java/io/milvus/v2/utils/DataUtilsTest.java line:429
Medium ---- Dynamic fields are named `Constant.DYNAMIC_FIELD_NAME` (`"$meta"`), and PyMilvus serializes that name on insert/upsert `FieldData`. This expectation currently passes only because `ParamUtils.genFieldData(..., isDynamic=true)` drops `field_name`, so the test locks in the Java helper's legacy omission instead of the intended cross-SDK contract. Please preserve `$meta` when building dynamic `FieldData` and assert the constant here (and in the insert assertion above).

sdk-core/src/main/java/io/milvus/v2/service/vector/VectorService.java
```diff
@@ -121,7 +122,7 @@ public InsertResp insert(MilvusServiceGrpc.MilvusServiceBlockingStub blockingStu
         InsertRequest rpcRequest;
         try {
             rpcRequest = buildInsertRequest(request, descResp);
-        } catch (Exception ignored) {
+        } catch (DataNotMatchException ignored) {
```
yhmo, line 125: sdk-core/src/main/java/io/milvus/v2/service/vector/VectorService.java line:125
Medium ---- This catch no longer covers the partition-name validation in `DataUtils`, even though that decision is derived from cached schema. If another client recreates a cached collection from “has partition key” to “no partition key,” an insert/upsert that validly supplies a partition name fails locally with `MilvusClientException` before any RPC, so neither this forced refresh nor the server `SchemaMismatch` retry can run; the stale cache then makes the call fail repeatedly. Please classify that schema-dependent guard as retryable (or otherwise include it in the one-time refresh path) and cover the recreate scenario.

sdk-core/src/test/java/io/milvus/v2/utils/DataUtilsTest.java
```diff
@@ -0,0 +1,772 @@
...
+
+        Assertions.assertEquals(Arrays.asList(
+                        FieldPartialUpdateOp.OpType.REPLACE,
+                        FieldPartialUpdateOp.OpType.ARRAY_APPEND,
+                        FieldPartialUpdateOp.OpType.ARRAY_REMOVE),
+                request.getFieldOpsList().stream().map(FieldPartialUpdateOp::getOp).collect(Collectors.toList()));
+    }
+
+    @Test
+    void testNonReplaceFieldOperationEnablesPartialUpdate() {
```
yhmo, line 545: sdk-core/src/test/java/io/milvus/v2/utils/DataUtilsTest.java line:545
Medium ---- This only verifies auto-promotion for non-REPLACE operations, but PyMilvus 2.6 uses `partial_update or bool(ops)`, so a REPLACE-only `fieldOps` request must also enter partial-update mode. With the current `hasNonReplaceFieldOps` logic, a caller that sends REPLACE for one field and omits unrelated fields is validated as a full upsert and fails before RPC. Please cover the REPLACE-only case and promote partial update for any non-empty operation list.

sdk-core/src/main/java/io/milvus/v2/service/vector/request/UpsertReq.java
```diff
@@ -167,6 +167,15 @@ public UpsertReqBuilder fieldOps(List<FieldPartialUpdateOp> fieldOps) {
         }
 
         public UpsertReq build() {
+            if (fieldOps != null) {
+                for (FieldPartialUpdateOp fieldOp : fieldOps) {
+                    if (fieldOp != null && fieldOp.getOpType() != null
+                            && fieldOp.getOpType() != FieldPartialUpdateOp.OpType.REPLACE) {
+                        this.partialUpdate = true;
```
yhmo: sdk-core/src/main/java/io/milvus/v2/service/vector/request/UpsertReq.java line:174
Medium ---- Even if builder promotion is expanded to every field operation, deriving partialUpdate only during build leaves the public mutable request paths inconsistent. A caller can build the request and then call setFieldOps with ARRAY_APPEND/ARRAY_REMOVE (or reset partialUpdate to false); unlike the previous conversion-time check, convertGrpcUpsertRequest now forwards partialUpdate=false, so the server receives a non-REPLACE field operation without partial-update semantics. Please recompute this invariant from the current request at conversion time, or keep the setters synchronized, and cover the post-build mutation path.

PR #1960, opened 2026-07-21
Support Milvus 3.0 collection schema additions

### LINKED ISSUE #1959: AddFunctionField does not send required bound index parameters
## Problem

`MilvusClientV2.addFunctionField()` does not expose or serialize the bound index metadata required by Milvus 3.0. The generated `AlterCollectionSchemaRequest.FieldInfo` contains only the function output field schema, so the server rejects the request because an explicit `index_type` is required.

The collection function APIs also need to preserve compatibility with older Milvus servers. `dropCollectionFunction()` should use its original `DropCollectionFunction` RPC, while the SDK documentation should clarify that Milvus 3.0 and later require the function-field APIs instead.

## Expected behavior

- `AddFunctionFieldReq` accepts one bound `IndexParam`.
- The SDK serializes `index_name`, `index_type`, `metric_type`, and additional build parameters into `FieldInfo`.
- Invalid or conflicting bound index declarations are rejected before the RPC.
- `dropCollectionFunction()` keeps using the legacy RPC for older Milvus compatibility.
- Public API documentation directs Milvus 3.0 users to `addFunctionField()` and `dropFunctionField()`.

## Additional context

Milvus 3.0 couples a function with its output field. Runtime function-field creation currently supports BM25 and MinHash and requires the output vector field to have a bound index.

sijie-ni-0214:
## What this PR does

- Route `addCollectionField()` through a field-only `AlterCollectionSchema` request.
- Add a bound `IndexParam` to `AddFunctionFieldReq`.
- Serialize `index_name`, `index_type`, `metric_type`, and additional build parameters into `AlterCollectionSchemaRequest.FieldInfo`.
- Validate missing, automatic, mismatched, and conflicting bound index declarations before sending the RPC.
- Keep the legacy collection-function APIs on their original RPCs for compatibility with older Milvus servers.
- Document the function-field alternatives required by Milvus 3.0 and later.
- Update unit tests, the Docker lifecycle test, and the add-field example.

## Compatibility and authorization

- `addCollectionField()` now requires a Milvus server that supports `AlterCollectionSchema`. The server authorizes this request with the `AlterCollectionSchema` privilege instead of the legacy `AddCollectionField` privilege.
- `addCollectionFunction()` and `dropCollectionFunction()` keep using their legacy RPCs so a new SDK can still call older Milvus servers.
- Milvus 3.0 rejects the legacy collection-function RPCs. Applications targeting Milvus 3.0 should use `addFunctionField()` and `dropFunctionField()`.

## Test plan

- [x] `CollectionTest`: 37 tests passed.
- [x] Local Milvus E2E: add a nullable collection field, describe the updated schema, and insert data containing the new field.
- [x] Local Milvus E2E: add a BM25 function field with a bound `SPARSE_INVERTED_INDEX`, verify its metric and build parameters, then cascade-drop the function field and index.
- [x] Local Milvus E2E: verify Milvus 3.0 returns the expected unsupported errors for the legacy add/drop collection-function RPCs.
- [x] JaCoCo covers the new `AddFunctionFieldReq` index accessor path.
- [x] `git diff --check`.

Fixes #1959


sdk-core/src/main/java/io/milvus/v2/service/collection/request/AddFunctionFieldReq.java
```diff
@@ -19,16 +19,20 @@
 
 package io.milvus.v2.service.collection.request;
 
+import io.milvus.v2.common.IndexParam;
+
 public class AddFunctionFieldReq extends AddFieldReq {
     private String collectionName;
     private String databaseName;
     private CreateCollectionReq.Function function;
+    private IndexParam indexParam;
```
yhmo, line 28: sdk-core/src/main/java/io/milvus/v2/service/collection/request/AddFunctionFieldReq.java line:28
Low ---- This new bound index is mandatory, must target the function output field, and must use an explicit non-AUTOINDEX type, but none of those preconditions are documented on the request or public client method. Please document them here or on `addFunctionField` so callers do not discover the contract only through runtime exceptions.

sdk-core/src/test/java/io/milvus/v2/client/MilvusClientV2DockerTest.java
```diff
@@ -4004,6 +4004,12 @@ void testFunctionFieldLifecycle() {
                 .collectionName(collectionName)
                 .fieldName("sparse")
                 .dataType(DataType.SparseFloatVector)
+                .indexParam(IndexParam.builder()
```
yhmo, line 4007: sdk-core/src/test/java/io/milvus/v2/client/MilvusClientV2DockerTest.java line:4007
Medium ---- This lifecycle test only verifies that the field and function appear; it never observes the bound index that this change is meant to create. Please describe/list `sparse_idx` after the add and assert its type, metric, and build parameters, then verify the index is gone after `dropFunctionField`, otherwise the server-facing index materialization/cascade path can regress while this test still passes.

sijie-ni-0214: Extended the lifecycle test to describe sparse_idx and assert its field, SPARSE_INVERTED_INDEX type, BM25 metric, and drop_ratio_build parameter. The test also verifies that sparse_idx is absent after dropFunctionField. This is included in commit 94160646.

sdk-core/src/main/java/io/milvus/v2/service/collection/request/AddFunctionFieldReq.java
sijie-ni-0214: Added public addFunctionField Javadoc documenting that the bound index is required, must target the function output field, and must use an explicit type other than None or AUTOINDEX. This is included in commit 94160646.

sdk-core/src/main/java/io/milvus/v2/service/collection/CollectionService.java
```diff
@@ -805,20 +853,15 @@ public Void dropCollectionFunction(MilvusServiceGrpc.MilvusServiceBlockingStub b
         String dbName = request.getDatabaseName();
         String collectionName = request.getCollectionName();
         String title = String.format("Drop function to collection: '%s' in database: '%s'", collectionName, dbName);
```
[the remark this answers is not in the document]
sijie-ni-0214: Thanks. This wording predates this PR and is already present on upstream/master. I am keeping this PR scoped to the collection schema routing and compatibility changes; the wording can be cleaned up separately.

Issue #1926, opened 2026-06-25
Dependency Vulnerability: grpc-netty-shaded (CVE-2025-55163) in milvus-store


Aitenry:
**Describe the issue**
Our project is using `org.springframework.ai:spring-a-starter-vector-store-milvus:1.1.8` and a vulnerability scan reports that its transitive dependency `io.grpc:grpc-netty-shaded:1.59.1` is affected by CVE-2025-55163 (Score: 7.5).

**Vulnerability Details**
- **CVE:** CVE-2025-55163 
- **Description:** Netty, which is shaded by `grpc-netty-shaded`, is vulnerable to a "MadeYouReset" DDoS attack. The vulnerability allows resource exhaustion by using malformed HTTP/2 control frames to break the max concurrent streams limit .
- **Affected Dependency:** `io.grpc:grpc-netty-shaded:1.59.1`
- **Fixed Version:** `io.grpc:grpc-netty-shaded:1.75.0` 

**Dependency Tree (Relevant Part)**
Here's the relevant dependency path from our build:
```
maven:org.springframework:spring-a-starter-vector-store-milvus:1.1.8
  └─ maven:io.milvus:milvus-sdk-java:2.5.8
      └─ maven:io.grpc:grpc-netty-shaded:1.59.1 (vulnerable)
```

**Proposed Solution**
Could the Spring AI team consider bumping the version of `io.milvus:milvus-sdk-java` or overriding the `grpc-netty-shaded` dependency to version `1.75.0` or later in a future patch release? This would resolve the vulnerability for all downstream projects.

**Additional Context**
- Spring AI Version: `1.1.8` 
- Milvus SDK Version: `2.5.8`
- Build Tool: Maven

Thank you for your attention to this security matter.

yhmo:
A PR to upgrade versions of grpc/protobuf: https://github.com/milvus-io/milvus-sdk-java/pull/1977
It still has risks(I have noticed some examples failed with this patch), I intend to publish a new version for v2.5 once I confirm that all things work well.

yhmo:
I have upgraded gRPC to 1.75.0 and published a new version v2.5.16.

yhmo:
[...]

PR #1966, opened 2026-07-23
Dump gRPC and protobuf versions

yhmo:
issue: https://github.com/milvus-io/milvus-sdk-java/issues/1926
Upgrades gRPC to 1.75.0 with aligned Protobuf and Guava versions, adopts centralized dependency management and
  gRPC shading for safer publishing, and updates Docker tests to Milvus v2.5.27.

### DIFF
--- a/sdk-core/pom.xml
+++ b/sdk-core/pom.xml
@@ -19,55 +19,6 @@
         <skip.maven.deploy>false</skip.maven.deploy>
     </properties>
 
-    <dependencyManagement>
-        <dependencies>
-            <dependency>
-                <groupId>io.grpc</groupId>
-                <artifactId>grpc-bom</artifactId>
-                <version>${grpc.version}</version>
-                <type>pom</type>
-                <scope>import</scope>
-            </dependency>
-            <dependency>
-                <groupId>org.slf4j</groupId>
-                <artifactId>slf4j-api</artifactId>
-                <version>${slf4j.api.version}</version>
-            </dependency>
-            <dependency>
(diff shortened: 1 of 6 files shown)


sdk-core/pom.xml
```diff
@@ -177,4 +128,75 @@
...
+                                        <exclude>META-INF/*.DSA</exclude>
+                                        <exclude>META-INF/*.RSA</exclude>
+                                        <exclude>META-INF/MANIFEST.MF</exclude>
+                                        <exclude>module-info.class</exclude>
+                                    </excludes>
+                                </filter>
+                            </filters>
+                            <relocations>
+                                <relocation>
+                                    <pattern>io.grpc</pattern>
```
yhmo, line 178: sdk-core/pom.xml line:178
High ---- Relocating the entire `io.grpc` namespace also rewrites gRPC types in sdk-core's own public bytecode. On the 2.5 line, `ClientUtils.getChannel()` has long returned `io.grpc.ManagedChannel` and `configureProxy(...)` accepts `io.grpc.ManagedChannelBuilder`; after this shade step their JVM descriptors become `io.milvus.shaded.io.grpc...`, so existing compiled callers fail with `NoSuchMethodError` and source callers cannot pass a normal gRPC builder. The reactor tests run before the `package` phase, so they do not exercise this published JAR. Please preserve these public descriptors—or avoid applying this relocation in a 2.5 patch—rather than introducing a binary/source-incompatible API change as part of the CVE bump.

Issue #1935, opened 2026-07-01
Potential memory leak in SearchIterator due to cached subList() retention


rajb957:
There appears to be a potential memory leak / memory retention issue in SearchIterator when running large search iterations.

SearchIterator uses IteratorCache to keep leftover results between calls to next(). When a cached page contains more rows than the requested batch size, the current implementation splits the cached list using List.subList() and stores the remaining suffix back into the cache.

```
List<QueryResultsWrapper.RowRecord> retPageRes = cachedPage.subList(0, (int) count);
List<QueryResultsWrapper.RowRecord> leftCachePage = cachedPage.subList((int) count, cachedPage.size());
iteratorCache.cache(cacheId, leftCachePage);
return retPageRes;
```
In Java, subList() returns a view backed by the original list. This means the cached suffix can still retain a reference to the full original list, including rows that have already been returned to the user. During very large iterations, this can cause already-consumed result rows to remain reachable and prevent them from being garbage collected.

This is especially risky when:

SearchIterator is used for very large topK values.
Each result row contains large output fields.
The iterator repeatedly overfetches and stores leftover cached pages.
The user does not call close() immediately after finishing iteration.

### Affected Area

[sdk-core/src/main/java/io/milvus/orm/iterator/SearchIterator.java](https://github.com/milvus-io/milvus-sdk-java/blob/master/sdk-core/src/main/java/io/milvus/orm/iterator/SearchIterator.java)
[sdk-core/src/main/java/io/milvus/orm/iterator/IteratorCache.java](https://github.com/milvus-io/milvus-sdk-java/blob/master/sdk-core/src/main/java/io/milvus/orm/iterator/IteratorCache.java)

**Expected Behavior**

After a batch of results has been returned by SearchIterator.next(), rows that are no longer needed should become eligible for garbage collection.

The cache should only retain rows that are still needed for future next() calls.

**Actual Behavior**

The cache may store subList() views that retain the backing list of a larger result page. As a result, already-consumed rows may remain reachable through the cached suffix, causing unnecessary memory retention and potentially leading to increased heap usage or OOM in large iterations.

### Possible Approaches

1. Store defensive copies in IteratorCache so cached entries never retain subList() backing lists.
2. Drain the cache in place after each next() call by copying the returned page and clearing consumed rows from the cached list.
3. Add an IteratorCache.drain(cacheId, count) helper to centralize cache consumption, release empty caches, and avoid repeating subList() handling in iterator code.

[...]

PR #1967, opened 2026-07-23
Fix memory retention risk for iterator

yhmo:
issue: https://github.com/milvus-io/milvus-sdk-java/issues/1935

### DIFF
--- a/sdk-core/src/main/java/io/milvus/orm/iterator/IteratorCache.java
+++ b/sdk-core/src/main/java/io/milvus/orm/iterator/IteratorCache.java
@@ -19,8 +19,10 @@
 
 package io.milvus.orm.iterator;
 
+import io.milvus.exception.ParamException;
 import io.milvus.response.QueryResultsWrapper;
 
+import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
@@ -30,21 +32,103 @@ import static io.milvus.param.Constant.NO_CACHE_ID;
 
 public class IteratorCache {
     private final AtomicInteger cacheId = new AtomicInteger(0);
-    private final Map<Integer, List<QueryResultsWrapper.RowRecord>> cacheMap = new ConcurrentHashMap<>();
+    private final Map<Integer, CacheEntry> cacheMap = new ConcurrentHashMap<>();
 
-    public int cache(int cacheId, List<QueryResultsWrapper.RowRecord> result) {
+    public synchronized int cache(int cacheId, List<QueryResultsWrapper.RowRecord> result) {
+        if (result == null) {
+            throw new ParamException("Cannot cache a null result");
+        }
         if (cacheId == NO_CACHE_ID) {
             cacheId = this.cacheId.incrementAndGet();
         }
-        cacheMap.put(cacheId, result);
+        cacheMap.put(cacheId, new CacheEntry(result));
         return cacheId;
     }
 
-    public List<QueryResultsWrapper.RowRecord> fetchCache(int cacheId) {
-        return cacheMap.getOrDefault(cacheId, null);
+    public synchronized List<QueryResultsWrapper.RowRecord> fetchCache(int cacheId) {
+        CacheEntry cached = cacheMap.get(cacheId);
+        return cached == null ? null : cached.snapshot();
     }
 
-    public void releaseCache(int cacheId) {
+    public synchronized int size(int cacheId) {
+        CacheEntry cached = cacheMap.get(cacheId);
+        return cached == null ? 0 : cached.size();
+    }
+
+    public synchronized int append(int cacheId, List<QueryResultsWrapper.RowRecord> result) {
+        if (result == null) {
+            throw new ParamException("Cannot append a null result to cache");
+        }
+        if (cacheId == NO_CACHE_ID) {
+            throw new ParamException("Cannot append to an invalid cache ID");
+        }
+
+        CacheEntry cached = cacheMap.computeIfAbsent(cacheId, ignored -> new CacheEntry());
+        cached.append(result);
+        return cached.size();
+    }
+
+    public synchronized List<QueryResultsWrapper.RowRecord> drain(int cacheId, int count) {
+        if (count == 0) {
+            return new ArrayList<>();
+        }
+
+        CacheEntry cached = cacheMap.get(cacheId);
+        if (count < 0 || cached == null || count > cached.size()) {
+            String msg = String.format("Cannot drain %s results from cache %s with size %s",
+                    count, cacheId, cached == null ? 0 : cached.size());
+            throw new ParamException(msg);
+        }
+
+        List<QueryResultsWrapper.RowRecord> result = cached.drain(count);
+        if (cached.size() == 0) {
+            cacheMap.remove(cacheId);
+        }
+        return result;
+    }
+
+    public synchronized void releaseCache(int cacheId) {
         cacheMap.remove(cacheId);
+    }
+
+    private static final class CacheEntry {
+        private List<QueryResultsWrapper.RowRecord> rows;
+        private int offset;
+
+        private CacheEntry() {
+            this.rows = new ArrayList<>();
+        }
+
+        private CacheEntry(List<QueryResultsWrapper.RowRecord> rows) {
+            this.rows = new ArrayList<>(rows);
+        }
+
+        private int size() {
+            return rows.size() - offset;
+        }
+
+        private void append(List<QueryResultsWrapper.RowRecord> result) {
+            rows.addAll(result);
+        }
+
+        private List<QueryResultsWrapper.RowRecord> snapshot() {
+            return new ArrayList<>(rows.subList(offset, rows.size()));
+        }
+
+        private List<QueryResultsWrapper.RowRecord> drain(int count) {
+            List<QueryResultsWrapper.RowRecord> result = new ArrayList<>(count);
+            int end = offset + count;
+            for (int i = offset; i < end; i++) {
+                result.add(rows.get(i));
+                rows.set(i, null);
+            }
+            offset = end;
+
+            if (offset > 0 && offset >= rows.size() - offset) {
+                rows = new ArrayList<>(rows.subList(offset, rows.size()));
+                offset = 0;
+            }
+            return result;
+        }
     }
 }
--- a/sdk-core/src/main/java/io/milvus/orm/iterator/QueryIterator.java
+++ b/sdk-core/src/main/java/io/milvus/orm/iterator/QueryIterator.java
@@ -19,6 +19,7 @@
 
 package io.milvus.orm.iterator;
 
+import io.milvus.exception.ParamException;
 import io.milvus.grpc.DataType;
 import io.milvus.grpc.KeyValuePair;
 import io.milvus.grpc.QueryRequest;
@@ -48,11 +49,14 @@ public class QueryIterator {
     private final FieldType primaryField;
 
     private final QueryIteratorReq queryIteratorReq;
+    private final long collectionID;
     private final int batchSize;
     private final long limit;
     private final String expr;
     private long offset;
     private Object nextId;
+    private Object nextElementOffset;
+    private final boolean elementFilterIterator;
     private int cacheIdInUse;
     private long returnedCount;
     private final RpcUtils rpcUtils;
@@ -60,14 +64,17 @@ public class QueryIterator {
 
     public QueryIterator(QueryIteratorParam queryIteratorParam,
                          RpcStubWrapper blockingStub,
-                         FieldType primaryField) {
+                         FieldType primaryField,
+                         long collectionId) {
         this.iteratorCache = new IteratorCache();
         this.blockingStub = blockingStub;
         this.primaryField = primaryField;
         this.queryIteratorReq = IteratorAdapterV2.convertV1Param(queryIteratorParam);
+        this.collectionID = collectionId;
 
         this.batchSize = (int) queryIteratorParam.getBatchSize();
         this.expr = queryIteratorParam.getExpr();
+        this.elementFilterIterator = StringUtils.containsIgnoreCase(this.expr, "element_filter");
         this.limit = queryIteratorParam.getLimit();
         this.offset = queryIteratorParam.getOffset();
         this.rpcUtils = new RpcUtils();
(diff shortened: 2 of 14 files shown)


sdk-core/src/test/java/io/milvus/orm/iterator/IteratorCacheTest.java
```diff
@@ -0,0 +1,123 @@
...
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
+import static io.milvus.param.Constant.NO_CACHE_ID;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNull;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+
+class IteratorCacheTest {
```
yhmo, line 36: sdk-core/src/test/java/io/milvus/orm/iterator/IteratorCacheTest.java line:35
Medium ---- These tests validate IteratorCache in isolation, but this PR also changes QueryIterator.next() and SearchIterator.next(): finite-limit cleanup, cached-page extraction, and returnedCount semantics. A wiring regression in either iterator can still pass every helper test and reintroduce retention or premature termination. Please add focused iterator-level tests that exercise cached results and a finite limit without relying on close().

sdk-core/src/main/java/io/milvus/orm/iterator/SearchIterator.java
```diff
@@ -152,16 +153,20 @@ public List<QueryResultsWrapper.RowRecord> next() {
         }
 
         // 3. if all result has return, clear the filteredIds
-        if (retPage.isEmpty()) {
+        if (retPage.isEmpty() && filteredIds != null) {
```
yhmo: sdk-core/src/main/java/io/milvus/orm/iterator/SearchIterator.java line:156
Low ---- releaseCacheIfFinished(retPage) immediately below already clears filteredIds when the page is empty, so this branch clears the same list twice and leaves one cleanup invariant split across two sites. Removing this block keeps cache/filter cleanup owned by the helper.

sdk-core/src/main/java/io/milvus/orm/iterator/IteratorCache.java
```diff
@@ -30,21 +32,99 @@
...
+        }
+
+        CacheEntry cached = cacheMap.computeIfAbsent(cacheId, ignored -> new CacheEntry());
+        cached.append(result);
+        return cached.size();
+    }
+
+    public synchronized List<QueryResultsWrapper.RowRecord> drain(int cacheId, int count) {
+        CacheEntry cached = cacheMap.get(cacheId);
+        if (count < 0 || cached == null || count > cached.size()) {
```
liliu-z, line 77: `drain(cacheId, count)` throws whenever the cache entry is absent (`cached == null`) regardless of `count`, so a zero-count drain is treated as an error instead of a no-op returning an empty list. `batchSize == 0` reaches this: it passes `QueryIteratorParam.build()` (only `batchSize < 0` is rejected), is unvalidated on the v2 `QueryIteratorReq` and on `SearchIteratorParam`, and in `QueryIterator.next()` the guard `iteratorCache.size(cacheIdInUse) >= batchSize` is `0 >= 0` on the first call, routing into `drain(NO_CACHE_ID, 0)` which throws a cryptic `ParamException`. Before this PR the same input returned an empty page (`res.subList(0, Math.min(0, size))`), and `SearchIterator` regresses the same way once `releaseCacheIfFinished` clears the entry, so this is a behavior regression on both iterators. Either treat `count == 0` as a no-op in `drain`, or reject `batchSize <= 0` in the iterator param builders (v1, v2, and search).

sdk-core/src/test/java/io/milvus/docker-compose.yml
```diff
@@ -34,7 +34,7 @@ services:
 
   standalone:
     container_name: milvus-javasdk-standalone
-    image: milvusdb/milvus:master-20260720-bf674af6-amd64
+    image: milvusdb/milvus:master-20260723-32a44262
```
yhmo, line 37: sdk-core/src/test/java/io/milvus/docker-compose.yml line:37
Low ---- This client-only cache fix and its new tests do not depend on a newer Milvus server, but this line (and the two matching changes in docker-compose-multi.yml) also advances the server-under-test. That unrelated environment change makes a later failure or behavior difference harder to attribute to the iterator fix; could we keep the existing image tags here unless the new build is required and document that dependency?

sdk-core/src/main/java/io/milvus/orm/iterator/SearchIteratorV2.java
```diff
@@ -230,7 +235,7 @@ private List<SearchResp.SearchResult> wrapReturnRes(List<SearchResp.SearchResult
 
         int currentLen = res.size();
         if (currentLen > leftResCnt) {
-            res = res.subList(0, leftResCnt);
+            res = res.subList(0, leftResCnt.intValue());
```
yhmo: sdk-core/src/main/java/io/milvus/orm/iterator/SearchIteratorV2.java line:238
Medium ---- When currentLen exceeds the remaining limit, this returns a subList view backed by the full page, so a caller retaining the small final page also retains every unreturned row and its output fields. With a large batch this recreates the backing-list memory-retention pattern this PR removes from the other iterators; please return an independent ArrayList copy here as those paths now do.

sdk-core/src/main/java/io/milvus/orm/iterator/SearchIteratorV2.java
```diff
@@ -142,10 +142,11 @@ private void probeForCompability() {
         searchParams.put("collection_id", this.collectionID);
         searchParams.put("iterator", true);
         searchParams.put("search_iter_v2", true);
-        searchParams.put("guarantee_timestamp", 0L);
+        searchParams.putIfAbsent("guarantee_timestamp", 0L);
```
yhmo, line 145: sdk-core/src/main/java/io/milvus/orm/iterator/SearchIteratorV2.java line:145
Medium ---- putIfAbsent now preserves a caller-provided numeric timestamp, but executeSearch immediately passes it to VectorUtils, which casts the map value directly to long. A type-valid searchParams.put("guarantee_timestamp", 42) therefore throws ClassCastException during construction before the probe RPC. Please normalize the preserved value to Long before executeSearch, or make the converter use Number.longValue(), and cover a non-Long Number.

sdk-core/src/main/java/io/milvus/orm/iterator/QueryIterator.java
```diff
@@ -83,9 +91,12 @@ public QueryIterator(QueryIteratorReq queryIteratorReq,
         this.blockingStub = blockingStub;
         this.queryIteratorReq = queryIteratorReq;
         this.primaryField = IteratorAdapterV2.convertV2Field(primaryField);
+        this.collectionID = resolveCollectionID(blockingStub, queryIteratorReq.getDatabaseName(),
```
yhmo: sdk-core/src/main/java/io/milvus/orm/iterator/QueryIterator.java line:94
Medium ---- Both V1 and V2 callers already describe the collection to obtain the primary-key schema, and VectorService.queryIterator() can satisfy that lookup from cache. Resolving the ID again here therefore adds an unconditional network RPC and a second failure point to every iterator construction. Please pass the collection ID from the existing DescribeCollectionResponse into QueryIterator instead.

sdk-core/src/main/java/io/milvus/orm/iterator/QueryIterator.java
```diff
@@ -48,26 +48,32 @@ public class QueryIterator {
...
     private int cacheIdInUse;
     private long returnedCount;
     private final RpcUtils rpcUtils;
     private long sessionTs = 0;
 
     public QueryIterator(QueryIteratorParam queryIteratorParam,
                          RpcStubWrapper blockingStub,
-                         FieldType primaryField) {
+                         FieldType primaryField,
+                         long collectionId) {
```
yhmo, line 68: sdk-core/src/main/java/io/milvus/orm/iterator/QueryIterator.java line:67
Medium ---- Adding collectionId by replacing the existing public three-argument constructors removes their JVM descriptors. Any application compiled against the current SDK that constructs QueryIterator directly will fail with NoSuchMethodError after upgrade (and source callers no longer compile), even though the new in-repo call sites are updated. Please keep the old overloads and delegate/resolve the collection ID so this feature does not break the published API.

sdk-core/src/main/java/io/milvus/v2/service/vector/request/QueryIteratorReq.java
```diff
@@ -189,7 +189,7 @@ public static class QueryIteratorReqBuilder {
         private boolean ignoreGrowing = false;
         private String timezone = "";
         private long batchSize = 1000L;
-        private boolean reduceStopForBest = false;
+        private boolean reduceStopForBest = true;
```
yhmo, line 192: sdk-core/src/main/java/io/milvus/v2/service/vector/request/QueryIteratorReq.java line:192
Medium ---- This flips the V2 builder default from false to true, so unchanged callers now select IReduceInOrderForBest instead of IReduceInOrder on the server. That is an observable public-API semantic change unrelated to the linked memory-retention issue; please preserve the previous default (callers can opt in explicitly), or document and test this as an intentional compatibility change.

sdk-core/src/main/java/io/milvus/orm/iterator/QueryIterator.java
yhmo: as design

sdk-core/src/main/java/io/milvus/v2/service/vector/request/QueryIteratorReq.java
yhmo: align with pymilvus

sdk-core/src/main/java/io/milvus/orm/iterator/QueryIterator.java
```diff
@@ -48,26 +48,32 @@ public class QueryIterator {
...
         this.iteratorCache = new IteratorCache();
         this.blockingStub = blockingStub;
         this.primaryField = primaryField;
         this.queryIteratorReq = IteratorAdapterV2.convertV1Param(queryIteratorParam);
+        this.collectionID = collectionId;
```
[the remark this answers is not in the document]
yhmo: as design

sdk-core/src/main/java/io/milvus/v2/service/vector/request/QueryIteratorReq.java
```diff
@@ -189,7 +189,7 @@ public static class QueryIteratorReqBuilder {
...
         private Map<String, Object> filterTemplateValues = new HashMap<>();
```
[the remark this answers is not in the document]
yhmo: as design

sdk-core/src/main/java/io/milvus/orm/iterator/QueryIterator.java
```diff
@@ -163,6 +177,7 @@ private void updateCursor(List<QueryResultsWrapper.RowRecord> res) {
             return;
         }
         nextId = res.get(res.size() - 1).get(primaryField.getName());
+        nextElementOffset = res.get(res.size() - 1).get(Constant.OFFSET);
```
liliu-z, line 179: The element-filter resume cursor is read via `record.get(Constant.OFFSET)` (a field literally named "offset" in fields_data), but the server returns the per-element offset in `QueryResults.element_indices` (proto field #7), which `RowRecordWrapper.buildRowRecord` never materializes — it only iterates fields_data. So `nextElementOffset` is always null, `hasElementCursor()` is always false, and the cursor params plus the `>=` PK operator are never emitted. Element-filter iteration then advances with `pk > lastPk`, silently dropping the remaining matching elements of any PK whose elements straddle a batch boundary. Read `element_indices` from the QueryResults and thread it through as the resume offset; note the new test masks this by fabricating an `offset` field the real server never sends. (Reviewer cited SearchIteratorV2.java, but this logic is entirely in QueryIterator.)

sdk-core/src/main/java/io/milvus/orm/iterator/QueryIterator.java
```diff
@@ -78,14 +84,17 @@ public QueryIterator(QueryIteratorParam queryIteratorParam,
...
+                         long collectionId) {
         this.iteratorCache = new IteratorCache();
         this.blockingStub = blockingStub;
         this.queryIteratorReq = queryIteratorReq;
         this.primaryField = IteratorAdapterV2.convertV2Field(primaryField);
+        this.collectionID = collectionId;
 
         this.batchSize = (int) queryIteratorReq.getBatchSize();
         this.expr = queryIteratorReq.getExpr();
+        this.elementFilterIterator = StringUtils.containsIgnoreCase(this.expr, "element_filter");
```
liliu-z, line 98: Element-filter mode is detected with `StringUtils.containsIgnoreCase(expr, "element_filter")`, a substring test that also fires when `element_filter` appears as a field name or string literal in an ordinary query; an explicit request flag would be more robust for this newly added code. The practical impact is currently masked because `hasElementCursor()` additionally requires a non-null element offset (which never populates today — see the element_indices defect), so this is a fragility rather than a live misroute. (Reviewer cited SearchIteratorV2.java:145; the check is in QueryIterator's two constructors.)

sdk-core/src/main/java/io/milvus/orm/iterator/SearchIteratorV2.java
```diff
@@ -230,9 +235,9 @@ private List<SearchResp.SearchResult> wrapReturnRes(List<SearchResp.SearchResult
 
         int currentLen = res.size();
         if (currentLen > leftResCnt) {
-            res = res.subList(0, leftResCnt);
+            res = new ArrayList<>(res.subList(0, leftResCnt.intValue()));
```
liliu-z, line 239: In the external-filter path, once the limit is reached `wrapReturnRes` drives `leftResCnt` to zero and the subsequent `next()` returns early, but any results accumulated in the `cache` field beyond the last returned page are never drained, and `close()` is an empty no-op that cannot release them. Unlike `SearchIterator.close()` (now clears filteredIds) and `QueryIterator` (releases its cache) which this PR updated, `SearchIteratorV2` leaves this bounded-but-unreleasable tail pinned for the iterator's lifetime — directly relevant to this PR's memory-retention theme. Trim the cache on exhaustion and clear it in close().

sdk-core/src/main/java/io/milvus/orm/iterator/QueryIterator.java
yhmo: align with pymilvus

PR #1971, opened 2026-07-27
Fix memory retention risk for iterator

yhmo:


sdk-core/src/main/java/io/milvus/orm/iterator/QueryIterator.java
```diff
@@ -60,11 +61,13 @@ public class QueryIterator {
 
     public QueryIterator(QueryIteratorParam queryIteratorParam,
                          RpcStubWrapper blockingStub,
-                         FieldType primaryField) {
+                         FieldType primaryField,
+                         long collectionId) {
```
yhmo, line 65: sdk-core/src/main/java/io/milvus/orm/iterator/QueryIterator.java line:65
Medium ---- These public constructors previously accepted three arguments and are listed in the checked-in Javadoc, so replacing them with four-argument signatures breaks existing callers at compile time and can cause `NoSuchMethodError` for already-compiled applications. Could we keep the old overloads (for both V1 and V2) and delegate after resolving the collection ID, so this 2.6 patch remains source/binary compatible?

sdk-core/src/main/java/io/milvus/param/Constant.java
```diff
@@ -34,6 +34,9 @@ public class Constant {
     public static final String TIMEZONE = "timezone";
     public static final String REDUCE_STOP_FOR_BEST = "reduce_stop_for_best";
     public static final String ITERATOR_FIELD = "iterator";
+    public static final String QUERY_ITER_LAST_PK = "query_iter_last_pk";
```
yhmo: sdk-core/src/main/java/io/milvus/param/Constant.java line:37
Low ---- Could we remove `QUERY_ITER_LAST_PK` and `QUERY_ITER_LAST_ELEMENT_OFFSET` from this change? Neither symbol is referenced on the PR branch; they were only used by an earlier iteration of the element-filter implementation, so keeping them adds dead public API surface that future code may assume is supported.

sdk-core/src/main/java/io/milvus/orm/iterator/QueryIterator.java
yhmo: as design

PR #1973, opened 2026-07-28
Refine cache machinery

yhmo:
Refine schema and timestamp caching for V1 and V2 clients.

  - Add shared, endpoint-aware SchemaCache and CollectionTsCache.
  - Coalesce concurrent schema loads and prevent stale cache repopulation.
  - Invalidate caches after relevant DDL operations.
  - Preserve SESSION timestamps across rename and global-cluster failover.
  - Replace the legacy GTsDict.
  - Document client thread-safety expectations.
  - Add comprehensive cache and concurrency tests.


sdk-core/src/main/java/io/milvus/common/utils/cache/CollectionTsCache.java
```diff
@@ -0,0 +1,116 @@
...
+
+    public long get(String endpoint, String databaseName, String collectionName) {
+        CollectionCacheKey key = CollectionCacheKey.create(endpoint, databaseName, collectionName);
+        lock.readLock().lock();
+        try {
+            return cache.getOrDefault(key, 0L);
+        } finally {
+            lock.readLock().unlock();
+        }
+    }
```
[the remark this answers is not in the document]
yhmo: > the single-argument
  > convertSearchParam(SearchParam) overload is only used internally by
  > SearchIterator, which immediately overwrites guarantee_timestamp with its
  > probe/session timestamp. Normal search, query, and hybrid-search client paths
  > use the endpoint-aware overloads.
  >
  > Treating a blank endpoint as endpoint-agnostic could mix timestamps from
  > different Milvus servers when database and collection names overlap.
  > Therefore, we prefer retaining endpoint isolation and the 1L fallback when no
  > endpoint context is available. No code change is needed here.

sdk-core/src/main/java/io/milvus/v2/client/MilvusClientV2.java
```diff
@@ -208,6 +221,10 @@ private void connect(ConnectConfig connectConfig) {
     private synchronized void updatePrimaryConnection(MilvusClientV2 primaryClient) {
         this.channel = primaryClient.channel;
         this.blockingStub = primaryClient.blockingStub;
+        this.cacheEndpoint = primaryClient.cacheEndpoint;
```
yhmo: sdk-core/src/main/java/io/milvus/v2/client/MilvusClientV2.java line:224
Medium ---- On a global-cluster primary change, replacing cacheEndpoint with the physical new primary means CollectionTsCache no longer sees timestamps recorded under the old primary. The same client therefore loses SESSION read-your-writes immediately after failover, whereas the previous global GTsDict survived the swap. Please preserve or migrate a logical global-client timestamp scope across primary changes while still isolating unrelated clusters, and add a failover test.

sdk-core/src/main/java/io/milvus/v2/utils/VectorUtils.java
```diff
@@ -143,19 +154,18 @@ public QueryRequest ConvertToGrpcQueryRequest(QueryReq request) {
...
+            return ts == 0L ? 1L : ts;
         }
         switch (consistencyLevel) {
             case STRONG:
                 return 0L;
             case SESSION: {
-                String key = GTsDict.CombineCollectionName(dbName, collectionName);
-                Long ts = GTsDict.getInstance().getCollectionTs(key);
-                return (ts == null) ? 1L : ts;
+                long ts = CollectionTsCache.getInstance().get(endpoint, cacheDbName, collectionName);
```
yhmo: sdk-core/src/main/java/io/milvus/v2/utils/VectorUtils.java line:167
Medium ---- The earlier empty-endpoint thread concluded that only SearchIterator uses an unconfigured conversion and overwrites the timestamp, but SearchIteratorV2 also calls new VectorUtils().ConvertToGrpcSearchRequest(...) and does not overwrite the top-level guarantee timestamp before its first request. Writes are cached under host:port while this instance reads endpoint "", so a SESSION iterator starts with 1L and can miss the latest write. Please pass endpoint/current-database context into this iterator path and cover its initial request.

sdk-core/src/main/java/io/milvus/v2/service/collection/CollectionService.java
```diff
@@ -518,6 +522,17 @@ public Void renameCollection(MilvusServiceGrpc.MilvusServiceBlockingStub blockin
...
         rpcUtils.handleResponse(title, status);
 
+        String sourceDb = actualDbName(dbName);
+        String targetDb = StringUtils.isNotEmpty(targetDbName) ? targetDbName : sourceDb;
+        invalidateSchema(sourceDb, collectionName);
+        invalidateSchema(targetDb, newName);
+        if (sourceDb.equals(targetDb)) {
+            CollectionTsCache.getInstance().rename(getEndpoint(), sourceDb, collectionName, newName);
+        } else {
+            invalidateTimestamp(sourceDb, collectionName);
```
yhmo: sdk-core/src/main/java/io/milvus/v2/service/collection/CollectionService.java line:532
Medium ---- A cross-database rename moves the same collection, but this branch deletes the source timestamp and clears the target instead of transferring the latest write timestamp. The subsequent SESSION read through the target database falls back to 1L, unlike the same-database rename above. Please support rename across two database keys in both V2 and V1 and test write -> cross-database rename -> SESSION read.

sdk-core/src/main/java/io/milvus/common/utils/cache/SchemaCache.java
```diff
@@ -0,0 +1,271 @@
...
+        CollectionCacheKey key = CollectionCacheKey.create(endpoint, databaseName, collectionName);
+        DescribeCollectionResponse initial = getCached(key);
+        if (initial != null && !forceUpdate) {
+            return initial;
+        }
+
+        LoadState newState = new LoadState();
+        LoadState state = loading.putIfAbsent(key, newState);
+        if (state != null) {
+            return await(state);
```
yhmo, line 132: sdk-core/src/main/java/io/milvus/common/utils/cache/SchemaCache.java line:92
Medium ---- After invalidate(...) marks an in-flight LoadState, a caller that arrives after invalidation still hits this branch and awaits/returns that state response. If the load began before an alter/drop/recreate, the post-invalidation insert or query can consume the old schema even though it must no longer be served; the current test only proves it is not re-cached. Please make invalidated generations unavailable to new callers and add a post-invalidation waiter test.

sdk-core/src/main/java/io/milvus/v2/utils/VectorUtils.java
yhmo: align with pymlivus

sdk-core/src/main/java/io/milvus/common/utils/cache/CollectionTsCache.java
```diff
@@ -0,0 +1,121 @@
...
+    public void rename(String endpoint, String oldDatabaseName, String oldCollectionName,
+                       String newDatabaseName, String newCollectionName) {
+        CollectionCacheKey oldKey = CollectionCacheKey.create(endpoint, oldDatabaseName, oldCollectionName);
+        CollectionCacheKey newKey = CollectionCacheKey.create(endpoint, newDatabaseName, newCollectionName);
+        lock.writeLock().lock();
+        try {
+            if (oldKey.equals(newKey)) {
+                return;
+            }
+            Long timestamp = cache.remove(oldKey);
```
yhmo: sdk-core/src/main/java/io/milvus/common/utils/cache/CollectionTsCache.java line:94
Medium ---- A successful DML can race this rename under the newly documented concurrent-RPC contract: if the server processes insert(old) before rename(old→new) but the insert response arrives afterward, this code moves the timestamp and the delayed DML completion then writes it back under old. A subsequent SESSION read of new falls back to 1 or an older timestamp, breaking read-your-writes. Please make timestamp lifecycle changes generation-aware (or otherwise serialize/migrate late DML completions) and add a delayed-response rename/drop test.

sdk-core/src/main/java/io/milvus/common/utils/cache/SchemaCache.java
```diff
@@ -0,0 +1,289 @@
...
+        }
+    }
+
+    private DescribeCollectionResponse await(LoadState state) {
+        boolean interrupted = false;
+        synchronized (state) {
+            while (!state.completed) {
+                try {
+                    state.wait();
+                } catch (InterruptedException ignored) {
```
yhmo: sdk-core/src/main/java/io/milvus/common/utils/cache/SchemaCache.java line:234
Medium ---- This catch consumes interruption and continues waiting until the shared loader completes. Since RPC deadlines default to disabled, one stuck describeCollection can leave every coalesced caller blocked even after its thread is cancelled or the client is shutting down. Please let an interrupted waiter exit promptly while restoring or propagating interruption, and add a test with a blocked loader.

sdk-core/src/main/java/io/milvus/v2/service/vector/VectorService.java
```diff
@@ -133,18 +126,18 @@ public InsertResp insert(MilvusServiceGrpc.MilvusServiceBlockingStub blockingStu
...
         }
 
         // if illegal data, server fails to process insert, else succeed
         cleanCacheIfFailed(response.getStatus(), dbName, collectionName);
         rpcUtils.handleResponse(title, response.getStatus());
 
         // update the last write timestamp for SESSION consistency
-        String key = GTsDict.CombineCollectionName(actualDbName(dbName), collectionName);
-        GTsDict.getInstance().updateCollectionTs(key, response.getTimestamp());
+        CollectionTsCache.getInstance().set(getEndpoint(), actualDbName(dbName), collectionName,
```
liliu-z: renameCollection calls CollectionTsCache.rename(), which removes the old key and moves its timestamp to the new key. But an insert/upsert/delete dispatched under the old name whose response returns after the rename runs set(getEndpoint(), db, oldName, T) here; because set() uses max-merge (CollectionTsCache.set compute) and rename already removed the old key, the late write re-creates the orphaned old key with the just-committed timestamp T while the new key keeps the older moved value. A later SESSION read under the new name (VectorUtils.getGuaranteeTimestamp -> CollectionTsCache.get) then reads the smaller timestamp and can miss the just-committed write. rename() cannot observe in-flight DML responses, so the fix must redirect in-flight writes to the new key or block a post-rename set() on the old key. The same late-set hazard applies to the upsert (line 202) and delete set() calls added in this PR.

sdk-core/src/main/java/io/milvus/common/utils/cache/CollectionTsCache.java
yhmo: Concurrent DML and DQL operations are supported. A DQL operation that overlaps a DML operation is not guaranteed to observe that write; wait for the DML operation to complete and use the appropriate consistency level when this ordering is required. DDL operations that change a database or collection's lifecycle or schema, such as drop, rename, truncate, or schema alteration, must not run concurrently with DML or DQL operations on the affected database or collection. Serialize these operations to avoid undefined ordering and unexpected results.

yhmo: we will document it in the documentation

sdk-core/src/main/java/io/milvus/v2/service/vector/VectorService.java
yhmo: This can occur only when collection rename overlaps an in-flight DML operation. Our thread-safety contract now explicitly excludes concurrent lifecycle/schema-changing DDL and DML/DQL on the same collection and requires callers to serialize them. Concurrent DML/DQL remains supported. Generation-aware redirection would add substantial complexity and the same limitation exists in the C++ SDK and PyMilvus, so we are defining this ordering as unsupported rather than adding cache lifecycle tracking.

sdk-core/src/main/java/io/milvus/common/utils/cache/SchemaCache.java
```diff
@@ -0,0 +1,286 @@
...
+
+        try {
+            DescribeCollectionResponse current = getCached(key);
+            if (current != null && (!forceUpdate || current != initial)) {
+                complete(state, current, null);
+                return current;
+            }
+
+            DescribeCollectionResponse loaded = loader.load();
+            if (!state.invalidated.get()) {
```
yhmo: sdk-core/src/main/java/io/milvus/common/utils/cache/SchemaCache.java line:109
High ---- The invalidation check and cache write are not atomic: if invalidate(), invalidateDb(), or clear() marks and removes this LoadState after the flag read but before setCached(), the old loader repopulates a schema after invalidation has completed. That violates the lifecycle guarantee and can make later DML reuse stale metadata; please serialize the generation check with publication (or compare a generation token under one lock) and add a test for this exact check-then-publish interleaving.

sdk-core/src/main/java/io/milvus/param/ParamUtils.java
```diff
@@ -866,6 +866,16 @@ public static void compatibleSearchParams(Map<String, Object> searchParams, Sear
 
     @SuppressWarnings("unchecked")
     public static SearchRequest convertSearchParam(SearchParam requestParam) throws ParamException {
+        return convertSearchParam(requestParam, "");
```
yhmo: sdk-core/src/main/java/io/milvus/param/ParamUtils.java line:869
Medium ---- The earlier thread considered only in-repo callers, but this existing overload is public and historically obtained the session timestamp populated by client DML (the database-aware behavior was restored in #1472). It now always reads endpoint "", while normal clients write under host:port, so unchanged external code that converts and sends a SESSION/default request gets guarantee_timestamp=1 instead of its last write. Please preserve the legacy overload semantics or explicitly stage a compatible deprecation and migration.

sdk-core/src/main/java/io/milvus/common/utils/GTsDict.java
```diff
@@ -1,74 +0,0 @@
...
-
-package io.milvus.common.utils;
-
-import io.milvus.exception.ParamException;
-import org.apache.commons.lang3.StringUtils;
-
-import java.util.concurrent.ConcurrentHashMap;
-import java.util.concurrent.ConcurrentMap;
-
-public class GTsDict {
```
yhmo, line 28: sdk-core/src/main/java/io/milvus/common/utils/GTsDict.java line:28
Medium ---- GTsDict is a public class with public singleton, update, and query methods, so deleting it outright breaks existing source imports and already-compiled applications with NoClassDefFoundError in a released 3.0.x line. Please retain a deprecated forwarding compatibility shim, or make this removal part of an explicit major-version migration, while moving SDK internals to CollectionTsCache.

sdk-core/src/test/java/io/milvus/client/MilvusClientDockerTest.java
```diff
@@ -2984,7 +2984,7 @@ void testCacheCollectionSchema() throws InterruptedException {
 
         // use the temp client to insert correct data into the default collection
         // there will be a schema cache for this collection in the temp client
-        // there will be timestamp for this collection in the global GTsDict
+        // there will be a timestamp for this collection in the global timestamp cache
```
yhmo, line 2987: sdk-core/src/test/java/io/milvus/client/MilvusClientDockerTest.java line:2987
Low ---- The preceding comment still says the schema cache is "in the temp client", but this PR intentionally makes SchemaCache process-global and shared across clients; the same stale wording appears in MilvusClientV2DockerTest. Please update both comments so these cache tests describe the ownership they are verifying.

sdk-core/src/main/java/io/milvus/common/utils/GTsDict.java
yhmo: as design

sdk-core/src/main/java/io/milvus/common/utils/cache/SchemaCache.java
```diff
@@ -0,0 +1,299 @@
...
+        LoadState newState = new LoadState();
+        LoadState state;
+        synchronized (loadingLock) {
+            state = loading.get(key);
+            if (state == null) {
+                loading.put(key, newState);
+            }
+        }
+        if (state != null) {
+            return await(state);
```
yhmo, line 132: sdk-core/src/main/java/io/milvus/common/utils/cache/SchemaCache.java line:97
High ---- The earlier interruption fix only lets an explicitly interrupted waiter escape. Because in-flight loads are process-global and keyed only by endpoint, database, and collection, a second client with a shorter RPC deadline or different credentials still joins the first client load here, while await() has no deadline. If the first client uses the default no-deadline stub, the second client insert or upsert can block indefinitely instead of honoring its configured timeout. Please scope in-flight coordination to the client or caller policy, or make waiting honor the joining caller deadline, and add a two-client mixed-deadline test.

sdk-core/src/main/java/io/milvus/client/AbstractMilvusGrpcClient.java
```diff
@@ -88,11 +86,16 @@ public abstract class AbstractMilvusGrpcClient implements MilvusClient {
 
     protected abstract String currentDbName();
 
+    protected String currentEndpoint() {
```
yhmo: sdk-core/src/main/java/io/milvus/client/AbstractMilvusGrpcClient.java line:89
Medium ---- Returning the same empty endpoint for every existing subclass silently changes the old per-instance schema cache into one process-global bucket. A third-party AbstractMilvusGrpcClient subclass compiled before this hook was added will not override it, so clients connected to different clusters but using the same database and collection can reuse a DescribeCollectionResponse loaded by the other client and build DML with the wrong schema. Please make the compatibility default isolate per instance, or require/provide a real endpoint without silently sharing, and cover two legacy subclasses targeting different stubs.

sdk-core/src/main/java/io/milvus/common/utils/cache/CollectionTsCache.java
```diff
@@ -0,0 +1,148 @@
...
+     * <p>This method exists for compatibility with legacy public request-conversion helpers that
+     * do not accept an endpoint. Normal client request paths should use {@link #get(String, String,
+     * String)} so timestamps from different Milvus clusters remain isolated.</p>
+     */
+    public long getAnyEndpoint(String databaseName, String collectionName) {
+        CollectionCacheKey target = CollectionCacheKey.create("", databaseName, collectionName);
+        lock.readLock().lock();
+        try {
+            long latestTimestamp = 0L;
+            for (Map.Entry<CollectionCacheKey, Long> entry : cache.entrySet()) {
```
yhmo: sdk-core/src/main/java/io/milvus/common/utils/cache/CollectionTsCache.java line:59
Medium ---- getAnyEndpoint() scans the entire process-global timestamp map, which has no capacity or expiry. This runs in every legacy conversion; in particular, V1 SearchIterator calls ParamUtils.convertSearchParam(...) on every page and immediately overwrites the resulting guarantee timestamp, so an iterator pays O(all cached collections and endpoints) pure overhead per batch. Please keep an O(1) aggregate for the compatibility lookup, or pass endpoint context or skip the lookup in the iterator path.

yhmo: no need to fix, the risk is only for v1

sdk-core/src/main/java/io/milvus/v2/client/MilvusClientV2.java
```diff
@@ -149,6 +177,7 @@ private void initServices(String dbName) {
      */
     private void connect(ConnectConfig connectConfig) {
         this.connectConfig = connectConfig;
+        this.cacheEndpoint = connectConfig.getHost() + ":" + connectConfig.getPort();
```
liliu-z, line 180: The shared-cache scope key is built from host:port only and ignores serverName, which ClientUtils.getChannel passes to gRPC overrideAuthority. Since this PR makes SchemaCache and CollectionTsCache process-global singletons shared across client instances, two clients that target the same host:port but use different authorities (for example an ingress that routes to distinct clusters by SNI/authority) share one cache entry and can read schema or session timestamps computed for the other server. The V1 client has the same host:port-only key at MilvusServiceClient.java:86. Fold the effective authority into the cache endpoint. In common deployments each cluster has a distinct host, so real-world exposure is narrow.

sdk-core/src/main/java/io/milvus/v2/service/collection/CollectionService.java
```diff
@@ -102,6 +102,7 @@ public Void createCollection(MilvusServiceGrpc.MilvusServiceBlockingStub blockin
 
         Status status = blockingStub.createCollection(builder.build());
         rpcUtils.handleResponse(title, status);
+        invalidateTimestamp(dbName, collectionName);
```
liliu-z: createCollection invalidates only the timestamp cache, not the schema cache; createCollectionWithSchema (line 182) and the V1 create path at AbstractMilvusGrpcClient.java:640 have the same gap, whereas every other DDL path (drop, rename, alter/add/drop field) invalidates the schema. If a stale schema entry for this name survives in the shared cache (for example the collection was dropped out-of-band by another process, so this client's entry was never invalidated) and this client then successfully recreates it with a different primary key, the stale schema stays cached and is used to build delete-by-id/get requests (describeCollection itself is a fresh RPC and insert self-heals via the SchemaMismatch retry, so exposure is limited to those paths). Invalidate the schema cache on create for consistency with the other DDL paths.

sdk-core/src/main/java/io/milvus/v2/service/vector/VectorService.java
```diff
@@ -353,14 +351,11 @@ public DeleteResp delete(MilvusServiceGrpc.MilvusServiceBlockingStub blockingStu
         DeleteRequest rpcRequest = dataUtils.ConvertToGrpcDeleteRequest(request);
         MutationResult response = blockingStub.delete(rpcRequest);
 
-        // if illegal data, server fails to process delete, clean the schema cache
-        // so that the next call of dml can update the cache
-        cleanCacheIfFailed(response.getStatus(), dbName, collectionName);
         rpcUtils.handleResponse(title, response.getStatus());
```
liliu-z, line 334: This delete path no longer evicts the schema cache on failure: the prior cleanCacheIfFailed(response.getStatus(), ...) call was removed here, yet it is still retained for insert (line 135) and upsert. delete-by-id derives its filter from the cached primary-key name (getCollectionInfo(..., false) plus getExprById at lines 347-349) and, unlike insert, has no SchemaMismatch retry or force-reload. So when the cached schema is stale, the malformed delete is rejected and the same bad schema remains cached, so retries keep failing with no self-healing; if the stale primary-key name still resolves to a valid field, the delete can even succeed against the wrong field and remove wrong rows. Restore eviction of the collection's schema entry on the delete failure path.

sdk-core/src/main/java/io/milvus/v2/client/MilvusClientV2.java
yhmo: unnecessary

sdk-core/src/main/java/io/milvus/param/ParamUtils.java
```diff
@@ -1036,7 +1056,8 @@ public static SearchRequest convertAnnSearchParam(AnnSearchParam annSearchParam,
         return builder.build();
     }
 
-    public static HybridSearchRequest convertHybridSearchParam(HybridSearchParam requestParam) throws ParamException {
+    public static HybridSearchRequest convertHybridSearchParam(HybridSearchParam requestParam, String endpoint,
```
yhmo, line 1074: sdk-core/src/main/java/io/milvus/param/ParamUtils.java line:1059
High ---- This replaces the existing public one-argument convertHybridSearchParam(HybridSearchParam) overload, and the same happens to convertQueryParam below, so unchanged applications compiled against these helpers no longer compile (and already-compiled code can fail with NoSuchMethodError). The PR already preserves convertSearchParam(SearchParam); please retain equivalent delegating overloads for hybrid search and query so adding endpoint context does not break the released API.

sdk-core/src/main/java/io/milvus/orm/iterator/RpcStubWrapper.java
```diff
@@ -30,11 +32,14 @@ public class RpcStubWrapper {
     // we have to reset this value for each time QueryIterator calls the query() interface.
     // the rpcDeadlineMs value is passed from MilvusClient
     private long rpcDeadlineMs = 0L;
+    private final CacheContext cacheContext;
 
     public RpcStubWrapper(MilvusServiceGrpc.MilvusServiceBlockingStub blockingStub,
```
yhmo, line 37: sdk-core/src/main/java/io/milvus/orm/iterator/RpcStubWrapper.java line:37
Medium ---- This replaces the public two-argument RpcStubWrapper constructor introduced for iterator deadline handling, so external iterator integrations using new RpcStubWrapper(stub, deadline) stop compiling or fail with NoSuchMethodError. Please keep that constructor as a compatibility overload, or stage the break through an explicit major-version migration.

sdk-core/src/main/java/io/milvus/v2/service/vector/VectorService.java
```diff
@@ -71,35 +69,13 @@ private DescribeCollectionResponse describeCollection(MilvusServiceGrpc.MilvusSe
...
-     * clean the cache so that the next insert will call describeCollection() to get the latest info.
-     */
-    private void cleanCacheIfFailed(Status status, String databaseName, String collectionName) {
-        if ((status.getCode() != 0 && status.getCode() != 8) ||
-                (!status.getErrorCode().equals(io.milvus.grpc.ErrorCode.Success) &&
-                        status.getErrorCode() != io.milvus.grpc.ErrorCode.RateLimit)) {
-            removeCollectionCache(databaseName, collectionName);
-        }
+        String dbName = actualDbName(databaseName);
+        return SchemaCache.getInstance().getOrLoad(getEndpoint(), dbName, collectionName, forceUpdate, this,
```
yhmo, line 72: sdk-core/src/main/java/io/milvus/v2/service/vector/VectorService.java line:73
Low ---- The comment above still says any insert/upsert server error removes the cached schema, but this PR intentionally preserves the schema for ordinary invalid-data failures and invalidates only for SchemaMismatch (the new tests assert that policy). Please update this comment, and the identical V1 comment, so maintainers do not reintroduce broad invalidation based on stale documentation.

sdk-core/src/main/java/io/milvus/client/AbstractMilvusGrpcClient.java
```diff
@@ -1736,15 +1748,15 @@ public ListenableFuture<R<MutationResult>> insertAsync(InsertParam requestParam)
...
+                        if (result.getStatus().getErrorCode() == ErrorCode.SchemaMismatch) {
+                            removeCollectionCache(dbName, collectionName);
+                        }
                         if (result.getStatus().getErrorCode() == ErrorCode.Success) {
                             logDebug("{} successfully!", title);
 
                             // update the last write timestamp for SESSION consistency
-                            String key = GTsDict.CombineCollectionName(actualDbName(dbName), collectionName);
-                            GTsDict.getInstance().updateCollectionTs(key, result.getTimestamp());
+                            CollectionTsCache.getInstance().set(currentEndpoint(), actualDbName(dbName), collectionName,
```
liliu-z: In insertAsync/upsertAsync the session-timestamp CollectionTsCache.set() runs in a Futures.addCallback listener that is registered separately from the Futures.transform listener producing the returned future; both attach to the same `response` with directExecutor() and have no guaranteed relative ordering. If `response` completes between the two registrations, the set() runs on the gRPC completion thread while the transform runs inline on the caller's thread, so a caller awaiting the returned future can proceed before the timestamp is written and miss its own write on a later SESSION read. Fold the set() into the transform so the returned future resolves only after the write. This is realistically reachable only under in-process/synchronous transports, not real network RPCs.

sdk-core/src/main/java/io/milvus/param/ParamUtils.java
yhmo: ParamUtils is used internally. no need to keep it

sdk-core/src/main/java/io/milvus/orm/iterator/RpcStubWrapper.java
yhmo: RpcStubWrapper is internally used, no need to concern

sdk-core/src/main/java/io/milvus/client/AbstractMilvusGrpcClient.java
yhmo: no need to fix

sdk-core/src/main/milvus-proto
```diff
@@ -1 +1 @@
-Subproject commit e378dbfa23a2abc16378e8c7e525754c48d82e4b
+Subproject commit 9f7d27f3eb7fe02b8598806f01ea1b03f19b721e
```
yhmo, line 1: sdk-core/src/main/milvus-proto line:1
Medium ---- This cache refactor also advances the proto submodule from e378dbfa to 9f7d27f3, pulling in 11 unrelated API features. None of the changed cache code depends on those definitions, so merging this PR would silently broaden the generated SDK surface and make cache-review and rollback risk include unrelated protocol changes. Please revert this pointer and update it in a dedicated proto/API PR.

yhmo: as design

sdk-core/src/main/java/io/milvus/common/utils/cache/CacheContext.java
```diff
@@ -0,0 +1,63 @@
...
+
+import java.util.Objects;
+
+/**
+ * Immutable cache identity propagated with an RPC path.
+ */
+public final class CacheContext {
+    private final String endpoint;
+    private final String databaseName;
+    private final Object loadScope;
```
yhmo: sdk-core/src/main/java/io/milvus/common/utils/cache/CacheContext.java line:30
Low ---- `loadScope` is stored and exposed here, but `getLoadScope()` has no callers and `SchemaCache` coalescing still receives each service instance directly. This dead field forces every iterator `RpcStubWrapper` to supply an otherwise meaningless object and contributed to replacing its constructor; please remove it from `CacheContext` (or actually thread it into schema loading) so the context carries only state that participates in the cache dataflow.

sdk-core/src/main/java/io/milvus/param/ParamUtils.java
```diff
@@ -893,8 +893,13 @@ public static void compatibleSearchParams(Map<String, Object> searchParams, Sear
         }
     }
 
-    @SuppressWarnings("unchecked")
-    public static SearchRequest convertSearchParam(SearchParam requestParam) throws ParamException {
+    public static SearchRequest convertSearchParam(SearchParam requestParam, String endpoint,
+                                                   String cacheDatabaseName) throws ParamException {
+        return convertSearchParamInternal(requestParam, endpoint, cacheDatabaseName);
+    }
+
```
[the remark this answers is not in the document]
yhmo: ParamUtils is internal use, no need to care

sdk-core/src/main/java/io/milvus/orm/iterator/RpcStubWrapper.java
```diff
@@ -29,12 +30,21 @@ public class RpcStubWrapper {
...
+                          String endpoint,
+                          String databaseName) {
+        this.blockingStub = Objects.requireNonNull(blockingStub, "blockingStub cannot be null");
         this.rpcDeadlineMs = rpcDeadlineMs;
+        if (endpoint == null || endpoint.trim().isEmpty()) {
+            throw new IllegalArgumentException("Cache endpoint cannot be empty");
+        }
+        this.endpoint = endpoint;
+        this.databaseName = databaseName == null || databaseName.isEmpty() ? "default" : databaseName;
     }
```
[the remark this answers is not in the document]
yhmo: RpcStubWrapper is for internal use, no need to care

sdk-core/src/main/java/io/milvus/client/AbstractMilvusGrpcClient.java
```diff
@@ -88,55 +86,56 @@ public abstract class AbstractMilvusGrpcClient implements MilvusClient {
 
     protected abstract String currentDbName();
 
+    protected abstract String currentEndpoint();
```
yhmo, line 89: sdk-core/src/main/java/io/milvus/client/AbstractMilvusGrpcClient.java line:89
High ---- Making this new hook abstract breaks every existing third-party subclass of AbstractMilvusGrpcClient: source upgrades must implement it, and already-compiled subclasses can throw AbstractMethodError as soon as cache-backed operations call currentEndpoint(). The prior thread's concrete fallback had isolation problems, but replacing it with an abstract method introduces a new binary/source compatibility failure; please provide a concrete per-instance-isolated default and let endpoint-aware subclasses override it.

yhmo: user shouldn't derive AbstractMilvusGrpcClient

sdk-core/src/main/java/io/milvus/v2/service/collection/CollectionService.java
```diff
@@ -102,6 +102,8 @@ public Void createCollection(MilvusServiceGrpc.MilvusServiceBlockingStub blockin
 
         Status status = blockingStub.createCollection(builder.build());
         rpcUtils.handleResponse(title, status);
+        invalidateSchemaCache(dbName, collectionName);
+        invalidateTsCache(dbName, collectionName);
```
liliu-z: The three create-collection paths invalidate the SESSION write timestamp after a successful create (CollectionService.java:106 and :184 via invalidateTsCache, AbstractMilvusGrpcClient.java:641 via removeTsCache). Because Milvus returns success when createCollection targets an already-existing, identical-schema collection, an idempotent "ensure exists" call made after a write wipes that collection's cached write timestamp, so the next SESSION query falls back to guaranteeTimestamp=1 and silently loses read-your-writes until a later write re-seeds it — and since CollectionTsCache is a JVM-wide singleton, one client's create can clear another client's live timestamp. dropCollection already clears the timestamp, so remove the ts invalidation from all three create paths while keeping the schema-cache invalidation.

PR #1974, opened 2026-07-29
Refine cache machinery

yhmo:
Refine schema and timestamp caching for V1 and V2 clients.

  - Add shared, endpoint-aware SchemaCache and CollectionTsCache.
  - Coalesce concurrent schema loads and prevent stale cache repopulation.
  - Invalidate caches after relevant DDL operations.
  - Preserve SESSION timestamps across rename and global-cluster failover.
  - Replace the legacy GTsDict.
  - Document client thread-safety expectations.
  - Add comprehensive cache and concurrency tests.


sdk-core/src/main/java/io/milvus/v2/utils/VectorUtils.java
```diff
@@ -127,19 +137,18 @@ public QueryRequest ConvertToGrpcQueryRequest(QueryReq request) {
...
+            return ts == 0L ? 1L : ts;
         }
         switch (consistencyLevel) {
             case STRONG:
                 return 0L;
             case SESSION: {
-                String key = GTsDict.CombineCollectionName(dbName, collectionName);
-                Long ts = GTsDict.getInstance().getCollectionTs(key);
-                return (ts == null) ? 1L : ts;
+                long ts = CollectionTsCache.getInstance().get(endpoint, cacheDbName, collectionName);
```
yhmo, line 150: sdk-core/src/main/java/io/milvus/v2/utils/VectorUtils.java line:150
Medium ---- `VectorUtils` still has a blank endpoint when callers instantiate this public converter directly, so this exact lookup cannot see timestamps that normal clients now store under `host:port`. The same caller pattern previously read the process-global `GTsDict`, meaning `new VectorUtils().ConvertToGrpc*Request(...)` with `SESSION` now silently emits guarantee timestamp `1` and loses read-your-writes. Please use the same blank-endpoint `getAnyEndpoint` compatibility fallback as `ParamUtils` (and add a regression test), or otherwise preserve the old endpoint-less behavior.

PR #1975, opened 2026-07-30
Refine cache machinery

yhmo:


sdk-core/src/main/java/io/milvus/v2/service/vector/VectorService.java
```diff
@@ -351,14 +330,10 @@ public DeleteResp delete(MilvusServiceGrpc.MilvusServiceBlockingStub blockingStu
         DeleteRequest rpcRequest = dataUtils.ConvertToGrpcDeleteRequest(request);
         MutationResult response = blockingStub.delete(rpcRequest);
 
-        // if illegal data, server fails to process delete, clean the schema cache
-        // so that the next call of dml can update the cache
-        cleanCacheIfFailed(response.getStatus(), dbName, collectionName);
         rpcUtils.handleResponse(title, response.getStatus());
```
yhmo, line 333: sdk-core/src/main/java/io/milvus/v2/service/vector/VectorService.java line:333
Medium ---- A delete-by-IDs request derives its filter from the cached primary-key name at lines 325-328. If another client drops and recreates this collection with a different primary-key name, this RPC returns a normal server error rather than SchemaMismatch; preserving the cache here makes every fresh delete-by-IDs request reuse the stale field indefinitely. Please invalidate or refresh on failures for the schema-dependent IDs path while keeping filter-only failures cached, and cover the recreation case.

sdk-core/src/main/java/io/milvus/param/ParamUtils.java
```diff
@@ -864,8 +864,13 @@ public static void compatibleSearchParams(Map<String, Object> searchParams, Sear
         }
     }
 
-    @SuppressWarnings("unchecked")
-    public static SearchRequest convertSearchParam(SearchParam requestParam) throws ParamException {
+    public static SearchRequest convertSearchParam(SearchParam requestParam, String endpoint,
```
yhmo, line 867: sdk-core/src/main/java/io/milvus/param/ParamUtils.java line:867
Medium ---- This replaces the shipped one-argument public method instead of overloading it, so source callers no longer compile and callers compiled against 2.6.x fail with NoSuchMethodError. Please retain deprecated one-argument overloads for convertSearchParam, convertHybridSearchParam, and convertQueryParam that delegate through a backward-compatible legacy cache scope; the same compatibility audit should cover the removed public GTsDict class.

sdk-core/src/main/java/io/milvus/orm/iterator/RpcStubWrapper.java
```diff
@@ -29,12 +30,21 @@ public class RpcStubWrapper {
...
-    private long rpcDeadlineMs = 0L;
+    private final long rpcDeadlineMs;
+    private final String endpoint;
+    private final String databaseName;
 
     public RpcStubWrapper(MilvusServiceGrpc.MilvusServiceBlockingStub blockingStub,
-                          long rpcDeadlineMs) {
-        this.blockingStub = blockingStub;
+                          long rpcDeadlineMs,
+                          String endpoint,
```
yhmo, line 39: sdk-core/src/main/java/io/milvus/orm/iterator/RpcStubWrapper.java line:39
Medium ---- This removes the public two-argument constructor that has shipped since the iterator deadline fix. External code constructing RpcStubWrapper or the public iterator classes will stop compiling, and precompiled code gets NoSuchMethodError on a 2.6 patch upgrade. Please keep a deprecated two-argument overload and define a backward-compatible cache identity or fallback for it.

sdk-core/src/main/java/io/milvus/client/AbstractMilvusGrpcClient.java
```diff
@@ -88,55 +86,56 @@ public abstract class AbstractMilvusGrpcClient implements MilvusClient {
 
     protected abstract String currentDbName();
 
+    protected abstract String currentEndpoint();
```
yhmo, line 89: sdk-core/src/main/java/io/milvus/client/AbstractMilvusGrpcClient.java line:89
Medium ---- Adding a new abstract method to this public base class breaks every external subclass on recompilation, and existing binaries can hit AbstractMethodError when the new cache paths call it; removing the protected cacheCollectionInfo field also breaks subclasses that referenced it. Please provide a non-abstract compatibility bridge and retain the protected surface as needed rather than changing the subclass contract in a 2.6 patch release.

sdk-core/src/main/java/io/milvus/v2/service/vector/VectorService.java
yhmo: coner case, ignore

sdk-core/src/main/java/io/milvus/param/ParamUtils.java
yhmo: ParamUtils is internal use, ignore

sdk-core/src/main/java/io/milvus/orm/iterator/RpcStubWrapper.java
yhmo: RpcStubWrapper is internal use, ignore

sdk-core/src/main/java/io/milvus/client/AbstractMilvusGrpcClient.java
yhmo: ignore

PR #1978, opened 2026-07-31
Add auto-publish workflow

yhmo:


.github/workflows/publish-release.yaml
```diff
@@ -0,0 +1,110 @@
...
+      - name: Publish to Maven Central
+        env:
+          MAVEN_CENTRAL_USERNAME: ${{ secrets.MAVEN_CENTRAL_USERNAME }}
+          MAVEN_CENTRAL_PASSWORD: ${{ secrets.MAVEN_CENTRAL_PASSWORD }}
+          MAVEN_GPG_PASSPHRASE: ${{ secrets.MAVEN_GPG_PASSPHRASE }}
+        run: >-
+          mvn --batch-mode --no-transfer-progress
+          -Prelease
+          -Dmaven.test.skip=true
+          clean deploy
```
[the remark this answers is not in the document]
yhmo: This repository uses central-publishing-maven-plugin with extensions=true. The plugin replaces the standard Maven Deploy Plugin execution during the deploy lifecycle, stages release artifacts into a bundle, and uploads it through the Central Portal API. Therefore, a release repository under distributionManagement or -DaltDeploymentRepository is not required. The configured snapshotRepository is only for snapshot publication.

.github/workflows/publish-release.yaml
```diff
@@ -0,0 +1,110 @@
...
+  group: maven-central-publish
+  cancel-in-progress: false
+
+jobs:
+  publish:
+    name: Publish Maven artifacts
+    if: github.repository == 'milvus-io/milvus-sdk-java'
+    runs-on: ubuntu-latest
+    environment: maven-central
+    timeout-minutes: 30
```
yhmo: .github/workflows/publish-release.yaml line:23
Medium ---- This job timeout can cancel a valid Central deployment before Maven finishes: central-publishing-maven-plugin 0.11.0 may wait up to 1800 seconds for validation after checkout, compilation, Javadocs, signing, and upload have already consumed part of the same 30-minute job budget. If GitHub kills the job after the bundle is accepted, Central can continue publishing while the dependent GitHub release never runs, and retrying the immutable version may fail. Please give the job more headroom than the plugin wait window, or configure a shorter plugin wait with explicit reconciliation.

.github/workflows/publish-release.yaml
yhmo, line 75: .github/workflows/publish-release.yaml line:74
Medium ---- The Central plugin default waitUntil is VALIDATED, so this deploy can return successfully after automatic publication is requested but before the artifacts reach PUBLISHED. The release job then publishes a GitHub release and triggers downstream release consumers while Maven Central may still be unavailable. This is distinct from the resolved repository-url thread: the plugin does handle deploy, but add waitUntil=published, for example via the POM or -DwaitUntil=published, before using job success as the release gate.

.github/workflows/publish-release.yaml
```diff
@@ -0,0 +1,110 @@
...
+
+      - name: Create release
+        uses: ncipollo/release-action@v1
+        with:
+          tag: ${{ github.ref_name }}
+          name: milvus-sdk-java-${{ steps.release-info.outputs.version }}
+          body: |
+            Release date: ${{ steps.release-info.outputs.release_date }}
+            Compatible with Milvus v${{ steps.release-info.outputs.milvus_minor }}.x
+          draft: false
```
yhmo, line 107: .github/workflows/publish-release.yaml line:106
Medium ---- This workflow sets the tag release to non-draft, but the existing .github/workflows/draft-release.yaml runs on the same tag and uses the same action with allowUpdates: true and draft: true. Because ncipollo/release-action applies draft on updates unless omitDraftDuringUpdate is set, whichever workflow or rerun updates last controls the final state; a successful Maven release can end up draft again and release events become order-dependent. Please retire the old workflow or sequence both stages in one workflow with a single owner of release state.

.github/workflows/publish-release.yaml
```diff
@@ -0,0 +1,110 @@
...
+            exit 1
+          fi
+
+          echo "Publishing milvus-sdk-java $project_version from tag $tag."
+
+      - name: Publish to Maven Central
+        env:
+          MAVEN_CENTRAL_USERNAME: ${{ secrets.MAVEN_CENTRAL_USERNAME }}
+          MAVEN_CENTRAL_PASSWORD: ${{ secrets.MAVEN_CENTRAL_PASSWORD }}
+          MAVEN_GPG_PASSPHRASE: ${{ secrets.MAVEN_GPG_PASSPHRASE }}
```
yhmo, line 70: .github/workflows/publish-release.yaml line:69
High ---- This passes a protected-key passphrase through setup-java's generated Maven settings, but the release profile still uses maven-gpg-plugin 1.6 without `--pinentry-mode loopback`. On `ubuntu-latest` with GPG 2.1+, that plugin invokes GPG with `--batch`/`--passphrase-fd`, but GPG delegates to pinentry unless loopback mode is enabled, so `clean deploy` can fail before any artifacts are uploaded. Please add the loopback `gpgArguments` to the release profile, as setup-java's Maven Central guidance requires, or use a signing setup proven non-interactive.

.github/workflows/publish-release.yaml
```diff
@@ -0,0 +1,110 @@
...
+    tags:
+      - 'v*.*.*'
+
+permissions:
+  contents: read
+
+# Maven Central releases are immutable. Do not publish two release tags at the same time,
+# and do not cancel a deployment after it has started uploading artifacts.
+concurrency:
+  group: maven-central-publish
```
yhmo, line 14: .github/workflows/publish-release.yaml line:14
Medium ---- A constant concurrency group serializes releases, but GitHub keeps only one pending run by default: if tag C arrives while A is publishing and B is pending, B is canceled and never reaches Maven Central. `cancel-in-progress: false` only protects A. Please use a durable multi-run queue, such as `queue: max` where available, or otherwise serialize without replacing pending release tags.

pom.xml
```diff
@@ -290,6 +296,7 @@
                 <configuration>
                     <publishingServerId>central</publishingServerId>
                     <autoPublish>true</autoPublish>
+                    <waitUntil>published</waitUntil>
```
yhmo, line 299: pom.xml line:299
Medium ---- The 60-minute job timeout fixes the earlier outer-timeout issue, but `waitUntil=published` still inherits this plugin version's 1800-second `waitMaxTime`. The plugin uploads before polling and throws on timeout; Central can finish afterward, leaving immutable artifacts published while the dependent GitHub release is skipped and a full rerun collides with the existing coordinates. Recent Central incidents have exceeded 30 minutes, so please set an explicit wait below the workflow limit (for example, `<waitMaxTime>3300</waitMaxTime>`) or add equivalent reconciliation.

.github/workflows/publish-release.yaml
```diff
@@ -0,0 +1,111 @@
...
+  queue: max
+  cancel-in-progress: false
+
+jobs:
+  publish:
+    name: Publish Maven artifacts
+    if: github.repository == 'milvus-io/milvus-sdk-java'
+    runs-on: ubuntu-latest
+    environment: maven-central
```
yhmo, line 23: .github/workflows/publish-release.yaml line:23
Low ---- This introduces a protected environment and four required Central/GPG secrets, but the repository has no release runbook describing how they are provisioned, which approvals apply, the exact tag procedure, or how to recover if Central accepts a bundle before a later step fails. Since tag pushes now publish immutable artifacts, please document those prerequisites and the manual reconciliation path before making this the release entry point.

PR #1982, opened 2026-07-31
Support client-side bloom filter blobs for bloom_match

### LINKED ISSUE #1981: Support client-side bloom filter blobs for bloom_match
## Background

`bloom_match(field, {blob})` is a new filter expression that tests approximate set membership against a **client pre-built** Split-Block Bloom Filter. It shipped server-side in milvus-io/milvus#51140 (merged 2026-07-28) and is on `master`.

The point is request size. A membership set of 10M int64 values is roughly 90 MB as a raw `in [...]` list, which blows past the proxy's gRPC receive limit. The same set as a bloom filter is about 14 MB at the default false-positive rate — the client builds the filter, ships the compact blob as an expression template parameter, and the proxy embeds it verbatim after validating the envelope. **The server never rebuilds the filter**, so every SDK must produce byte-identical blobs.

Design doc: [`docs/design-docs/design_docs/20260707-bloom-filter-expression.md`](https://github.com/milvus-io/milvus/blob/master/docs/design-docs/design_docs/20260707-bloom-filter-expression.md)

## What to implement

1. A public helper that turns a membership set (int64 or UTF-8 string) plus a false-positive rate into a blob.
2. A `bytes` branch in the SDK's existing expression-template-value conversion, so the blob travels as protobuf `TemplateValue.bytes_val` rather than being base64-inflated through a string field (the body is not valid UTF-8, so a string round trip corrupts it besides).
3. Unit tests that check the produced blob against the shared golden vectors.

## Wire format

The blob is a 32-byte **MBF1** envelope followed by a Parquet Split-Block Bloom Filter body. All integers little-endian.

| offset | size | field |
|---|---|---|
| 0  | 4 | magic `MBF1` |
| 4  | 2 | version = 1 |
| 6  | 2 | algo = 1 (parquet_sbbf_xxh64) |
| 8  | 8 | n_declared (uint64, informational) |
| 16 | 8 | fpr_declared (float64, informational) |
| 24 | 4 | num_blocks (uint32; body length must equal num_blocks × 32) |
| 28 | 1 | domains bitmask: `0x01` = int64, `0x02` = utf8 |
| 29 | 3 | reserved, must be 0 |
| 32 | … | body |

The body is bit-identical to Arrow C++'s `parquet::BlockSplitBloomFilter` and thus to the parquet-format `BloomFilter.md` spec:

- The body is a power-of-two number of 32-byte blocks; each block is eight little-endian `uint32` words.
- Values are hashed with **XXH64, seed 0**. An integer hashes its **8-byte little-endian two's-complement encoding**; a string hashes its **raw UTF-8 bytes**. (This matches Parquet plain encoding for INT64 / BYTE_ARRAY.)
- Block index is the multiply-shift reduction `((hash >> 32) * num_blocks) >> 32`.
- Within the block, for each word `i` in `0..7`, set bit `1 << ((uint32(hash) * SALT[i]) >> 27)`, where

  ```
  SALT = [0x47b6137b, 0x44974d91, 0x8824ad5b, 0xa2b7289d,
          0x705495c7, 0x2df1424b, 0x9efc4947, 0x5c6bfb31]
  ```

  Note `uint32(hash)` is the **low** 32 bits, the multiply wraps mod 2³², and the shift is **logical**.

Sizing mirrors Arrow's `BlockSplitBloomFilter::OptimalNumOfBytes`:

```
bits = -8 * n / ln(1 - fpr^(1/8))
bits = clamp(bits, 32*8, 128*1024*1024*8)   # MIN_FILTER_BYTES=32, MAX_FILTER_BYTES=128 MiB
bits = round_up_to_power_of_two(bits)
num_bytes = bits / 8
```

Bounds: `fpr ∈ [0.0001, 0.05]`, default `0.005`. An empty member set is legal — it records `domains = 0` and matches nothing.

**The `domains` bitmask matters.** The two hash domains share one XXH64 output space, so an 8-byte string and the int64 with the same byte image hash identically. Recording which domains were actually inserted is what keeps a probe from aliasing across domains, and it lets the proxy reject a blob built for the wrong field type instead of silently returning fewer rows. Build the blob from the **same value domain as the target field**: integer fields hash int64, VARCHAR fields hash UTF-8.

## Reference implementations

Both are already merged and produce identical bytes:

- **Go** — [`client/sbbf/sbbf.go`](https://github.com/milvus-io/milvus/blob/master/client/sbbf/sbbf.go) (builder + parser) and [`client/milvusclient/bloom_filter.go`](https://github.com/milvus-io/milvus/blob/master/client/milvusclient/bloom_filter.go) (public API + template wiring). This is the cleanest one to port from.
- **Python** — [`pymilvus/client/bloom_filter.py`](https://github.com/milvus-io/pymilvus/blob/master/pymilvus/client/bloom_filter.py) (milvus-io/pymilvus#3690).

## Acceptance criteria

- [ ] Blob is **byte-identical** to the shared golden vectors: [`client/sbbf/testdata/golden_vectors.json`](https://github.com/milvus-io/milvus/blob/master/client/sbbf/testdata/golden_vectors.json) (3 cases: int64-only, string-only, mixed-domain) and [`client/sbbf/testdata/cpp_generated_100_int64.json`](https://github.com/milvus-io/milvus/blob/master/client/sbbf/testdata/cpp_generated_100_int64.json) (100 int64 values generated by Arrow C++ itself). Copy both files into the SDK's test fixtures and compare full hex — this is the real interop check, and it is worth more than any self-consistent round-trip test.
- [ ] `fpr` outside `[0.0001, 0.05]`, NaN, and infinity are rejected.
- [ ] Members of mixed type are rejected by the convenience API (one call, one domain).
- [ ] The blob reaches the server as `TemplateValue.bytes_val`, verified by a test on the template-conversion function.
- [ ] A size-estimate helper (`n`, `fpr` → exact blob size) so callers can check against `proxy.maxBloomFilterSize` (64 MiB default) before building. Bodies are powers of two, so a count just past a tier boundary doubles the blob and a slightly higher `fpr` usually brings it back under.

## Protobuf dependency

`TemplateValue.bytes_val = 6` was added in milvus-io/milvus-proto#632 (commit `7efdf25`). SDKs pinned before that commit must bump milvus-proto and regenerate.

```proto
message TemplateValue {
  oneof val {
    bool bool_val = 1;
    int64 int64_val = 2;
    double float_val = 3;
    string string_val = 4;
    TemplateArrayValue array_val = 5;
    bytes bytes_val = 6;   // <-- needed
  };
}
```

## Status across SDKs

| SDK | status |
|---|---|
| Go | ✅ merged with the server PR (milvus-io/milvus#51140) |
| Python | 🔄 milvus-io/pymilvus#3690 open, awaiting review |
| Java | **this issue** |
| Node | 📋 milvus-io/milvus-sdk-node#597 |
| C++ | 📋 milvus-io/milvus-sdk-cpp#562 |
| Rust | 📋 milvus-io/milvus-sdk-rust#131 |

## Java specifics

**Landing spots**

- New `sdk-core/src/main/java/io/milvus/common/utils/BloomFilterUtils.java` — `io.milvus.common.utils` is where the version-agnostic public static helpers live (`Float16Utils` is the closest analogue), so it stays usable from both the v1 and v2 clients without dragging in grpc types.
- `io.milvus.v2.utils.VectorUtils.deduceAndCreateTemplateValue(Object)` currently throws `INVALID_PARAMS` for a `byte[]`; it needs a branch mapping `byte[]` → `setBytesVal(ByteString.copyFrom(...))`. All four call sites (query, search, hybrid ANN sub-request, delete) funnel through that one method, and `filterTemplateValues` is already `Map<String, Object>`, so no signature changes are needed.
- The `sdk-core/src/main/milvus-proto` submodule is pinned at `551b5154`, which predates `bytes_val`. It needs a bump to a milvus-proto commit ≥ `7efdf25`.

**No XXH64 on the classpath.** Guava's `com.google.common.hash.Hashing` offers murmur3 / sipHash / farmHash but not XXH64, and its `BloomFilter` is a different (murmur3-based, non-SBBF) layout that cannot produce this blob. parquet-column ships the right algorithm but is a `sdk-bulkwriter` dependency and drags in hadoop. So XXH64 should be hand-written in `sdk-core` — it is about 100 lines, and Java's `long` is two's complement with wrapping arithmetic, so only the shifts need care (`>>>`, never `>>`).

**Java 8 target** (`maven.compiler.source/target = 8`): no `var`, no `List.of`, no records.

Prefer a builder that takes a primitive `long` alongside the `List` convenience overload — a 10M-member `List<Long>` is roughly 240 MB of boxed objects on its own, which defeats the purpose of shipping a compact blob.

---

I have this implemented and tested locally (12 JUnit tests, blobs byte-identical to the Go and Python SDKs across all golden vectors, mutation-tested). PR to follow shortly.

> Note for anyone building this repo on Apple Silicon: `io.grpc:protoc-gen-grpc-java`'s `osx-aarch_64` artifact is actually an x86_64 binary in every published version, so `mvn` fails at codegen with `program not found or is not executable` unless Rosetta 2 is installed (`softwareupdate --install-rosetta`). Also, on JDK 26 the existing Mockito-based tests produce 93 pre-existing errors (`Cannot mock final class MilvusServiceGrpc$MilvusServiceBlockingStub`) — CI's JDK 11 is unaffected.

xiaofan-luan:
Closes #1981.

Adds the Java side of `bloom_match(field, {blob})`, the approximate-membership filter expression that shipped server-side in milvus-io/milvus#51140. A membership set of 10M int64 values is roughly 90 MB as a raw `in [...]` list and blows past the proxy's gRPC receive limit; the same set is about 14 MB as a bloom filter. The client builds the filter, ships the blob as an expression template parameter, and the proxy embeds it verbatim after validating the envelope — **the server never rebuilds it**, so the bytes have to match exactly across SDKs.

## What's here

- **`io.milvus.common.utils.BloomFilterUtils`** — `buildBloomFilter(List<?>, fpr)`, a size estimator, and a public `Builder`. `common/utils` is where the version-agnostic public static helpers live (`Float16Utils` is the closest analogue), so it stays usable from both the v1 and v2 clients without dragging in grpc types.
- **`io.milvus.common.utils.XXHash64`** — package-private, hand-written. sdk-core has no XXH64 on its classpath: Guava's `Hashing` offers murmur3 / sipHash / farmHash but not XXH64, and parquet-column's copy is a `sdk-bulkwriter` dependency behind hadoop. It is ~100 lines and Java's `long` already wraps mod 2^64, so only the shifts need care.
- **`VectorUtils.deduceAndCreateTemplateValue`** — a `byte[]` branch mapping to `TemplateValue.bytes_val`. All four call sites (query, search, hybrid ANN, delete) funnel through this one method and `filterTemplateValues` is already `Map<String, Object>`, so no public signature changed.

No proto submodule change is needed — `master`'s current pin (`9f7d27f`) already carries `bytes_val = 6` from milvus-io/milvus-proto#632.

## Usage

```java
byte[] blob = BloomFilterUtils.buildBloomFilter(userIds, BloomFilterUtils.DEFAULT_FPR);

QueryReq req = QueryReq.builder()
        .collectionName("docs")
        .filter("bloom_match(user_id, {bf})")
        .filterTemplateValues(Collections.singletonMap("bf", blob))
        .build();
```

For very large sets use the `Builder` directly — `addInt64(long)` takes a primitive, so a 10M-member filter costs no boxed `List<Long>` (which would be ~240 MB of `Long` objects on its own).

## Testing

12 new tests in `BloomFilterUtilsTest`. The load-bearing one compares the **full hex** of the produced blob against the golden vectors shared with the Go SDK (`client/sbbf/testdata/golden_vectors.json`, 3 cases including a mixed-domain one) and against `cpp_generated_100_int64.json`, which Arrow C++ generated itself. Both fixture files are copied into `sdk-core/src/test/resources/bloom/`. That is the real interop check; a self-consistent round-trip test would prove nothing about matching the server.

Also covered: XXH64 known-answer vectors cross-checked against the reference C implementation, the int64 fast path checked against the generic hash over 5000+ values, header layout, domain bitmask recording, integer widening (`Byte`/`Short`/`Integer`/`Long` produce identical filters), size-estimate agreement with actual output including the 128 MiB clamp, fpr boundaries, and rejection of mixed-type members / `Boolean` / nulls / out-of-range fpr.

Verified with `mvn -pl sdk-core test`: 12/12 green. The suite as a whole is unchanged against a pristine-`master` baseline — I ran both and the only difference is the 12 added tests. (On JDK 26 the existing Mockito-based tests produce 93 errors, `Cannot mock final class MilvusServiceGrpc$MilvusServiceBlockingStub`, on `master` and on this branch alike; CI's JDK 11 is unaffected.)

I also mutation-tested the implementation — corrupting a salt constant, the block-word byte order, the header version, the int64 encoding, or the domain bitmask each turn the golden-vector tests red.

https://claude.ai/code/session_01HWMAbqV46tWrVPhWRDedpG


sdk-core/src/test/java/io/milvus/common/utils/BloomFilterUtilsTest.java
```diff
@@ -0,0 +1,304 @@
...
+        // The generic hash has four length-dependent branches (32-byte stripes, then 8/4/1-byte
+        // tails). Lengths 0..80 walk every combination of them; the golden vectors above only
+        // reach a handful.
+        Random random = new Random(20260731L);
+        for (int length = 0; length <= 80; length++) {
+            byte[] data = new byte[length];
+            random.nextBytes(data);
+            long hash = XXHash64.hash(data);
+            // Re-hashing a copy must be stable, and the digest must not collapse to a constant.
+            Assertions.assertEquals(hash, XXHash64.hash(Arrays.copyOf(data, length)));
```
yhmo, line 130: sdk-core/src/test/java/io/milvus/common/utils/BloomFilterUtilsTest.java line:130
Medium ---- This assertion only compares XXHash64.hash with the same implementation on a copied input, so it cannot validate the stripe/merge code this test claims to cover. All fixed reference strings and shared golden fixtures are under 32 UTF-8 bytes, leaving long VARCHAR values unverified even though blob bytes must be cross-SDK identical. Please pin independent 31/32/33-byte vectors (as the PyMilvus counterpart does) or compare these payloads against an independent XXH64 implementation.

sdk-core/src/main/java/io/milvus/common/utils/BloomFilterUtils.java
```diff
@@ -0,0 +1,354 @@
...
+import java.nio.charset.StandardCharsets;
+import java.util.List;
+
+/**
+ * Builds a client-side Split-Block Bloom Filter (SBBF) blob for the
+ * <code>bloom_match(field, {blob})</code> filter expression.
+ *
+ * <p>Building the filter on the client and shipping the compact blob (about 32 MB for ~24M
+ * members at the default false-positive rate) lets large membership sets pass the proxy gRPC
+ * receive limit, which a raw value list (about 90 MB for 10M int64) would exceed. Pass the
```
yhmo: sdk-core/src/main/java/io/milvus/common/utils/BloomFilterUtils.java line:36
Low ---- This says a roughly 90 MB list would exceed the proxy receive limit, but the current Milvus master default is 128 MiB (and this class cites that value below in estimateBlobSize). Could we describe this as a scaling/overhead cliff instead of a current-limit violation, so callers are not given conflicting capacity guidance?

PR #1992, opened 2026-08-04
Redact credentials from connection configuration logs

yhmo:
- Redact passwords, tokens, authorization values, and URI user-info from V1 and V2 diagnostic output.
  - Keep standalone usernames visible for authentication and RBAC troubleshooting.
  - Sanitize global-cluster endpoint logs throughout connection and topology refresh paths.
  - Centralize redaction logic in a shared utility and add regression coverage.

### DIFF
--- a/sdk-core/src/main/java/io/milvus/common/utils/RedactCredential.java
+++ b/sdk-core/src/main/java/io/milvus/common/utils/RedactCredential.java
@@ -1,0 +1,36 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package io.milvus.common.utils;
+
+public final class RedactCredential {
+    private static final String REDACTED_CREDENTIAL = "<redacted>";
+
+    private RedactCredential() {
+    }
+
+    public static String redactCredential(String credential) {
+        return credential == null ? null : REDACTED_CREDENTIAL;
+    }
+
+    public static String redactUriUserInfo(String uri) {
+        return uri == null ? null : uri.replaceAll(
+                "^([^:/?#]+://)?[^/?#]*@", "$1" + REDACTED_CREDENTIAL + "@");
+    }
+}
--- a/sdk-core/src/main/java/io/milvus/param/credential/CreateCredentialParam.java
+++ b/sdk-core/src/main/java/io/milvus/param/credential/CreateCredentialParam.java
@@ -22,6 +22,8 @@ package io.milvus.param.credential;
 import io.milvus.exception.ParamException;
 import io.milvus.param.ParamUtils;
 
+import static io.milvus.common.utils.RedactCredential.redactCredential;
+
 public class CreateCredentialParam {
     private final String username;
     private final String password;
@@ -52,7 +54,7 @@ public class CreateCredentialParam {
     public String toString() {
         return "CreateCredentialParam{" +
                 "username='" + username + '\'' +
-                ", password='" + password + '\'' +
+                ", password='" + redactCredential(password) + '\'' +
                 '}';
     }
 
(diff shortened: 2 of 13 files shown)


sdk-core/src/test/java/io/milvus/param/CredentialRedactionTest.java
```diff
@@ -0,0 +1,87 @@
...
+                .build();
+
+        assertCredentialsRedacted(createParam.toString(),
+                "sensitive-user", "sensitive-password");
+        assertCredentialsRedacted(updateParam.toString(),
+                "sensitive-user", "sensitive-old-password", "sensitive-new-password");
+        assertTrue(createParam.toString().contains("username='<redacted>'"));
+        assertTrue(createParam.toString().contains("password='<redacted>'"));
+        assertTrue(updateParam.toString().contains("oldPassword='<redacted>'"));
+        assertTrue(updateParam.toString().contains("newPassword='<redacted>'"));
```
yhmo, line 117: sdk-core/src/test/java/io/milvus/param/CredentialRedactionTest.java line:79
Medium ---- This regression coverage stops at the legacy V1 credential params, while the active V2 CreateUserReq and UpdatePasswordReq still include userName/password/newPassword verbatim in their toString() output. That leaves V2 diagnostic output able to expose the same credentials this PR is intended to protect; please redact those request types and add corresponding V2 assertions.

sdk-core/src/main/java/io/milvus/param/credential/CreateCredentialParam.java
```diff
@@ -51,11 +53,15 @@ public String getPassword() {
     @Override
     public String toString() {
         return "CreateCredentialParam{" +
-                "username='" + username + '\'' +
-                ", password='" + password + '\'' +
+                "username='" + redactCredential(username) + '\'' +
```
yhmo: sdk-core/src/main/java/io/milvus/param/credential/CreateCredentialParam.java line:56
Medium ---- Redacting username here does not remove it from the actual V1 operation logs: AbstractMilvusGrpcClient still builds CreateCredentialRequest/UpdateCredentialRequest/DeleteCredentialRequest titles with requestParam.getUsername(), and handleResponse/logError emits those titles on success or failure. Since this PR explicitly treats usernames as credentials, these paths still leak them; please remove or redact the username in those diagnostic titles too.

sdk-core/src/main/java/io/milvus/v2/client/ConnectConfig.java
```diff
@@ -305,10 +307,10 @@ public Boolean isSecure() {
     @Override
     public String toString() {
         return "ConnectConfig{" +
-                "uri='" + uri + '\'' +
-                ", token='" + token + '\'' +
+                "uri='" + redactUriUserInfo(uri) + '\'' +
```
yhmo, line 310: sdk-core/src/main/java/io/milvus/v2/client/ConnectConfig.java line:310
High ---- This sanitizes only ConnectConfig.toString(), but MilvusClientV2.connect() still logs connectConfig.getUri() directly at INFO for global-cluster endpoints. A valid URI such as https://user:password@foo.global-cluster.example therefore bypasses this helper and exposes both values in production logs; please redact the URI at that log sink and cover the actual logging path.

sdk-core/src/test/java/io/milvus/v2/client/MilvusClientV2Test.java
```diff
@@ -504,6 +504,29 @@ void connectConfigDefaults() {
...
+                .dbName("default")
+                .build();
+
+        String loggedConfig = config.toString();
+        Assertions.assertFalse(loggedConfig.contains("sensitive-token"));
+        Assertions.assertFalse(loggedConfig.contains("sensitive-password"));
+        Assertions.assertFalse(loggedConfig.contains("uri-user"));
+        Assertions.assertFalse(loggedConfig.contains("uri-password"));
+        Assertions.assertTrue(loggedConfig.contains("token='<redacted>'"));
+        Assertions.assertTrue(loggedConfig.contains("username='sensitive-user'"));
```
yhmo, line 523: sdk-core/src/test/java/io/milvus/v2/client/MilvusClientV2Test.java line:523
Medium ---- Is plaintext username output intentional? The PR body explicitly promises to redact usernames, but this new assertion, together with the V1 and RBAC assertions, now locks in the opposite behavior. This is new evidence beyond the previously resolved delete-credential thread: the current head still exposes usernames across every credential/config renderer, so either redact those fields and update the assertions or narrow the stated security contract.

sdk-core/src/main/java/io/milvus/common/utils/RedactCredential.java
```diff
@@ -0,0 +1,36 @@
...
+    private RedactCredential() {
+    }
+
+    public static String redactCredential(String credential) {
+        return credential == null ? null : REDACTED_CREDENTIAL;
+    }
+
+    public static String redactUriUserInfo(String uri) {
+        return uri == null ? null : uri.replaceAll(
+                "^([^:/?#]+://)?[^@/]*@", "$1" + REDACTED_CREDENTIAL + "@");
```
yhmo: sdk-core/src/main/java/io/milvus/common/utils/RedactCredential.java line:34
Medium ---- This regex redacts only through the first @ and can therefore leave part of a credential visible for builder-accepted strings such as https://user:p@ss@host (it logs https://<redacted>@ss@host). It also treats an @ in a query or fragment as user-info, so https://host?owner=a@b becomes https://<redacted>@b. Please constrain matching to the authority and redact through its last @, with regression cases for both shapes.

PR #1995, opened 2026-08-05
fix: use array max capacity in iterator adapter

adrian-wang:
Use the schema's maxCapacity value instead of maxLength when converting V2 fields. Add a regression test with distinct varchar array length and capacity values.


yhmo:
good catch!

PR #2003, opened 2026-08-07
Align addFunctionField validation with pymilvus

yhmo:


sdk-core/src/main/java/io/milvus/v2/service/collection/CollectionService.java
```diff
@@ -786,25 +787,35 @@ public Void addFunctionField(MilvusServiceGrpc.MilvusServiceBlockingStub blockin
...
-        if (!StringUtils.equals(request.getFieldName(), indexParam.getFieldName())) {
+        if (StringUtils.isNotEmpty(indexParam.getFieldName())
+                && !StringUtils.equals(request.getFieldName(), indexParam.getFieldName())) {
             throw new MilvusClientException(ErrorCode.INVALID_PARAMS,
                     "Bound index field must match the function output field.");
         }
         if (indexParam.getIndexType() == null
-                || indexParam.getIndexType() == IndexParam.IndexType.None
-                || indexParam.getIndexType() == IndexParam.IndexType.AUTOINDEX) {
+                || indexParam.getIndexType() == IndexParam.IndexType.None) {
```
yhmo, line 820: sdk-core/src/main/java/io/milvus/v2/service/collection/CollectionService.java line:816
Medium ---- This now accepts the builder's default AUTOINDEX, but Milvus's PrepareFunctionOutputIndexParams explicitly rejects AUTOINDEX for function-output bound indexes ("an explicit index_type is required"). PyMilvus parity does not justify this: an omitted Python index_type is the empty string and is rejected, whereas omission in Java becomes AUTOINDEX. Consequently the new default/AcceptsAutoIndex path only reaches a server error despite the Javadoc and test advertising success. Please keep AUTOINDEX in this client-side rejection and restore the rejection test/documentation.

yhmo: no need

sdk-core/src/main/java/io/milvus/v2/service/collection/CollectionService.java
```diff
@@ -786,25 +787,35 @@ public Void addFunctionField(MilvusServiceGrpc.MilvusServiceBlockingStub blockin
         if (request.getFunction() == null) {
             throw new MilvusClientException(ErrorCode.INVALID_PARAMS, "Function cannot be null.");
         }
-        if (CollectionUtils.isEmpty(request.getFunction().getOutputFieldNames())
-                || request.getFunction().getOutputFieldNames().size() != 1
-                || !StringUtils.equals(request.getFieldName(), request.getFunction().getOutputFieldNames().get(0))) {
+        FunctionType functionType = request.getFunction().getFunctionType();
```
yhmo, line 794: sdk-core/src/main/java/io/milvus/v2/service/collection/CollectionService.java line:790
Medium ---- Removing the null-safe output-field guard changes an existing public-input failure contract: `Function.builder().outputFieldNames(null).build()` (also possible through the setter) now passes these checks, then `SchemaUtils.convertToGrpcFunction()` iterates the null list and `RpcUtils` reports `CLIENT_ERROR` before any RPC. Previously this path returned `INVALID_PARAMS`, and it contradicts the new Javadoc that output names are validated by the server. Please retain a representation-level null check (without restoring the field-name equality check) and cover it with a regression test.

PR #2005, opened 2026-08-07
Support async DQL interfaces

yhmo:
- Add CompletableFuture-based async APIs for query, search, and hybrid search.
  - Use gRPC future stubs with retry, deadline, global-cluster refresh, and cancellation support.
  - Add async APIs to client sessions and regression tests for success, errors, retries, and cancellation.

### DIFF
--- a/sdk-core/src/main/java/io/milvus/common/utils/cache/SchemaCache.java
+++ b/sdk-core/src/main/java/io/milvus/common/utils/cache/SchemaCache.java
@@ -25,6 +25,9 @@ import java.util.Comparator;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Objects;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionException;
+import java.util.concurrent.ExecutionException;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.concurrent.locks.ReentrantReadWriteLock;
@@ -35,6 +38,11 @@ public class SchemaCache {
     @FunctionalInterface
     public interface Loader {
         DescribeCollectionResponse load();
+    }
+
+    @FunctionalInterface
+    public interface AsyncLoader {
+        CompletableFuture<DescribeCollectionResponse> load();
     }
 
     private static final SchemaCache INSTANCE = new SchemaCache();
@@ -54,6 +62,7 @@ public class SchemaCache {
         private boolean completed;
         private DescribeCollectionResponse response;
         private Throwable failure;
+        private final CompletableFuture<DescribeCollectionResponse> future = new CompletableFuture<>();
     }
 
     private static final class LoadKey {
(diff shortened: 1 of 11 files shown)


sdk-core/src/main/java/io/milvus/v2/utils/RpcUtils.java
```diff
@@ -222,4 +236,226 @@ public <T> T retry(Callable<T> callable) {
...
+                future.cancel(true);
+            }
+        }
+
+        private void clearInFlight(CompletableFuture<?> future) {
+            inFlight.compareAndSet(future, null);
+        }
+
+        private void setScheduled(ScheduledFuture<?> future) {
+            ScheduledFuture<?> previous = scheduled.getAndSet(future);
```
liliu-z: `ScheduledFuture<?> previous = scheduled.getAndSet(future); previous.cancel(false);` assumes the stored handle always belongs to an older generation, but the handle is published only after `ASYNC_RETRY_EXECUTOR.schedule(...)` returns (~line 374), so the scheduled task can already be running. With `initialBackOffMs(0)` — accepted, since `RetryConfig` validates nothing and the multiplier keeps the interval at 0 — attempt N+1 can execute, fail and register attempt N+2's handle before the gRPC thread publishes N+1's; the late `getAndSet` then returns and cancels N+2, so the returned future never completes and the caller hangs permanently instead of getting an error, with up to 75 zero-delay attempts sampling the window repeatedly. Give each retry generation atomic ownership (e.g. CAS on a generation counter, or publish the handle before the task can run) and add a zero-backoff regression test.

sdk-core/src/main/java/io/milvus/common/utils/cache/SchemaCache.java
```diff
@@ -156,6 +163,64 @@ public DescribeCollectionResponse getOrLoad(String endpoint, String databaseName
...
+        }
+
+        LoadState loadingState = state;
+        loadedFuture.whenComplete((loaded, throwable) -> {
+            try {
+                if (throwable == null) {
+                    setCachedIfValid(key, loaded, loadingState);
+                    complete(loadingState, loaded, null);
+                } else {
+                    complete(loadingState, null, throwable);
```
liliu-z: `complete(loadingState, null, throwable);` runs inside the `try` while `removeLoading(loadKey, loadingState)` only happens in the `finally`, and because `awaitAsync` returns `state.future.thenApply(...)` with no executor, the entire dependent chain (VectorService → RetryFuture → user code) executes inline on the completing thread — strictly before the entry is removed. An application that retries immediately from that continuation (e.g. with SDK retries disabled) re-enters `getOrLoadAsync`, joins the already-failed `LoadState` still sitting in `loading`, and fails instantly without ever issuing another `describeCollection`; this now also affects the sync path, since `complete()` resolves `state.future` and a sync failure at line 155 fires async waiters' continuations inline before its own `finally`. Remove the loading entry before making the failure observable.

sdk-core/src/main/java/io/milvus/v2/service/vector/VectorService.java
```diff
@@ -196,39 +226,103 @@ private UpsertResp upsert(MilvusServiceGrpc.MilvusServiceBlockingStub blockingSt
...
+        }
+
+        if (CollectionUtils.isEmpty(request.getIds())) {
+            request.setDatabaseName(actualDbName(dbName));
+            return CompletableFuture.completedFuture(vectorUtils.ConvertToGrpcQueryRequest(request));
+        }
+
+        return getCollectionInfoAsync(futureStub, dbName, collectionName, false)
+                .thenApply(descResp -> {
+                    QueryReq effectiveRequest = copyQueryReq(request);
```
liliu-z: In the query-by-ids async path the caller's request object is not snapshotted at submission: `copyQueryReq(request)` and `request.getIds()` are read inside the `thenApply` that fires after the describeCollection round-trip, so any field the caller mutates in between is what actually goes on the wire. `MilvusClientV2Session.queryAsync` mutates the passed request (`request.setClusterId(clusterId)`), so a single-threaded caller reusing one QueryReq — `session("a").queryAsync(req)` followed immediately by `session("b").queryAsync(req)` — has the first call's clusterId, collectionName, outputFields and limit re-read as session b's values once the schema load returns, silently routing call a to the wrong cluster. Note the database name is not affected: `dbName` is captured in a local at invocation (line 269) and re-applied at line 284. Take the defensive copy synchronously before starting the schema load (and re-take it per retry attempt, since retryAsync re-invokes the supplier against the same live object).

sdk-core/src/main/java/io/milvus/common/utils/cache/SchemaCache.java
liliu-z: getOrLoadAsync has no unit test of its own. The only coverage is VectorServiceCacheTest.queryAsyncByIdsDoesNotBlockOnSchemaCacheMiss, a single-caller success path exercised through VectorService, while SchemaCacheTest covers the sync getOrLoad extensively including concurrency and failure cases. Nothing pins two async callers joining one LoadState, a sync caller joining an in-flight async load (and the reverse — the reason LoadState.future was added), or a failing load followed by an immediate retry, which is exactly the ordering the failure-visible-before-removal bug on this line breaks. Add those three cases so the shared-state handoff is regression-protected.

sdk-core/src/main/java/io/milvus/v2/service/vector/VectorService.java
```diff
@@ -196,39 +226,103 @@ private UpsertResp upsert(MilvusServiceGrpc.MilvusServiceBlockingStub blockingSt
...
+        List<FieldSchema> fields = descResp.getSchema().getFieldsList();
+        for (FieldSchema field : fields) {
+            if (field.getIsPrimaryKey()) {
+                return field.getName();
+            }
+        }
+        throw new MilvusClientException(ErrorCode.SERVER_ERROR, "cannot find the primary key field in collection schema");
+    }
+
+    private QueryReq copyQueryReq(QueryReq request) {
```
liliu-z, line 298: The `copyQueryReq` defensive copy fixes the sync bug where the SDK wrote collection metadata back into the caller's request object across retries, but no test pins that behaviour. A future refactor that drops or relocates the copy silently reintroduces caller-request mutation with a green CI; add a test that runs a query through a retry and asserts the caller's request instance is left unchanged.

sdk-core/src/main/java/io/milvus/v2/service/vector/VectorService.java
```diff
@@ -196,39 +226,103 @@ private UpsertResp upsert(MilvusServiceGrpc.MilvusServiceBlockingStub blockingSt
...
+        QueryRequest queryRequest = buildQueryRequest(blockingStub, request);
+        String title = String.format("Query collection: '%s' in database: '%s'",
+                request.getCollectionName(), queryRequest.getDbName());
+        QueryResults response = blockingStub.query(queryRequest);
+        return convertQueryResponse(title, response);
+    }
+
+    public CompletableFuture<QueryResp> queryAsync(MilvusServiceGrpc.MilvusServiceFutureStub futureStub,
+                                                   QueryReq request) {
+        return buildQueryRequestAsync(futureStub, request)
```
liliu-z: Cancelling a queryAsync future does not cancel the underlying gRPC call, unlike searchAsync and hybridSearchAsync. queryAsync returns the stage produced by thenCompose, whereas the cancel-forwarding behaviour lives on the CompletableFuture subclass that transformFuture returns; cancelling a composed stage does not propagate to its source or to the inner future, so RetryFuture.cancel only completes that dependent stage while the describeCollection and query RPCs keep running to completion on the server. The PR advertises cancellation support and tests it, but only on searchAsync, which happens to return the transformFuture instance directly. Either return a future that forwards cancellation into the active gRPC call for the query path too, or document that queryAsync cancellation is best-effort.

brycechesternewman: For our use case `searchAsync` is most likely the one we would cancel, so documenting `queryAsync` is fine for us, and returning an _optional_ future that forwards the cancelling is a nice to have. The only reason I can see the latter as beneficial is to release the server from holding the gRPC open. Or, perhaps include a TTL in the request, so if the the gRPC can auto cancel after X seconds, etc?

PR #2006, opened 2026-08-11
Support async DQL interfaces

yhmo:
- Add CompletableFuture-based async Query, Search, and Hybrid Search APIs.
  - Add async DQL methods to MilvusClientV2Session with session cluster routing.
  - Add non-blocking retry, backoff, and cancellation support.
  - Add asynchronous schema loading with shared waiter management.
  - Preserve client_request_id across async callbacks, schema loading, and retries.
  - Add async DQL examples with result validation and performance metrics.
  - Add focused tests for async requests, retries, cancellation, schema caching, and request metadata.

### DIFF
--- a/examples/src/main/java/io/milvus/v2/AsyncDQLExample.java
+++ b/examples/src/main/java/io/milvus/v2/AsyncDQLExample.java
@@ -1,0 +1,322 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package io.milvus.v2;
+
+import com.google.gson.Gson;
+import com.google.gson.JsonObject;
+import io.milvus.v2.client.ConnectConfig;
+import io.milvus.v2.client.MilvusClientV2;
+import io.milvus.v2.common.ConsistencyLevel;
+import io.milvus.v2.service.collection.request.CreateCollectionReq;
+import io.milvus.v2.service.collection.request.DropCollectionReq;
+import io.milvus.v2.service.vector.request.AnnSearchReq;
+import io.milvus.v2.service.vector.request.FunctionScore;
+import io.milvus.v2.service.vector.request.GetReq;
+import io.milvus.v2.service.vector.request.HybridSearchReq;
+import io.milvus.v2.service.vector.request.InsertReq;
+import io.milvus.v2.service.vector.request.QueryReq;
+import io.milvus.v2.service.vector.request.SearchReq;
+import io.milvus.v2.service.vector.request.data.FloatVec;
+import io.milvus.v2.service.vector.request.ranker.RRFRanker;
+import io.milvus.v2.service.vector.response.GetResp;
+import io.milvus.v2.service.vector.response.QueryResp;
+import io.milvus.v2.service.vector.response.SearchResp;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.Semaphore;
+import java.util.concurrent.atomic.AtomicLong;
+import java.util.function.Supplier;
+import java.util.stream.Collectors;
+
+public class AsyncDQLExample {
+    private static final String COLLECTION_NAME = "java_sdk_example_async_dql_v2";
+    private static final int DEFAULT_REQUEST_ROUNDS = 10000;
+    private static final int DEFAULT_MAX_CONCURRENT_REQUESTS = 50;
+    private static final int RESULT_LIMIT = 5;
+
+    public static void main(String[] args) {
+        int requestRounds = args.length > 0 ? Integer.parseInt(args[0]) : DEFAULT_REQUEST_ROUNDS;
+        int maxConcurrentRequests = args.length > 1
+                ? Integer.parseInt(args[1]) : DEFAULT_MAX_CONCURRENT_REQUESTS;
+        MilvusClientV2 client = new MilvusClientV2(ConnectConfig.builder()
+                .uri("http://localhost:19530")
+                .build());
+
+        try {
+            prepareCollection(client);
+            runAsyncQuery(client, requestRounds, maxConcurrentRequests);
+            runAsyncGet(client, requestRounds, maxConcurrentRequests);
+            runAsyncSearch(client, requestRounds, maxConcurrentRequests);
+            runAsyncHybridSearch(client, requestRounds, maxConcurrentRequests);
+        } finally {
+            try {
+                client.dropCollection(DropCollectionReq.builder()
+                        .collectionName(COLLECTION_NAME)
+                        .build());
+            } finally {
+                client.close();
+            }
+        }
+    }
+
+    private static void prepareCollection(MilvusClientV2 client) {
+        client.dropCollection(DropCollectionReq.builder()
+                .collectionName(COLLECTION_NAME)
+                .build());
+        client.createCollection(CreateCollectionReq.builder()
+                .collectionName(COLLECTION_NAME)
+                .dimension(4)
+                .build());
+
+        Gson gson = new Gson();
+        List<JsonObject> rows = new ArrayList<>();
+        for (int i = 0; i < 100; i++) {
+            JsonObject row = new JsonObject();
+            row.addProperty("id", i);
+            row.addProperty("name", "item-" + i);
+            row.add("vector", gson.toJsonTree(
+                    new float[]{i, (float) i / 2, (float) i / 3, (float) i / 4}));
+            rows.add(row);
+        }
+        client.insert(InsertReq.builder()
+                .collectionName(COLLECTION_NAME)
+                .data(rows)
+                .build());
+        System.out.println("Collection prepared with 100 rows\n");
+    }
+
+    private static void runAsyncQuery(MilvusClientV2 client, int requestRounds,
+                                      int maxConcurrentRequests) {
+        validateRunParameters(requestRounds, maxConcurrentRequests);
+        Semaphore concurrencyLimit = new Semaphore(maxConcurrentRequests);
+        List<CompletableFuture<?>> futures = new ArrayList<>(requestRounds);
+        long startNanos = System.nanoTime();
+        System.out.printf("Submitting %d async query requests, max concurrency=%d%n",
+                requestRounds, maxConcurrentRequests);
+
+        for (int i = 0; i < requestRounds; i++) {
+            int requestId = i;
+            AtomicLong requestStartNanos = new AtomicLong();
+            CompletableFuture<Void> callback = submitWithLimit(
+                    concurrencyLimit, () -> {
+                        requestStartNanos.set(System.nanoTime());
+                        return client.queryAsync(QueryReq.builder()
+                                .collectionName(COLLECTION_NAME)
+                                .filter(String.format("id >= %d and id < %d",
+                                        requestId % 95, requestId % 95 + 5))
+                                .outputFields(Arrays.asList("id", "name"))
+                                .consistencyLevel(ConsistencyLevel.STRONG)
+                                .build());
+                    }).thenAccept(response ->
+                    System.out.printf("Query request %d completed in %.2f ms, result: %s%n",
+                            requestId, elapsedMillis(requestStartNanos.get()),
+                            formatQueryResults(response)));
+            futures.add(callback);
+        }
+
+        CompletableFuture.allOf(futures.toArray(new CompletableFuture<?>[0])).join();
+        printThroughput("query", futures.size(), maxConcurrentRequests, startNanos);
+    }
+
+    private static void runAsyncGet(MilvusClientV2 client, int requestRounds,
+                                    int maxConcurrentRequests) {
+        validateRunParameters(requestRounds, maxConcurrentRequests);
+        Semaphore concurrencyLimit = new Semaphore(maxConcurrentRequests);
+        List<CompletableFuture<?>> futures = new ArrayList<>(requestRounds);
+        long startNanos = System.nanoTime();
+        System.out.printf("Submitting %d async get requests, max concurrency=%d%n",
+                requestRounds, maxConcurrentRequests);
+
+        for (int i = 0; i < requestRounds; i++) {
+            int requestId = i;
+            CompletableFuture<GetResp> future = submitWithLimit(concurrencyLimit,
+                    () -> client.getAsync(GetReq.builder()
+                            .collectionName(COLLECTION_NAME)
+                            .ids(Collections.singletonList((long) (requestId % 100)))
+                            .outputFields(Arrays.asList("id", "name"))
+                            .build()));
+            futures.add(future);
+        }
+
+        CompletableFuture.allOf(futures.toArray(new CompletableFuture<?>[0])).join();
+        printThroughput("get", futures.size(), maxConcurrentRequests, startNanos);
+    }
+
+    private static void runAsyncSearch(MilvusClientV2 client, int requestRounds,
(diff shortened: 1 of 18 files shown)


sdk-core/src/main/java/io/milvus/v2/service/vector/VectorService.java
```diff
@@ -196,39 +230,104 @@ private UpsertResp upsert(MilvusServiceGrpc.MilvusServiceBlockingStub blockingSt
...
+    }
+
+    private CompletableFuture<QueryResp> queryAsync(
+            MilvusServiceGrpc.MilvusServiceFutureStub futureStub,
+            QueryRequest baseRequest, String idsExpressionSuffix) {
+        CompletableFuture<QueryRequest> requestFuture;
+        if (idsExpressionSuffix == null) {
+            requestFuture = CompletableFuture.completedFuture(baseRequest);
+        } else {
+            requestFuture = getCollectionInfoAsync(futureStub, baseRequest.getDbName(),
```
yhmo: sdk-core/src/main/java/io/milvus/v2/service/vector/VectorService.java line:282
Medium ---- Cancelling a query-by-ID while its schema load is pending does not cancel the active `describeCollection` RPC. `composeFuture` cancels only the dependent future returned here, while `SchemaCache.dependentFuture()` intentionally prevents cancellation from reaching the shared loader, so abandoned requests can keep consuming RPC capacity even though this PR promises cancellation preservation. Please make loader cancellation reference-aware (cancel it when the last waiter cancels, while preserving other waiters) and cover this schema-stage case end to end.

sdk-core/src/main/java/io/milvus/v2/service/vector/VectorService.java
```diff
@@ -196,39 +230,104 @@ private UpsertResp upsert(MilvusServiceGrpc.MilvusServiceBlockingStub blockingSt
...
+                    baseRequest.getCollectionName(), false)
+                    .thenApply(descResp -> baseRequest.toBuilder()
+                            .setExpr(getPrimaryKeyName(descResp) + idsExpressionSuffix)
+                            .build());
+        }
+
+        return composeFuture(requestFuture, queryRequest -> {
+            String title = String.format("Query collection: '%s' in database: '%s'",
+                    queryRequest.getCollectionName(), queryRequest.getDbName());
+            return transformFuture(futureStub.query(queryRequest),
```
liliu-z, line 298: The `client_request_id` header is pulled from a ThreadLocal inside the channel interceptor at the moment the gRPC call is started, so it only lands on RPCs issued from the caller's own thread. The first attempt of searchAsync/hybridSearchAsync still starts on the caller thread and is fine, but the query RPC here is started inside the composeFuture continuation that runs after the schema load completes, and every scheduled retry starts its RPC on the shared single `milvus-async-retry` thread — both go out with no request id. Worse, SchemaCache completes waiters inline on whichever thread finished the shared load, so when a synchronous caller finishes a describeCollection that an async query-by-id joined, that async query is stamped with the synchronous caller's request id. Capture the id at async entry and attach it explicitly (CallOptions/Metadata) rather than reading it on the executing thread.

sdk-core/src/main/java/io/milvus/v2/utils/RpcUtils.java
```diff
@@ -222,4 +242,268 @@ public <T> T retry(Callable<T> callable) {
...
+
+        @Override
+        public boolean cancel(boolean mayInterruptIfRunning) {
+            boolean cancelled = super.cancel(mayInterruptIfRunning);
+            if (cancelled) {
+                CompletableFuture<?> active = inFlight.getAndSet(null);
+                if (active != null) {
+                    active.cancel(mayInterruptIfRunning);
+                }
+                ScheduledRetry pending = scheduled.getAndSet(null);
```
yhmo, line 481: sdk-core/src/main/java/io/milvus/v2/utils/RpcUtils.java line:481
Low ---- The tests cover cancelling an active attempt and verify remove-on-cancel, but they never cancel after a retry has entered a nonzero backoff. This CAS/late-ScheduledFuture path is the part that guarantees no RPC starts after cancellation; please add a focused test that fails attempt 1, cancels during backoff, and asserts the supplier remains at one invocation.

sdk-core/src/main/java/io/milvus/v2/client/MilvusClientV2.java
```diff
@@ -239,6 +248,7 @@ private void connect(ConnectConfig connectConfig) {
     private synchronized void updatePrimaryConnection(MilvusClientV2 primaryClient) {
         this.channel = primaryClient.channel;
         this.blockingStub = primaryClient.blockingStub;
+        this.futureStub = primaryClient.futureStub;
```
yhmo, line 251: sdk-core/src/main/java/io/milvus/v2/client/MilvusClientV2.java line:251
Low ---- This line and the per-attempt stub supplier are what preserve async global-cluster failover, but the new tests never swap primaries and verify a retry uses the replacement future stub. Please add a focused global-cluster test so a future refactor cannot leave async calls on the stale primary while synchronous routing still passes.

sdk-core/src/main/java/io/milvus/common/utils/cache/SchemaCache.java
```diff
@@ -125,35 +136,99 @@ public DescribeCollectionResponse getOrLoad(String endpoint, String databaseName
...
+                throw new NullPointerException("Async schema loader returned null future");
+            }
+        } catch (Throwable throwable) {
+            publishLoadResult(loadKey, state, null, throwable);
+            return dependent;
+        }
+
+        LoadState finalState = state;
+        synchronized (state) {
+            state.loaderFuture = loadFuture;
```
yhmo: sdk-core/src/main/java/io/milvus/common/utils/cache/SchemaCache.java line:218
Medium ---- Cancellation can race with this loader registration. If the sole dependent is cancelled after `loader.load()` returns but before `loaderFuture` is assigned, `releaseWaiter()` removes and invalidates the generation while seeing a null loader, so the describe RPC keeps running and a following caller starts a duplicate load. The earlier cancellation fix does not cover this window; please register the loader atomically with the cancellation state (and cancel immediately if the last waiter already left), with a focused boundary test.

examples/src/main/java/io/milvus/v2/AsyncDQLExample.java
```diff
@@ -0,0 +1,195 @@
...
+import io.milvus.v2.service.collection.request.DropCollectionReq;
+import io.milvus.v2.service.vector.request.AnnSearchReq;
+import io.milvus.v2.service.vector.request.FunctionScore;
+import io.milvus.v2.service.vector.request.HybridSearchReq;
+import io.milvus.v2.service.vector.request.InsertReq;
+import io.milvus.v2.service.vector.request.QueryReq;
+import io.milvus.v2.service.vector.request.SearchReq;
+import io.milvus.v2.service.vector.request.data.FloatVec;
+import io.milvus.v2.service.vector.request.ranker.RRFRanker;
+import io.milvus.v2.service.vector.response.QueryResp;
```
yhmo, line 39: examples/src/main/java/io/milvus/v2/AsyncDQLExample.java line:37
Low ---- `QueryResp` is not referenced anywhere in this example, so this import is unnecessary and should be removed.

sdk-core/src/main/java/io/milvus/v2/client/MilvusClientV2.java
```diff
@@ -114,6 +116,7 @@ public class MilvusClientV2 {
     private static final Logger logger = LoggerFactory.getLogger(MilvusClientV2.class);
     private ManagedChannel channel;
     private MilvusServiceGrpc.MilvusServiceBlockingStub blockingStub;
+    private MilvusServiceGrpc.MilvusServiceFutureStub futureStub;
```
yhmo: sdk-core/src/main/java/io/milvus/v2/client/MilvusClientV2.java line:119
Medium ---- The replacement test added for the earlier failover thread invokes updatePrimaryConnection synchronously, but production calls it from TopologyRefresher's separate executor. This field is then read by caller/retry threads without volatile or the writer's monitor, so Java provides no happens-before edge and an async retry can keep using the stale or closed primary. Please safely publish the replacement (for example, make futureStub volatile or synchronize the read) and make the test perform the swap from another thread.

sdk-core/src/main/java/io/milvus/v2/utils/RpcUtils.java
```diff
@@ -222,4 +242,268 @@ public <T> T retry(Callable<T> callable) {
...
+        long nextInterval = nextRetryIntervalMs;
+        ScheduledRetry scheduledRetry = result.registerScheduled();
+        try {
+            ScheduledFuture<?> scheduled = ASYNC_RETRY_EXECUTOR.schedule(() -> {
+                        if (result.startScheduled(scheduledRetry)) {
+                            attemptAsync(supplier, result, begin, maxRetryTimes,
+                                    attemptNumber + 1, nextInterval);
+                        }
+                    },
+                    retryIntervalMs,
```
yhmo, line 383: sdk-core/src/main/java/io/milvus/v2/utils/RpcUtils.java line:383
Low ---- RetryConfig accepts negative backoff values, and ScheduledThreadPoolExecutor treats this negative delay as immediately eligible. The synchronous retry path instead reaches TimeUnit.MILLISECONDS.sleep and rejects it, so retryAsync can hot-loop all attempts for a malformed-but-type-valid config despite documenting the same retry policy. Please validate non-negative backoffs (preferably in RetryConfig) or reject them before scheduling, with a parity test.

sdk-core/src/main/java/io/milvus/v2/service/vector/VectorService.java
```diff
@@ -270,14 +360,38 @@ public SearchResp search(MilvusServiceGrpc.MilvusServiceBlockingStub blockingStu
...
+            SearchReq request, String clusterId, RpcUtils retryUtils) {
+        final SearchRequest searchRequest;
+        try {
+            searchRequest = vectorUtils.ConvertToGrpcSearchRequest(request).toBuilder()
+                    .setDbName(actualDbName(request.getDatabaseName()))
+                    .build();
+        } catch (Throwable throwable) {
+            return failedFuture(throwable);
+        }
+        final SearchRequest effectiveRequest = withSearchClusterId(searchRequest, clusterId);
```
yhmo, line 370: sdk-core/src/main/java/io/milvus/v2/service/vector/VectorService.java line:377
Medium ---- This helper removes an existing cluster_id before adding the session value, but the synchronous search path above still appends the session value directly. Since SearchReq.searchParams forwards raw keys, a request already containing cluster_id sends one value asynchronously and two conflicting values synchronously. Please route both paths through withSearchClusterId and cover raw-param parity.

sdk-core/src/main/java/io/milvus/v2/utils/RpcUtils.java
```diff
@@ -222,4 +242,268 @@ public <T> T retry(Callable<T> callable) {
...
+                        statusException);
+                logger.error(msg);
+                result.completeExceptionally(new MilvusClientException(ErrorCode.RPC_ERROR, statusException));
+                return;
+            }
+            handleGlobalConnectionError(statusException);
+            retryable = true;
+        } else if (cause instanceof MilvusClientException) {
+            MilvusClientException clientException = (MilvusClientException) cause;
+            retryable = retryConfig.isRetryOnRateLimit()
```
yhmo, line 327: sdk-core/src/main/java/io/milvus/v2/utils/RpcUtils.java line:327
Medium ---- `retryConfig` and `globalRefreshTrigger` in RpcUtils are still plain (non-volatile) fields, but the new async path reads them from the shared `milvus-async-retry` executor thread and from gRPC callback threads: `handleAsyncFailure()`/`retryTimedOut()` read `retryConfig` fields on every retry, and `handleGlobalConnectionError`/`handleGlobalRoutingError` read `globalRefreshTrigger`. The setters run on the caller thread, and the first attempt's failure is handled on the gRPC completion thread, so there is no formal happens-before edge. This PR already made `blockingStub`/`futureStub` volatile for exactly this cross-thread-visibility pattern; please apply the same to `retryConfig` and `globalRefreshTrigger` (or capture them into the async flow) so a concurrent `retryConfig(...)`/`setGlobalRefreshTrigger(...)` call is consistently observed.

sdk-core/src/main/java/io/milvus/v2/service/vector/VectorService.java
```diff
@@ -261,23 +351,40 @@ public SearchResp search(MilvusServiceGrpc.MilvusServiceBlockingStub blockingStu
...
+            SearchReq request, String clusterId, RpcUtils retryUtils) {
+        final SearchRequest searchRequest;
+        try {
+            searchRequest = vectorUtils.ConvertToGrpcSearchRequest(request).toBuilder()
+                    .setDbName(actualDbName(request.getDatabaseName()))
                     .build();
+        } catch (Throwable throwable) {
+            return failedFuture(throwable);
         }
+        final SearchRequest effectiveRequest = withSearchClusterId(searchRequest, clusterId);
```
yhmo, line 370: sdk-core/src/main/java/io/milvus/v2/service/vector/VectorService.java line:370
Medium ---- `searchAsync`/`hybridSearchAsync` build `effectiveRequest` once and reuse it across every retry attempt, and `queryAsync` freezes `baseRequest` the same way. That means the SESSION/BOUNDED-consistency `guarantee_timestamp` (read from CollectionTsCache inside `ConvertToGrpcSearchRequest`/`ConvertToGrpcQueryRequest`) is also frozen at the first attempt. The synchronous path recomputes the timestamp on each retry because `search()`/`query()` re-run the conversion inside `retry()`. So if a write lands during the retry backoff, a sync retry picks up the newer timestamp while the async retry keeps the stale one and can miss the write. Please confirm this stale-timestamp-on-retry behavior is intentional; if parity with the sync path is desired, the conversion (or at least the guarantee timestamp) needs to be refreshed per attempt.

sdk-core/src/test/java/io/milvus/v2/client/MilvusClientV2SessionDockerTest.java
```diff
@@ -0,0 +1,391 @@
...
+    private static final String COLLECTION_NAME = "session_test_collection";
+    private static final int DIMENSION = 4;
+    private static final File DockerComposeFile = TestUtils.dockerComposeFile("docker-compose.yml");
+    private static final File DockerComposeVolumeDirectory = new File("target/milvus-compose");
+    private static final List<String> DockerComposeContainerNames = Arrays.asList(
+            "milvus-javasdk-etcd", "milvus-javasdk-minio", "milvus-javasdk-standalone");
+
+    @BeforeAll
+    public static void setUp() {
+        TestUtils.startMilvusStandalone(DockerComposeFile, DockerComposeVolumeDirectory, DockerComposeContainerNames);
```
yhmo, line 63: sdk-core/src/test/java/io/milvus/v2/client/MilvusClientV2SessionTest.java line:63
Low ---- This new test launches a real Milvus standalone via `TestUtils.startMilvusStandalone` (docker compose) in `@BeforeAll`, but it does not follow the repository naming convention: the build-and-test docs state that server-backed tests are named `*DockerTest` (e.g. `MilvusClientV2DockerTest`) and auto-launch the server. Since Surefire runs every `*Test` class, this test now boots a full Milvus stack during ordinary unit-test runs / CI (`mvn -Pcoverage verify`) wherever Docker is available, and fails where it is not. Consider renaming to `MilvusClientV2SessionDockerTest` (or gating the server startup) so the unit build stays hermetic.

sdk-core/src/main/java/io/milvus/v2/service/vector/VectorService.java
yhmo: as designed

sdk-core/src/main/java/io/milvus/v2/utils/RpcUtils.java
```diff
@@ -222,4 +242,268 @@ public <T> T retry(Callable<T> callable) {
...
+        } else if (cause instanceof Error) {
+            result.completeExceptionally(cause);
+            return;
+        } else {
+            result.completeExceptionally(new MilvusClientException(ErrorCode.CLIENT_ERROR, cause));
+            return;
+        }
+
+        long now = System.currentTimeMillis();
+        if (retryTimedOut(begin, now)) {
```
yhmo, line 346: sdk-core/src/main/java/io/milvus/v2/utils/RpcUtils.java line:346
Low ---- The async retry timeout branches (`retryTimedOut`, `completeAsyncTimeout`, and the predicted-backoff-exceeds-timeout check at line 359) have no direct test, while the sync path is covered by `testEarlyExitWhenPredictedBackoffExceedsMaxRetryTimeoutMs` and the slow-call TIMEOUT test. The new RpcUtilsTest cases cover success, retry, no-retry, cancellation, and backoff exhaustion, but never drive a real `maxRetryTimeoutMs` early exit, so a regression that breaks or inverts the async timeout check would go unnoticed (codecov also flags RpcUtils at 62.42% patch coverage). Please add a focused async test that fails attempts until the retry window elapses and asserts the future completes exceptionally with `ErrorCode.TIMEOUT`.

PR #2011, opened 2026-08-13
Refine clusterId for MilvusClientV2Session

yhmo:


sdk-core/src/main/java/io/milvus/v2/service/vector/VectorService.java
```diff
@@ -241,6 +261,14 @@ public SearchResp search(MilvusServiceGrpc.MilvusServiceBlockingStub blockingStu
         // reset the db name so that the timestamp cache can set correct key for this collection
         request.setDatabaseName(actualDbName(dbName));
         SearchRequest searchRequest = vectorUtils.ConvertToGrpcSearchRequest(request);
+        if (StringUtils.isNotEmpty(clusterId)) {
+            searchRequest = searchRequest.toBuilder()
```
yhmo, line 265: sdk-core/src/main/java/io/milvus/v2/service/vector/VectorService.java line:265
Medium ---- The converter has already serialized `request.getClusterId()` before this block, so a session call with a pre-populated request (for example, request `cluster-b` through session `cluster-a`) sends two `cluster_id` entries instead of overriding the request value. The regression test at `testSessionOverridesRequestClusterIdAndRejectsClosed` reads the first matching parameter, so it will still see `cluster-b`, and server routing is left dependent on duplicate-key ordering. Please remove or replace the existing `cluster_id` entry before adding the session value (and apply the same fix to query and hybrid search).

sdk-core/src/main/java/io/milvus/v2/utils/VectorUtils.java
```diff
@@ -107,13 +107,6 @@ public QueryRequest ConvertToGrpcQueryRequest(QueryReq request) {
                     .build());
         });
 
-        if (StringUtils.isNotEmpty(request.getClusterId())) {
```
yhmo, line 110: sdk-core/src/main/java/io/milvus/v2/utils/VectorUtils.java line:110
High ---- Removing these request-level `clusterId` conversions also removes routing for the existing iterator paths: `MilvusClientV2Session.queryIterator/searchIterator/searchIteratorV2` still put the session ID into their request objects, and `QueryIterator`, `SearchIterator`, and `SearchIteratorV2` build their page RPCs through these converters. The current iterator tests expect `cluster-a` on those RPCs, so session iterators will now send no cluster ID; please preserve the converter path or add an equivalent session-aware iterator path.

yhmo, line 110: sdk-core/src/main/java/io/milvus/v2/utils/VectorUtils.java line:110
Medium ---- This removal changes the behavior of the still-public `QueryReq/ SearchReq/ HybridSearchReq.clusterId` setters for direct `MilvusClientV2.query/search/hybridSearch` calls: those values are now silently ignored, even though the fields and builders remain available (only deprecated). That is a backward-incompatible semantic change for existing callers. Please either keep request-level serialization for direct calls or fail/redirect explicitly, and define how it should interact with the session value.

yhmo: as design

PR #2014, opened 2026-08-17
fix: avoid sending empty analyzer params in runAnalyzer

### LINKED ISSUE #2013: [Bug]runAnalyzer fails when using collection mode without analyzerParams
### Problem

Calling `runAnalyzer` with a `collectionName` (and optional `fieldName`) but **without** `analyzerParams` fails with:

```
run analyzer can't use analyzer params and (collection,field) in same time
```

This makes the collection-based mode of `runAnalyzer` unusable.

### Steps to reproduce

```java
MilvusClientV2 client = new MilvusClientV2(ConnectConfig.builder()
        .uri("http://<host>:19530")
        .dbName("<db>")
        .build());

RunAnalyzerResp resp = client.runAnalyzer(RunAnalyzerReq.builder()
        .databaseName("<db>")
        .collectionName("<collection>")
        .fieldName("page_content")
        .texts(List.of("some text"))
        .build());
```

### Expected

The analyzer configured on the `page_content` field of the collection is used to analyze the texts and tokens are returned.

### Actual

The request is rejected by the server with the error above.

### Root cause
RunAnalyzerReqBuilder` initializes `analyzerParams` to an empty `HashMap` (not null). So when the user does not call `.analyzerParams()`, `getAnalyzerParams()` returns an empty map, and
     `JsonUtils.toJson(emptyMap)` (backed by `Gson.toJson`) returns the string `"{}"`, which is sent to the server as a non-empty `analyzer_params` field. The server then treats it as
     conflicting with the collection/field mode.

### Environment

- Milvus server: 2.6.8
- milvus-sdk-java: 2.6.22 / 2.6.23 / 3.0.6 (reproduced on all)

zch0214:
# fix: avoid sending empty analyzer params in runAnalyzer

Fixes #2013

See also #2017 for the master branch.

## Problem
  `VectorService.runAnalyzer()` unconditionally calls `setAnalyzerParams(JsonUtils.toJson(request.getAnalyzerParams()))`. Since `RunAnalyzerReqBuilder` initializes `analyzerParams` to an
     empty `HashMap`, the collection-based mode (passing `collectionName` / `fieldName` without `analyzerParams`) results in `analyzer_params = "{}"`, which the server rejects with:

```
run analyzer can't use analyzer params and (collection,field) in same time
```

## Fix

Only set `analyzer_params` on the gRPC request when `analyzerParams` is non-empty. The `!isEmpty()` check is the operative fix (the builder always initializes `analyzerParams` to an empty map); the `!= null` check is defensive-only.

## Verification

- Before: `runAnalyzer(collectionName=..., fieldName="page_content", texts=[...])` fails with the error above.
- After: the same request succeeds, while the `analyzerParams`-only mode still works as before.

### DIFF
--- a/sdk-core/src/main/java/io/milvus/v2/service/vector/VectorService.java
+++ b/sdk-core/src/main/java/io/milvus/v2/service/vector/VectorService.java
@@ -377,14 +377,16 @@ public class VectorService extends BaseService {
 
         String params = JsonUtils.toJson(request.getAnalyzerParams());
         logger.debug(params);
-        RunAnalyzerRequest runRequest = builder.addAllPlaceholder(byteStrings)
-                .setAnalyzerParams(params)
+        RunAnalyzerRequest.Builder runRequestBuilder = builder.addAllPlaceholder(byteStrings)
                 .setWithDetail(request.getWithDetail())
                 .setWithHash(request.getWithHash())
                 .setDbName(request.getDatabaseName())
                 .setCollectionName(request.getCollectionName())
-                .setFieldName(request.getFieldName())
-                .build();
+                .setFieldName(request.getFieldName());
+        if (request.getAnalyzerParams() != null && !request.getAnalyzerParams().isEmpty()) {
+            runRequestBuilder.setAnalyzerParams(params);
+        }
+        RunAnalyzerRequest runRequest = runRequestBuilder.build();
         RunAnalyzerResponse response = blockingStub.runAnalyzer(runRequest);
         rpcUtils.handleResponse(title, response.getStatus());
 
--- a/sdk-core/src/main/java/io/milvus/v2/service/vector/VectorService.java
+++ b/sdk-core/src/main/java/io/milvus/v2/service/vector/VectorService.java
@@ -380,7 +380,7 @@ public class VectorService extends BaseService {
         RunAnalyzerRequest.Builder runRequestBuilder = builder.addAllPlaceholder(byteStrings)
                 .setWithDetail(request.getWithDetail())
                 .setWithHash(request.getWithHash())
-                .setDbName(request.getDatabaseName())
+                .setDbName(actualDbName(request.getDatabaseName()))
                 .setCollectionName(request.getCollectionName())
                 .setFieldName(request.getFieldName());
         if (request.getAnalyzerParams() != null && !request.getAnalyzerParams().isEmpty()) {
--- a/sdk-core/src/test/java/io/milvus/v2/client/MilvusClientV2DockerTest.java
+++ b/sdk-core/src/test/java/io/milvus/v2/client/MilvusClientV2DockerTest.java
@@ -4239,6 +4239,50 @@ class MilvusClientV2DockerTest {
     }
 
     @Test
+    void testRunAnalyzerCollectionMode() {
+        String collectionName = generator.generate(10);
+        CreateCollectionReq.CollectionSchema collectionSchema = CreateCollectionReq.CollectionSchema.builder()
+                .build();
+        collectionSchema.addField(AddFieldReq.builder()
+                .fieldName("id")
+                .dataType(DataType.Int64)
+                .isPrimaryKey(Boolean.TRUE)
+                .autoID(Boolean.FALSE)
+                .build());
+        collectionSchema.addField(AddFieldReq.builder()
+                .fieldName("vector")
+                .dataType(DataType.FloatVector)
+                .dimension(DIMENSION)
+                .build());
+        Map<String, Object> analyzerParams = new HashMap<>();
+        analyzerParams.put("tokenizer", "standard");
+        collectionSchema.addField(AddFieldReq.builder()
+                .fieldName("text")
+                .dataType(DataType.VarChar)
+                .maxLength(100)
+                .enableAnalyzer(true)
+                .analyzerParams(analyzerParams)
+                .build());
+        client.createCollection(CreateCollectionReq.builder()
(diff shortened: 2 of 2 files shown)


sdk-core/src/main/java/io/milvus/v2/service/vector/VectorService.java
```diff
@@ -377,14 +377,16 @@ public RunAnalyzerResp runAnalyzer(MilvusServiceGrpc.MilvusServiceBlockingStub b
...
-                .setAnalyzerParams(params)
+        RunAnalyzerRequest.Builder runRequestBuilder = builder.addAllPlaceholder(byteStrings)
                 .setWithDetail(request.getWithDetail())
                 .setWithHash(request.getWithHash())
                 .setDbName(request.getDatabaseName())
                 .setCollectionName(request.getCollectionName())
-                .setFieldName(request.getFieldName())
-                .build();
+                .setFieldName(request.getFieldName());
+        if (request.getAnalyzerParams() != null && !request.getAnalyzerParams().isEmpty()) {
```
yhmo, line 386: sdk-core/src/main/java/io/milvus/v2/service/vector/VectorService.java line:386
Low ---- The guard is correct, but the PR description’s stated root cause doesn’t match the default path: RunAnalyzerReqBuilder initializes analyzerParams to an empty HashMap, so in the #2013 repro (no .analyzerParams() call) getAnalyzerParams() returns an empty map and JsonUtils.toJson(...) yields "{}", not "null". The !isEmpty() branch is the operative fix; the != null check is effectively defensive-only given the builder default. Please update the description so the real trigger is documented and a future maintainer doesn’t "simplify" the condition back to a null-only check and reintroduce the bug.

sdk-core/src/main/java/io/milvus/v2/service/vector/VectorService.java
```diff
@@ -377,14 +377,16 @@ public RunAnalyzerResp runAnalyzer(MilvusServiceGrpc.MilvusServiceBlockingStub b
...
+            runRequestBuilder.setAnalyzerParams(params);
```
yhmo, line 387: sdk-core/src/main/java/io/milvus/v2/service/vector/VectorService.java line:387
Medium ---- No regression test covers the behavior this PR fixes. The existing testRunAnalyzer Docker test only exercises the analyzerParams-only path (no collectionName/fieldName), so an unconditional setAnalyzerParams would silently regress without being caught. Per CODE_REVIEW.md, bug-fix PRs should include test cases covering the bug; adding a collection+field-mode case (with and without analyzerParams) to MilvusClientV2DockerTest would close this gap.

sdk-core/src/main/java/io/milvus/v2/service/vector/VectorService.java
```diff
@@ -377,14 +377,16 @@ public RunAnalyzerResp runAnalyzer(MilvusServiceGrpc.MilvusServiceBlockingStub b
 
         String params = JsonUtils.toJson(request.getAnalyzerParams());
         logger.debug(params);
-        RunAnalyzerRequest runRequest = builder.addAllPlaceholder(byteStrings)
-                .setAnalyzerParams(params)
+        RunAnalyzerRequest.Builder runRequestBuilder = builder.addAllPlaceholder(byteStrings)
```
yhmo, line 380: sdk-core/src/main/java/io/milvus/v2/service/vector/VectorService.java line:380
Low ---- Follow-up: the same bug still exists on master/3.0, where runAnalyzer() calls setAnalyzerParams(params) unconditionally, and issue #2013 reports the failure on 3.0.6. If this PR is intended to fix #2013, consider landing the equivalent change on master (and any other supported release line) so the issue is fully closed across SDK versions.

zch0214: OK

sdk-core/src/main/java/io/milvus/v2/service/vector/VectorService.java
zch0214: Thanks! You're right — RunAnalyzerReqBuilder initializes analyzerParams to an empty HashMap, so the actual trigger is JsonUtils.toJson(emptyMap) yielding "{}". I've updated the PR/issue │ description accordingly, and kept the !isEmpty() guard as the operative check.

sdk-core/src/main/java/io/milvus/v2/service/vector/VectorService.java
zch0214: Added testRunAnalyzerCollectionMode to MilvusClientV2DockerTest, which covers the collection+field mode without analyzerParams. It fails before this fix and passes after.

sdk-core/src/main/java/io/milvus/v2/service/vector/VectorService.java
zch0214: Opened a follow-up PR targeting master with the same fix + test: #2017

sdk-core/src/main/java/io/milvus/v2/service/vector/VectorService.java
yhmo: The issue is confirmed fixed. Resolving this conversation.

sdk-core/src/main/java/io/milvus/v2/service/vector/VectorService.java
yhmo: The issue is confirmed fixed. Resolving this conversation.

PR #2015, opened 2026-08-17
Support async DQL interfaces

yhmo:


sdk-core/src/main/java/io/milvus/v2/client/MilvusClientV2.java
```diff
@@ -1591,6 +1670,8 @@ public UpdateReplicateConfigurationResp updateReplicateConfiguration(UpdateRepli
      * @throws InterruptedException throws InterruptedException if the client failed to close connection
      */
     public void close(long maxWaitSeconds) throws InterruptedException {
+        // cancel any pending async retries and release the per-client retry scheduler
+        rpcUtils.shutdown();
```
yhmo, line 1680: sdk-core/src/main/java/io/milvus/v2/client/MilvusClientV2.java line:1674
Medium ---- This shutdown() does not reliably stop an active async retry cycle. If an attempt is in flight when close() runs, its failure (UNAVAILABLE/CANCELLED on the shutdown channel) is classified as retryable, and RpcUtils.asyncRetryExecutor() lazily recreates the just-shutdown scheduler before scheduling the next attempt, so the cycle keeps running in the background up to maxRetryTimes. Note that master guards this differently: its close() nulls futureStub/blockingStub and its stub getters throw "MilvusClient is closed", while this backport neither nulls futureStub nor guards getFutureRpcStub(), so async calls issued after close() hit the terminated channel and churn retries instead of failing fast. Consider nulling futureStub in close() and throwing in getFutureRpcStub() when it is null (matching master), or explicitly refusing to recreate the scheduler once shutdown has been requested.

sdk-core/src/test/java/io/milvus/v2/utils/RpcUtilsTest.java
```diff
@@ -26,10 +26,214 @@
...
+        rpcUtils.shutdown();
+        TimeUnit.MILLISECONDS.sleep(200);
+
+        Assertions.assertEquals(1, callCount.get(),
+                "shutdown should cancel the pending scheduled retry");
+        result.cancel(true);
+    }
+
+    @Test
+    void testRetryAsyncRecreatesSchedulerAfterShutdown() throws Exception {
```
yhmo: sdk-core/src/test/java/io/milvus/v2/utils/RpcUtilsTest.java line:215
Low ---- This test does not actually exercise the lazy scheduler recreation it claims to verify. A successfully completed immediate future never touches asyncRetryExecutor(), because the executor is only reached on the retryable-failure scheduling path in handleAsyncFailure. The assertions would still pass even if asyncRetryExecutor() never recreated a scheduler after shutdown(). To genuinely cover the stated behavior, trigger a retryable failure after shutdown() (e.g. retryConfig with a small backoff) so the reschedule path forces recreation.

sdk-core/src/main/java/io/milvus/v2/utils/VectorUtils.java
```diff
@@ -684,7 +684,8 @@ public HybridSearchRequest ConvertToGrpcHybridSearchRequest(HybridSearchReq requ
     }
 
     private FunctionSchema convertFunctionSchema(CreateCollectionReq.Function function) {
-        Map<String, String> params = function.getParams();
+        Map<String, String> functionParams = function.getParams();
+        Map<String, String> params = functionParams == null ? new HashMap<>() : new HashMap<>(functionParams);
```
yhmo, line 688: sdk-core/src/main/java/io/milvus/v2/utils/VectorUtils.java line:688
Low ---- This convertFunctionSchema null-params hardening looks unrelated to the async DQL feature. The PR also has no description and no linked issue (CONTRIBUTING.md asks for the issue reference in the PR description, and CODE_REVIEW.md expects the change to match the title/commit). Consider splitting this defensive fix into a separate small PR, or at least documenting the mixed scope in the description so the async DQL change can be reviewed in isolation.

yhmo: keep it

sdk-core/src/main/java/io/milvus/v2/service/vector/VectorService.java
```diff
@@ -231,6 +264,83 @@ public QueryResp query(MilvusServiceGrpc.MilvusServiceBlockingStub blockingStub,
...
+
+    private QueryRequest buildBaseQueryRequest(QueryReq request) {
+        return vectorUtils.ConvertToGrpcQueryRequest(request).toBuilder()
+                .setDbName(actualDbName(request.getDatabaseName()))
+                .build();
+    }
+
+    private String buildIdsExpressionSuffix(List<?> ids) {
+        String expression = vectorUtils.getExprById("", ids);
+        return expression.substring(expression.indexOf(" in ["));
```
yhmo: sdk-core/src/main/java/io/milvus/v2/service/vector/VectorService.java line:322
Low ---- buildIdsExpressionSuffix derives the suffix by locating the literal " in [" in the output of getExprById("", ids). This couples the async ids path to the exact formatting of that helper; if its format ever changes (or a future refactor routes an empty ids list here), indexOf returns -1 and substring(-1) throws StringIndexOutOfBoundsException at runtime. Since the primary key name is only known after the async schema load, consider resolving the expression inside the transform (passing the primary key name into getExprById) instead of slicing the preformatted expression with substring.

sdk-core/src/main/java/io/milvus/v2/utils/RpcUtils.java
```diff
@@ -148,7 +192,7 @@ public <T> T retry(Callable<T> callable) {
                         || code == io.grpc.Status.UNIMPLEMENTED.getCode()) {
                     String msg = String.format("Encounter rpc error that cannot be retried, reason: %s", e);
                     logger.error(msg);
-                    throw new MilvusClientException(ErrorCode.RPC_ERROR, msg); // throw rpc error
+                    throw new MilvusClientException(ErrorCode.RPC_ERROR, e); // throw rpc error
```
yhmo, line 195: sdk-core/src/main/java/io/milvus/v2/utils/RpcUtils.java line:195
Low ---- This sync retry() change swaps the thrown MilvusClientException from a descriptive message ("Encounter rpc error that cannot be retried, reason: ...") to the raw StatusRuntimeException as the cause, so exception.getMessage() now becomes the bare gRPC status text. On the maintenance 2.6 line this is an observable change for existing callers that log or match on the message; the new test asserts the cause is preserved, so the intent is clear, but consider noting the message change in the release notes since callers may depend on the previous text.

sdk-core/src/main/java/io/milvus/v2/service/vector/VectorService.java
```diff
@@ -360,6 +517,39 @@ public GetResp get(MilvusServiceGrpc.MilvusServiceBlockingStub blockingStub, Get
...
+        }
+        // call queryAsync to get the result
+        return transformFuture(
+                queryAsync(futureStubSupplier, queryReq, retryUtils),
+                queryResp -> GetResp.builder()
+                        .getResults(queryResp.getQueryResults())
+                        .build());
+    }
+
+    private QueryReq toQueryReq(GetReq request) {
```
yhmo, line 529: sdk-core/src/main/java/io/milvus/v2/service/vector/VectorService.java line:541
Low ---- toQueryReq duplicates the QueryReq construction already inlined in the sync get() (same databaseName/collectionName/ids/outputFields mapping). Two sibling copies of this conversion will drift on the next GetReq change, and note that the existing partitionName field is currently ignored by both. Consider refactoring get() to reuse toQueryReq so there is a single conversion path to keep in sync.

sdk-core/src/main/java/io/milvus/v2/utils/RpcUtils.java
yhmo: as designed

PR #2022, opened 2026-08-18
Clean async tasks when shutdown

yhmo:


sdk-core/src/main/java/io/milvus/v2/utils/RpcUtils.java
```diff
@@ -494,6 +505,22 @@ private void clearScheduled(ScheduledRetry retry) {
...
         }
 
+        /**
+         * Cancels the active attempt and any pending retry, then completes exceptionally.
+         * Used by {@link RpcUtils#shutdown()} so in-flight retries are not left hanging.
+         */
+        private void fail(Throwable throwable) {
+            CompletableFuture<?> active = inFlight.getAndSet(null);
+            if (active != null) {
+                active.cancel(true);
```
yhmo: sdk-core/src/main/java/io/milvus/v2/utils/RpcUtils.java line:515
Medium ---- When shutdown() runs while a retryAsync attempt is still in flight, fail() cancels the active attempt before completing the future. Cancelling the attempt synchronously fires the whenComplete callback registered in attemptAsync() while result.isDone() is still false, so handleAsyncFailure() completes the future with a MilvusClientException(CLIENT_ERROR) wrapping a CancellationException instead of the intended "MilvusClient is closed"; the later completeExceptionally(throwable) in fail() then returns false and the intended exception is lost. Callers blocked on get()/join() are unblocked (the goal), but they observe a misleading cancellation error. The existing cancel() override completes the future first (super.cancel) and only then cancels the in-flight attempt; fail() should mirror that ordering (completeExceptionally() before active.cancel(true)). Please also add a test with a still-in-flight attempt at shutdown time - the current test only covers the pending-scheduled-retry path where inFlight is already null.

sdk-core/src/main/java/io/milvus/v2/utils/RpcUtils.java
```diff
@@ -494,6 +505,24 @@ private void clearScheduled(ScheduledRetry retry) {
             }
         }
 
+        /**
+         * Cancels the active attempt and any pending retry, then completes exceptionally.
```
yhmo: sdk-core/src/main/java/io/milvus/v2/utils/RpcUtils.java line:509
Medium ---- The first sentence of this javadoc ("Cancels the active attempt and any pending retry, then completes exceptionally.") states the opposite order of what the code does and contradicts the third sentence of the same comment: completeExceptionally() runs first, and only then are the in-flight attempt and the pending scheduled retry cancelled. This ordering is concurrency-critical - completing first is what prevents the attempt whenComplete callback from re-entering handleAsyncFailure() with a CancellationException, the exact issue the resolved thread at line 518 was about. A maintainer who trusts the first sentence could reorder the method and reintroduce that bug. Suggest rewording the first sentence, e.g. "Completes exceptionally first, then cancels the active attempt and any pending retry."

sdk-core/src/main/java/io/milvus/v2/utils/RpcUtils.java
```diff
@@ -494,6 +505,24 @@ private void clearScheduled(ScheduledRetry retry) {
...
+         * Completes the future first so the attempt's whenComplete callback sees it done and
+         * returns early instead of running handleAsyncFailure with a CancellationException.
+         */
+        private void fail(Throwable throwable) {
+            completeExceptionally(throwable);
+            CompletableFuture<?> active = inFlight.getAndSet(null);
+            if (active != null) {
+                active.cancel(true);
+            }
+            ScheduledRetry pending = scheduled.getAndSet(null);
```
yhmo, line 525: sdk-core/src/main/java/io/milvus/v2/utils/RpcUtils.java line:520
Low ---- fail() duplicates the in-flight/scheduled cancellation block already present in cancel(): both do inFlight.getAndSet(null) then cancel, and scheduled.getAndSet(null) then cancel. Consider extracting a small shared private helper (e.g. cancelPendingAttempts(boolean mayInterruptIfRunning)) so the two teardown paths cannot drift apart if interrupt-vs-cancel semantics ever change in only one place.

PR #2026, opened 2026-08-19
Add cherry-pick workflow

yhmo:


.github/workflows/cherry-pick.yml
```diff
@@ -0,0 +1,138 @@
...
+          git push origin --delete "$BRANCH" 2>/dev/null || true
+          git checkout -b "$BRANCH" "origin/$TO"
+          echo "branch=$BRANCH" >> "$GITHUB_OUTPUT"
+          echo "short_sha=$SHORT_SHA" >> "$GITHUB_OUTPUT"
+        env:
+          TO: ${{ github.event.inputs.to_branch }}
+          COMMIT: ${{ github.event.inputs.commit_id }}
+
+      - name: Cherry-pick the commit
+        run: git cherry-pick "$COMMIT"
```
yhmo: .github/workflows/cherry-pick.yml line:100
Medium ---- git cherry-pick "$COMMIT" fails opaquely on this workflow's most common failure modes. The "already on target" guard at line 76 only checks ancestry by the original SHA, so a commit whose content was already applied to the target under a different SHA (e.g. after this workflow's own PR was squash-merged) passes validation and then dies here with git's "previous cherry-pick is now empty" error and no cleanup; conflicts fail the same opaque way. Consider handling the empty/conflict cases explicitly (git cherry-pick --abort plus a clear "already applied / manual backport required" message) and use git cherry-pick -x so the new commit records the "(cherry picked from commit <sha>)" trailer, which also makes the already-applied case detectable on rerun. pymilvus's auto-cherrypick workflow already does this.

.github/workflows/cherry-pick.yml
```diff
@@ -0,0 +1,165 @@
...
+            exit 1
+          fi
+          # HEAD is already checked out to the target branch, so use it instead of a
+          # remote-tracking ref (origin/$TO) which may not exist locally.
+          if git merge-base --is-ancestor "$COMMIT" HEAD; then
+            echo "::error::commit '$COMMIT' is already on branch '$TO'"
+            exit 1
+          fi
+          FULL_SHA=$(git rev-parse "$COMMIT")
+          if git log HEAD --format=%B | grep -q "cherry picked from commit ${FULL_SHA}"; then
```
yhmo: .github/workflows/cherry-pick.yml line:90
Medium ---- This "already cherry-picked onto TO" guard does not reliably fire on a repository this size because of SIGPIPE under pipefail. GitHub Actions' default `run` shell is `bash -e -o pipefail`, and `git log HEAD --format=%B` here is ~100 KB over 906 commits, larger than the 64 KB pipe buffer. When the trailer exists, `grep -q` exits 0 immediately, `git log` is killed by SIGPIPE (exit 141), and pipefail makes the pipeline return 141, so the `if` evaluates false exactly when a match was found. I reproduced this against the PR head: pipeline status 141 (ps=141 0) for both a top-of-log and a deep match, and status 1 only when the pattern is absent. Consider `git log HEAD --format=%B --grep="cherry picked from commit ${FULL_SHA}"` (no `grep -q` early-exit) so the check is deterministic. Note line 118 (`git ls-files -u | grep -q .`) has the same pattern and would misreport if a conflict list ever exceeded the pipe buffer.

.github/workflows/cherry-pick.yml
```diff
@@ -0,0 +1,165 @@
...
+          COMMIT: ${{ github.event.inputs.commit_id }}
+          RUN_ID: ${{ github.run_id }}
+
+      - name: Cherry-pick the commit
+        run: |
+          if ! git cherry-pick -x "$COMMIT" 2>/tmp/cherry_pick_err; then
+            if git ls-files -u | grep -q .; then
+              echo "::error::cherry-pick of '$COMMIT' has conflicts; a manual backport is required"
+            else
+              echo "::error::commit '$COMMIT' appears to already be applied on '$TO' (cherry-pick produced an empty change)"
```
yhmo, line 132: .github/workflows/cherry-pick.yml line:121
Low ---- The else branch labels every non-conflict `git cherry-pick` failure as "already applied (empty change)", but the most likely other failure is a merge commit: `commit_id` is free-form, and `git cherry-pick -x` on a merge commit fails with "is a merge but no -m option was given" without leaving unmerged files, so this branch reports the commit as "already applied". pymilvus's auto-cherrypick workflow handles this by retrying with `-m 1`. Consider detecting merge commits (e.g. `git rev-list --parents -n 1 "$COMMIT"`) or falling back to `-m 1` so the failure reason is accurate.

.github/workflows/cherry-pick.yml
```diff
@@ -0,0 +1,165 @@
...
+          COMMIT_TITLE=$(git show -s --format=%s "$COMMIT")
+          COMMIT_BODY=$(git show -s --format=%b "$COMMIT")
+          TITLE="[cherry-pick] ${COMMIT_TITLE}"
+          BODY_FILE=$(mktemp)
+          printf 'Cherry-picks commit `%s` (%s) from `%s` to `%s`.\n\n' "$COMMIT" "$SHORT_SHA" "$FROM" "$TO" > "$BODY_FILE"
+          printf 'Created by the cherry-pick workflow — please review and merge manually.\n' >> "$BODY_FILE"
+          if [ -n "$COMMIT_BODY" ]; then
+            printf '\n%s\n' "$COMMIT_BODY" >> "$BODY_FILE"
+          fi
+          if gh pr create --base "$TO" --head "$BRANCH" --title "$TITLE" --body-file "$BODY_FILE" 2>/tmp/pr_err; then
```
yhmo: .github/workflows/cherry-pick.yml line:154
Low ---- Re-dispatching the same (from_branch, to_branch, commit_id) while a previous run's PR is still open creates a second, duplicate PR. Each run gets a unique branch name (`RUN_ID` suffix, line 106) and the line 90 guard only inspects merged history of HEAD, so the fallback `gh pr list --head "$BRANCH"` here can never match the earlier PR (different branch name) and `gh pr create` succeeds again. pymilvus's auto-cherrypick workflow searches for an existing open backport PR by title before creating a new one; consider a similar open-PR check before `gh pr create`.

PR #2028, opened 2026-08-19
Reject JSON/Geometry/Timestamptz in Struct field

yhmo:
JSON/Geometry/Timestamptz in Struct field ------- allowed during creating collection, rejected during insert/upsert
This behavior is aligned with pymilvus.

### DIFF
--- a/sdk-core/src/main/java/io/milvus/v2/utils/DataUtils.java
+++ b/sdk-core/src/main/java/io/milvus/v2/utils/DataUtils.java
@@ -562,6 +562,12 @@ public class DataUtils {
                             .build())
                     .build();
         } else {
+            // JSON/Geometry/Timestamptz are not supported as struct sub-field element types
+            // (aligned with pymilvus, which rejects them on insert)
+            if (dataType == DataType.JSON || dataType == DataType.Geometry || dataType == DataType.Timestamptz) {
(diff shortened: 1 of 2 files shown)


sdk-core/src/main/java/io/milvus/v2/utils/DataUtils.java
```diff
@@ -562,6 +562,12 @@ private static FieldData genStructSubFieldData(CreateCollectionReq.FieldSchema f
                             .build())
                     .build();
         } else {
+            // Geometry/Timestamptz are not supported as struct sub-field element types
+            // (aligned with pymilvus, which rejects them on insert)
+            if (dataType == DataType.Geometry || dataType == DataType.Timestamptz) {
```
yhmo: sdk-core/src/main/java/io/milvus/v2/utils/DataUtils.java line:567
Low ---- Question about parity completeness: pymilvus rejects not only Geometry and Timestamptz but also JSON as a struct sub-field element type on insert -- in its type registry these are the element types without an array-element representation, so convert_to_array() raises "Unsupported element type" for JSON sub-fields too. This check only covers Geometry and Timestamptz, while genScalarField() still serializes a JSON sub-field (case JSON -> JSONArray) on insert. If a JSON struct sub-field can be declared in the SDK schema (StructFieldSchema.addField only blocks Array/Struct), should JSON also be rejected here for full parity with pymilvus, or is JSON intentionally supported on the wire?

PR #2029, opened 2026-08-19
Adjust cherry-pick workflow

yhmo:
Use my token to submit pr

### DIFF
--- a/.github/workflows/cherry-pick.yml
+++ b/.github/workflows/cherry-pick.yml
@@ -9,6 +9,13 @@
 #   5. opens a PR from that branch into the target branch
 #
 # The PR is NOT auto-merged; merge it manually after review.
+#
+# The branch push and PR creation authenticate with the repository secret
+# `CHERRY_PICK_PAT` (a personal access token owned by a trusted member). This makes
+# the PR author a real collaborator instead of github-actions[bot], so CI runs
+# on the new PR without requiring "Approve workflows to run". Create a PAT with
+# `repo` scope (or fine-grained `contents: write` + `pull-requests: write`,
+# scoped to this repository) and store it as `CHERRY_PICK_PAT`.
 
 name: Cherry-pick Commit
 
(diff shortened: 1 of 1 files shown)


.github/workflows/cherry-pick.yml
```diff
@@ -142,19 +150,30 @@ jobs:
...
         env:
           BRANCH: ${{ steps.branch.outputs.branch }}
+          PUSH_TOKEN: ${{ secrets.CHERRY_PICK_PAT }}
+        run: |
+          if [ -z "$PUSH_TOKEN" ]; then
+            echo "::error::secrets.CHERRY_PICK_PAT is not configured; add a PAT with 'repo' scope (contents + PR write) and store it as CHERRY_PICK_PAT"
+            exit 1
+          fi
+          # push with the tokenized URL inline so the PAT is never written into .git/config
+          git push "https://x-access-token:${PUSH_TOKEN}@github.com/${GITHUB_REPOSITORY}.git" "$BRANCH"
```
yhmo, line 165: .github/workflows/cherry-pick.yml line:162
High ---- This push will not authenticate with CHERRY_PICK_PAT. The checkout step (actions/checkout@v4, default persist-credentials: true) leaves `http.https://github.com/.extraheader = AUTHORIZATION: basic <GITHUB_TOKEN>` in .git/config, and that Authorization header takes precedence over the credentials embedded in the push URL, so git sends the GITHUB_TOKEN header on the first request and never falls back to the PAT. With the new `permissions: contents: read` that token cannot push: GitHub returns 403 (authenticated but read-only), git does not retry URL credentials on a 403, and the job fails before "Create pull request" runs. This is a known checkout-extraheader behavior (see actions/checkout#162); the sibling pymilvus auto-cherrypick workflow avoids it by setting `persist-credentials: false` on checkout. Please set `persist-credentials: false` on the checkout step (or run `git config --unset-all http.https://github.com/.extraheader` before the push) and verify with one end-to-end run that the pushed branch and PR are authored by the PAT owner.

PR #2031, opened 2026-08-20
Align with pymilvus request/response behavior

yhmo:
Additive changes                                                                                                                                                                 
- CreateCollectionReq.FieldSchema — fieldId/isDynamic/isFunctionOutput                               
- CreateCollectionReq.Function — id/inputFieldIds/outputFieldIds                                     
- DescribeCollectionResp — aliases/updateTimestamp/enableNamespace/schemaVersion                     
- GetCompactionPlansResp — compactionId                                                              
- GetQuerySegmentInfoResp.QuerySegmentInfo — collectionName                                          
- ListAliasResp — dbName                                                                             
- RefreshExternalCollectionJobInfo — externalSpec                                                    
- InsertResp/UpsertResp/DeleteResp — cost                  
- QueryResp — cost                                                                                   
- GrantPrivilegeReq — dbName                                                                         
- GetReq — partitionNames

### DIFF
--- a/sdk-core/src/main/java/io/milvus/v2/service/rbac/RBACService.java
+++ b/sdk-core/src/main/java/io/milvus/v2/service/rbac/RBACService.java
@@ -126,15 +126,18 @@ public class RBACService extends BaseService {
 
     public Void grantPrivilege(MilvusServiceGrpc.MilvusServiceBlockingStub blockingStub, GrantPrivilegeReq request) {
         String title = String.format("Grant privilege for role: '%s'", request.getRoleName());
-        GrantEntity entity = GrantEntity.newBuilder()
+        GrantEntity.Builder entityBuilder = GrantEntity.newBuilder()
                 .setRole(RoleEntity.newBuilder()
                         .setName(request.getRoleName())
                         .build())
                 .setObjectName(request.getObjectName())
                 .setObject(ObjectEntity.newBuilder().setName(request.getObjectType()).build())
                 .setGrantor(GrantorEntity.newBuilder()
-                        .setPrivilege(PrivilegeEntity.newBuilder().setName(request.getPrivilege()).build()).build())
-                .build();
+                        .setPrivilege(PrivilegeEntity.newBuilder().setName(request.getPrivilege()).build()).build());
+        if (StringUtils.isNotEmpty(request.getDbName())) {
+            entityBuilder.setDbName(request.getDbName());
+        }
+        GrantEntity entity = entityBuilder.build();
         OperatePrivilegeRequest operatePrivilegeRequest = OperatePrivilegeRequest.newBuilder()
                 .setEntity(entity)
                 .setType(OperatePrivilegeType.Grant)
@@ -147,15 +150,18 @@ public class RBACService extends BaseService {
 
     public Void revokePrivilege(MilvusServiceGrpc.MilvusServiceBlockingStub blockingStub, RevokePrivilegeReq request) {
         String title = String.format("Revoke privilege for role: '%s'", request.getRoleName());
-        GrantEntity entity = GrantEntity.newBuilder()
+        GrantEntity.Builder entityBuilder = GrantEntity.newBuilder()
                 .setRole(RoleEntity.newBuilder()
                         .setName(request.getRoleName())
                         .build())
                 .setObjectName(request.getObjectName())
                 .setObject(ObjectEntity.newBuilder().setName(request.getObjectType()).build())
                 .setGrantor(GrantorEntity.newBuilder()
-                        .setPrivilege(PrivilegeEntity.newBuilder().setName(request.getPrivilege()).build()).build())
-                .build();
+                        .setPrivilege(PrivilegeEntity.newBuilder().setName(request.getPrivilege()).build()).build());
+        if (StringUtils.isNotEmpty(request.getDbName())) {
+            entityBuilder.setDbName(request.getDbName());
+        }
+        GrantEntity entity = entityBuilder.build();
         OperatePrivilegeRequest operatePrivilegeRequest = OperatePrivilegeRequest.newBuilder()
                 .setEntity(entity)
                 .setType(OperatePrivilegeType.Revoke)
--- a/sdk-core/src/main/java/io/milvus/v2/service/vector/VectorService.java
+++ b/sdk-core/src/main/java/io/milvus/v2/service/vector/VectorService.java
@@ -49,6 +49,7 @@ import org.slf4j.LoggerFactory;
 
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.atomic.AtomicReference;
@@ -165,6 +166,7 @@ public class VectorService extends BaseService {
         return InsertResp.builder()
                 .InsertCnt(response.getInsertCnt())
                 .primaryKeys(ids)
+                .cost(getCost(response.getStatus()))
                 .build();
     }
 
@@ -226,6 +228,7 @@ public class VectorService extends BaseService {
         return UpsertResp.builder()
                 .upsertCnt(response.getUpsertCnt())
                 .primaryKeys(ids)
+                .cost(getCost(response.getStatus()))
                 .build();
     }
 
@@ -328,7 +331,20 @@ public class VectorService extends BaseService {
         return QueryResp.builder()
                 .queryResults(convertUtils.getEntities(response))
                 .sessionTs(response.getSessionTs())
+                .cost(getCost(response.getStatus()))
                 .build();
+    }
+
+    private long getCost(Status status) {
+        String cost = status.getExtraInfoMap().get("report_value");
+        if (StringUtils.isNotEmpty(cost)) {
+            try {
+                return Long.parseLong(cost);
+            } catch (NumberFormatException e) {
+                logger.warn("Failed to parse report_value as cost: {}", cost);
+            }
+        }
+        return 0L;
     }
 
     public SearchResp search(MilvusServiceGrpc.MilvusServiceBlockingStub blockingStub, SearchReq request) {
@@ -376,7 +392,8 @@ public class VectorService extends BaseService {
         SearchResp.SearchRespBuilder respBuilder = SearchResp.builder()
                 .searchResults(convertUtils.getEntities(response))
                 .sessionTs(response.getSessionTs())
-                .recalls(response.getResults().getRecallsList());
+                .recalls(response.getResults().getRecallsList())
+                .cost(getCost(response.getStatus()));
         if (includeAggregations) {
             respBuilder.aggregationBuckets(convertUtils.getAggregationBuckets(response));
         }
@@ -590,9 +607,6 @@ public class VectorService extends BaseService {
     }
 
     private void fillSearchRespFromExtraInfo(SearchResp.SearchRespBuilder respBuilder, java.util.Map<String, String> extraInfo) {
-        if (extraInfo.containsKey("report_value")) {
-            respBuilder.cost(Long.parseLong(extraInfo.get("report_value")));
-        }
         if (extraInfo.containsKey("scanned_remote_bytes")) {
             respBuilder.scannedRemoteBytes(Long.parseLong(extraInfo.get("scanned_remote_bytes")));
         }
@@ -663,8 +677,10 @@ public class VectorService extends BaseService {
 
         // update the last write timestamp for SESSION consistency
         updateTsCache(dbName, collectionName, response.getTimestamp());
+
         return DeleteResp.builder()
                 .deleteCnt(response.getDeleteCnt())
+                .cost(getCost(response.getStatus()))
                 .build();
     }
 
@@ -712,8 +728,16 @@ public class VectorService extends BaseService {
                 .databaseName(request.getDatabaseName())
                 .collectionName(request.getCollectionName())
                 .ids(request.getIds());
+        List<String> partitionNames = new ArrayList<>();
         if (StringUtils.isNotEmpty(request.getPartitionName())) {
-            queryReqBuilder.partitionNames(Collections.singletonList(request.getPartitionName()));
+            partitionNames.add(request.getPartitionName());
+        }
+        if (request.getPartitionNames() != null) {
+            partitionNames.addAll(request.getPartitionNames());
+        }
+        if (!partitionNames.isEmpty()) {
+            // deduplicate while preserving order (a caller may set both partitionName and partitionNames)
+            queryReqBuilder.partitionNames(new ArrayList<>(new LinkedHashSet<>(partitionNames)));
         }
         QueryReq queryReq = queryReqBuilder.build();
         if (request.getOutputFields() != null) {
--- a/sdk-core/src/main/java/io/milvus/v2/utils/SchemaUtils.java
+++ b/sdk-core/src/main/java/io/milvus/v2/utils/SchemaUtils.java
@@ -251,6 +251,10 @@ public class SchemaUtils {
                 .defaultValue(ParamUtils.valueFieldToObject(fieldSchema.getDefaultValue(), fieldSchema.getDataType()))
                 .externalField(fieldSchema.getExternalField())
                 .build();
+        // server-assigned/derived attributes, populated only by describe_collection
+        schema.setFieldId(fieldSchema.getFieldID());
+        schema.setIsDynamic(fieldSchema.getIsDynamic());
+        schema.setIsFunctionOutput(fieldSchema.getIsFunctionOutput());
 
         Map<String, String> typeParams = new HashMap<>();
         for (KeyValuePair keyValuePair : fieldSchema.getTypeParamsList()) {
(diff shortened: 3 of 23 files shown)


sdk-core/src/main/java/io/milvus/v2/service/rbac/RBACService.java
```diff
@@ -131,6 +131,7 @@ public Void grantPrivilege(MilvusServiceGrpc.MilvusServiceBlockingStub blockingS
                         .setName(request.getRoleName())
                         .build())
                 .setObjectName(request.getObjectName())
+                .setDbName(request.getDbName())
```
yhmo: sdk-core/src/main/java/io/milvus/v2/service/rbac/RBACService.java line:134
High ---- GrantEntity.Builder.setDbName(null) throws NullPointerException (protobuf-generated string setters reject null), and GrantPrivilegeReq.getDbName() returns null by default since neither the field nor the builder initializes it. Any existing caller that does not call .dbName(...) now crashes at runtime — e.g. examples/src/main/java/io/milvus/v2/RBACExample.java builds GrantPrivilegeReq without dbName. The same applies to revokePrivilege at line 156 (RevokePrivilegeReq.dbName also defaults to null). pymilvus defaults db_name to "" in grant_privilege/revoke_privilege; please default dbName to "" or null-guard before setDbName so existing callers keep working.

sdk-core/src/main/java/io/milvus/v2/service/vector/VectorService.java
```diff
@@ -712,8 +736,15 @@ private QueryReq toQueryReq(GetReq request) {
                 .databaseName(request.getDatabaseName())
                 .collectionName(request.getCollectionName())
                 .ids(request.getIds());
+        List<String> partitionNames = new ArrayList<>();
         if (StringUtils.isNotEmpty(request.getPartitionName())) {
-            queryReqBuilder.partitionNames(Collections.singletonList(request.getPartitionName()));
+            partitionNames.add(request.getPartitionName());
+        }
+        if (request.getPartitionNames() != null) {
+            partitionNames.addAll(request.getPartitionNames());
```
yhmo, line 736: sdk-core/src/main/java/io/milvus/v2/service/vector/VectorService.java line:744
Low ---- partitionName and partitionNames are merged without deduplication, so a caller that sets partitionName("p1") plus partitionNames(Arrays.asList("p1", "p2")) sends ["p1", "p1", "p2"] in QueryRequest.partition_names. Consider deduplicating (or preferring partitionNames over partitionName when both are set) before building the request.

sdk-core/src/main/java/io/milvus/v2/service/vector/VectorService.java
```diff
@@ -331,6 +333,18 @@ private QueryResp convertQueryResponse(String title, QueryResults response) {
                 .build();
     }
 
+    private long getCost(Status status) {
```
yhmo, line 338: sdk-core/src/main/java/io/milvus/v2/service/vector/VectorService.java line:336
Low ---- For full pymilvus parity, query results still do not expose cost: QueryResp has no cost field while pymilvus's QueryResult surfaces the same status.extra_info["report_value"], and the existing search path reads it via fillSearchRespFromExtraInfo (which throws NumberFormatException on non-numeric values, unlike the defensive getCost added here). Consider routing query/search cost through this same defensive getCost helper and adding cost to QueryResp so insert/upsert/delete/search/query are consistent.

sdk-core/src/main/java/io/milvus/v2/utils/SchemaUtils.java
```diff
@@ -426,4 +438,22 @@ public static CreateCollectionReq.StructFieldSchema convertFieldReqToStructField
...
     }
 
+    /**
+     * Populates a private field on an already-built request model. Used for server-assigned /
+     * derived attributes (field ids, dynamic/function-output flags, function ids) that must be
+     * read-only to users but are filled in by {@link #convertFromGrpcFieldSchema} and
+     * {@link #convertFromGrpcFunction} when describing a collection. Reflection is required
+     * because these classes expose no setter/builder for such fields.
+     */
+    private static void setReadOnlyField(Object target, String fieldName, Object value) {
```
yhmo: sdk-core/src/main/java/io/milvus/v2/utils/SchemaUtils.java line:448
Low ---- Beyond the exception-handling note above, the reflection approach itself is brittle: the hardcoded field-name strings ("fieldId", "isDynamic", "isFunctionOutput", "id", "inputFieldIds", "outputFieldIds") must match private field names with no compile-time check, so a future rename silently breaks describeCollection with an unchecked RuntimeException. It also forces the builder-verification test to special-case FieldSchema/Function (MilvusClientV2Test.randomCallMethod) and to ignore the getter-only members. Consider package-private setters (or a small "populateServerAssigned" path) so the wiring is compile-checked and the reflection plus test workaround can be removed.

sdk-core/src/main/java/io/milvus/v2/utils/SchemaUtils.java
```diff
@@ -426,4 +438,22 @@ public static CreateCollectionReq.StructFieldSchema convertFieldReqToStructField
...
+        try {
+            Field field = target.getClass().getDeclaredField(fieldName);
+            field.setAccessible(true);
+            field.set(target, value);
+        } catch (ReflectiveOperationException e) {
+            throw new RuntimeException(
+                    "Failed to set read-only field '" + fieldName + "' on " + target.getClass().getSimpleName(), e);
+        }
```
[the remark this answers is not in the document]
yhmo: this issue is fixed, resolve the comment

sdk-core/src/main/java/io/milvus/v2/service/rbac/RBACService.java
yhmo: this issue is fixed, resolve the comment

sdk-core/src/main/java/io/milvus/v2/service/vector/VectorService.java
yhmo: this issue is fixed, resolve the comment

sdk-core/src/main/java/io/milvus/v2/service/vector/VectorService.java
yhmo: this issue is fixed, resolve the comment

sdk-core/src/main/java/io/milvus/v2/utils/SchemaUtils.java
yhmo: this issue is fixed, resolve the comment

sdk-core/src/main/java/io/milvus/v2/service/vector/VectorService.java
```diff
@@ -376,7 +392,8 @@ private SearchResp convertSearchResponse(String title, SearchResults response, b
         SearchResp.SearchRespBuilder respBuilder = SearchResp.builder()
                 .searchResults(convertUtils.getEntities(response))
                 .sessionTs(response.getSessionTs())
-                .recalls(response.getResults().getRecallsList());
+                .recalls(response.getResults().getRecallsList())
+                .cost(getCost(response.getStatus()));
```
yhmo, line 396: sdk-core/src/main/java/io/milvus/v2/service/vector/VectorService.java line:396
Low ---- getCost() always returns a value, so SearchResp.cost now becomes 0 when the server does not include "report_value", whereas previously it stayed null (cost was only set when report_value was present). Existing callers that use `getCost() != null` to detect whether the server reported a cost will now always observe 0 and cannot distinguish "cost not reported" from "cost 0". Note also the type inconsistency across the response family: SearchResp.cost is a boxed Long, while the new cost fields on InsertResp/UpsertResp/QueryResp/DeleteResp are primitive long. Consider aligning SearchResp.cost to primitive long (or documenting the 0 default) so all five responses share the same cost semantics and callers can rely on a consistent null/0 convention.

sdk-core/src/test/java/io/milvus/v2/client/MilvusClientV2Test.java
```diff
@@ -572,6 +577,12 @@ void testV2BuilderClasses() {
         VerifyClass(CreateCollectionReq.class.getName(), config);
         config.clearIgnoredMethods();
         VerifyClass(CreateCollectionReq.CollectionSchema.class.getName(), config);
+        // Function's server-assigned fields (id/inputFieldIds/outputFieldIds) have getters and
+        // setters but no builder methods, so ignore them in the builder-member check and the
+        // getter round-trip check. They can also be reached recursively from request classes
+        // that embed a Function member.
+        config.setIgnoredMethods(Arrays.asList("id", "inputFieldIds", "outputFieldIds",
```
yhmo, line 588: sdk-core/src/test/java/io/milvus/v2/client/MilvusClientV2Test.java line:584
Low ---- The ignoredMethods list set here for Function (id/inputFieldIds/outputFieldIds and their getters) is not scoped: it stays active through the subsequent VerifyClass calls (StructFieldSchema at line 587, DescribeCollectionReq at 588, ... ListCollectionsReq at 599) until it is replaced at line 600. None of those classes currently have an "id"/"inputFieldIds" member, so nothing is masked today, but if any of them later adds such a member the builder/getter round-trip checks would silently skip it. Consider calling config.clearIgnoredMethods() right after VerifyClass(CreateCollectionReq.Function...) to keep the ignore list scoped to Function.

sdk-core/src/main/java/io/milvus/v2/service/vector/VectorService.java
yhmo, line 396: sdk-core/src/main/java/io/milvus/v2/service/vector/VectorService.java line:396
Low ---- Behavior change for existing search callers: previously SearchResp.cost was set only when extra_info contained report_value (it stayed null otherwise); now getCost() always returns a value, defaulting to 0L when report_value is missing or non-numeric. A caller that checked getCost() != null to detect "no cost reported" now sees 0L. This also introduces an API inconsistency: SearchResp.cost remains boxed Long while the newly added cost fields on InsertResp/UpsertResp/DeleteResp/QueryResp are primitive long. If 0-as-default is intended (pymilvus also defaults cost to 0), consider documenting it; otherwise keeping the boxed nullable semantics for search would avoid the silent behavior change.

sdk-core/src/main/java/io/milvus/v2/service/vector/VectorService.java
```diff
@@ -663,8 +677,18 @@ public DeleteResp delete(MilvusServiceGrpc.MilvusServiceBlockingStub blockingStu
 
         // update the last write timestamp for SESSION consistency
         updateTsCache(dbName, collectionName, response.getTimestamp());
+
+        // pre-2.3.2 servers return the deleted primary keys; newer servers return only the count
+        List<Object> deletedPks = new ArrayList<>();
+        if (response.getIDs().hasIntId()) {
+            deletedPks = new ArrayList<>(response.getIDs().getIntId().getDataList());
+        } else if (response.getIDs().hasStrId()) {
```
yhmo: sdk-core/src/main/java/io/milvus/v2/service/vector/VectorService.java line:685
Low ---- For a String-pk collection on a pre-2.3.2 server, this branch returns the deleted primary keys as List<String>, which is correct. However, pymilvus aligns delete_primary_keys to the pk field type in the schema; the raw grpc str_id data here is used directly without conversion, so if the schema pk type were e.g. a VarChar field, the values are returned as-is which matches. This is mainly a note that the Int64 branch is unit-tested (testDeleteExposesCostAndPrimaryKeys) but the StrId branch is not; a small str-pk delete test would lock in the behavior.

sdk-core/src/main/java/io/milvus/v2/service/vector/VectorService.java
yhmo: The issue is confirmed fixed. Resolving this conversation.

yhmo: The issue is confirmed fixed. Resolving this conversation.

sdk-core/src/test/java/io/milvus/v2/client/MilvusClientV2Test.java
yhmo: The issue is confirmed fixed. Resolving this conversation.

sdk-core/src/test/java/io/milvus/v2/service/vector/VectorTest.java
```diff
@@ -606,6 +674,23 @@ void testGet() {
         logger.info(statusR.toString());
     }
 
+    @Test
+    void testGetWithMultiplePartitions() {
+        GetReq request = GetReq.builder()
+                .collectionName("book")
+                .ids(Collections.singletonList(1L))
+                .partitionName("p1")
+                .partitionNames(Arrays.asList("p2", "p3"))
```
yhmo, line 706: sdk-core/src/test/java/io/milvus/v2/service/vector/VectorTest.java line:683
Low ---- The new test uses partitionName("p1") + partitionNames(["p2","p3"]) — disjoint values — so the LinkedHashSet dedup branch added in VectorService.toQueryReq is never exercised. Consider adding a case with an overlapping value (e.g. partitionName("p1") + partitionNames(["p1","p2"]) → assert the request sends ["p1","p2"]) to cover the dedup, which is the whole point of the earlier fix.

sdk-core/src/main/java/io/milvus/v2/service/vector/VectorService.java
```diff
@@ -328,9 +331,22 @@ private QueryResp convertQueryResponse(String title, QueryResults response) {
         return QueryResp.builder()
                 .queryResults(convertUtils.getEntities(response))
                 .sessionTs(response.getSessionTs())
+                .cost(getCost(response.getStatus()))
                 .build();
     }
 
+    private long getCost(Status status) {
```
yhmo, line 338: sdk-core/src/main/java/io/milvus/v2/service/vector/VectorService.java line:338
Low ---- The two fallback branches of getCost() — report_value absent → 0L, and non-numeric report_value → warn + 0L — are not covered by tests. testInsertAndUpsertExposeServerCost / testDeleteExposesCost / testQueryExposesServerCost / the search cost path only exercise the numeric-present case. A small test asserting cost == 0 when report_value is missing, and cost == 0 when it is non-numeric, would lock in the defensive default and the intended pymilvus-aligned behavior.

sdk-core/src/main/java/io/milvus/v2/service/vector/response/SearchResp.java
```diff
@@ -74,11 +74,11 @@ public void setRecalls(List<Float> recalls) {
         this.recalls = recalls;
     }
 
-    public Long getCost() {
+    public long getCost() {
```
yhmo: sdk-core/src/main/java/io/milvus/v2/service/vector/response/SearchResp.java line:77
Low ---- Distinct from the null-vs-0 and Long/long consistency points already resolved above: switching cost from boxed Long to primitive long changes the JVM method descriptors (getCost()Ljava/lang/Long; -> ()J, setCost(Ljava/lang/Long;)V -> (J)V, and builder cost() likewise), so any caller compiled against the previous SDK jar hits NoSuchMethodError at runtime, and resp.setCost(null)/.cost(null) no longer compiles. No binary-compat check (japicmp/revapi) runs in CI, so this would surface only at runtime for downstream users. If primitive long is the intended pymilvus-aligned form, consider keeping the old boxed Long accessor as a deprecated shim, or at least call out the binary break in the release notes.

sdk-core/src/test/java/io/milvus/v2/client/MilvusClientV2Test.java
```diff
@@ -330,6 +330,11 @@ private static void randomCallMethod(Field field, Method method, Object callInst
                 // some method accepts Object as input such as AddFieldReq.defaultValue
                 Long obj = random.nextLong();
                 randomValues.put(field.getName(), obj);
+            } else if (fieldType == CreateCollectionReq.Function.class
+                    || fieldType == CreateCollectionReq.FieldSchema.class) {
+                // server-assigned read-only models with getter-only fields; constructed
+                // internally by describe_collection, so do not recurse into their builders
+                randomValues.put(field.getName(), null);
```
yhmo, line 337: sdk-core/src/test/java/io/milvus/v2/client/MilvusClientV2Test.java line:337
Low ---- Setting Function/FieldSchema members to null here means the embedded function of AddFunctionFieldReq (the only VerifyClass'd class with a direct Function member) is no longer constructed or round-trip checked: checkCallResult later calls getFunction() and evaluates temp.getClass() on the null result, throwing an NPE that the surrounding catch swallows — the assertion is skipped and a stack trace is printed on every run. Consider keeping the round-trip: guard checkCallResult against a null getter result (assertNull), or construct the embedded Function through its builder (id/inputFieldIds/outputFieldIds simply stay unset) while keeping the Function ignore list active for that recursion.

sdk-core/src/test/java/io/milvus/v2/service/vector/VectorTest.java
yhmo: this issue is fixed, resolve the comment

sdk-core/src/main/java/io/milvus/v2/service/vector/VectorService.java
yhmo: this issue is fixed, resolve the comment

sdk-core/src/main/java/io/milvus/v2/service/vector/response/SearchResp.java
yhmo: this issue is fixed, resolve the comment

sdk-core/src/test/java/io/milvus/v2/client/MilvusClientV2Test.java
yhmo: this issue is fixed, resolve the comment

PR #2037, opened 2026-08-21
Support FunctionChain

yhmo:


sdk-core/src/main/java/io/milvus/v2/service/vector/request/FunctionChain.java
```diff
@@ -0,0 +1,123 @@
...
+    private final String name;
+    private final List<FunctionChainOp> ops;
+
+    private FunctionChain(FunctionChainBuilder builder) {
+        this.stage = builder.stage;
+        this.name = builder.name;
+        this.ops = builder.ops;
+    }
+
+    public static FunctionChainBuilder builder() {
```
yhmo, line 60: sdk-core/src/main/java/io/milvus/v2/service/vector/request/FunctionChain.java line:57
Low ---- The built FunctionChain shares the builder's mutable ops list (this.ops = builder.ops; FunctionChainExpr.args/params and FunctionChainOp.inputs/outputs/params do the same). Because FunctionChainBuilder.map/sort/limit append to that same list, building a chain and then calling more builder methods retroactively changes the already-built chain, and getOps() returns the live mutable list. A defensive copy in build() (e.g. new ArrayList<>(builder.ops)) would keep the built object immutable as the fluent builder implies.

sdk-core/src/main/java/io/milvus/v2/service/vector/request/FunctionParamValue.java
```diff
@@ -0,0 +1,141 @@
...
+     * Converts an arbitrary Java object into a {@link FunctionParamValue}.
+     *
+     * <p>Supported types mirror PyMilvus's {@code _to_param_value}: {@code Boolean},
+     * {@code Integer}/{@code Long}/{@code Short}, {@code Float}/{@code Double}, {@code String},
+     * {@code byte[]}/{@link ByteString}, {@code List} (recursively), and {@code Map} (recursively).
+     */
+    public static FunctionParamValue from(Object value) {
+        if (value == null) {
+            throw new MilvusClientException(ErrorCode.INVALID_PARAMS, "Function chain parameters do not support null");
+        }
```
yhmo, line 98: sdk-core/src/main/java/io/milvus/v2/service/vector/request/FunctionParamValue.java line:94
Low ---- FunctionParamValue.from() carries most of the validation surface of this feature but has no direct unit tests: the supported-type branches (Boolean, integral types, Float/Double, String, byte[], ByteString, List, Map) and the three explicit error paths (null value, non-string/empty map key, unsupported type) are only exercised indirectly via the weights list, or not at all. The new tests cover the end-to-end chain serialization well; a small parameterized test for from() would lock in the type mapping and error behavior.

sdk-core/src/test/java/io/milvus/v2/service/vector/VectorTest.java
```diff
@@ -477,6 +482,77 @@ void testSearch() {
...
+        io.milvus.grpc.FunctionChain grpcChain = captor.getValue().getFunctionChains(0);
+        Assertions.assertEquals("rerank_chain", grpcChain.getName());
+        Assertions.assertEquals(io.milvus.grpc.FunctionChainStage.FunctionChainStageL2Rerank, grpcChain.getStage());
+        Assertions.assertEquals(3, grpcChain.getOpsCount());
+        Assertions.assertEquals("map", grpcChain.getOps(0).getOp());
+        Assertions.assertEquals("num_combine", grpcChain.getOps(0).getExpr().getName());
+        Assertions.assertEquals(2, grpcChain.getOps(0).getExpr().getArgsCount());
+        Assertions.assertEquals("weighted", grpcChain.getOps(0).getExpr().getParamsOrThrow("mode").getStringValue());
+        Assertions.assertEquals(3, grpcChain.getOps(0).getExpr().getParamsOrThrow("weights").getArrayValue().getValuesCount());
+        Assertions.assertEquals("sort", grpcChain.getOps(1).getOp());
```
yhmo: sdk-core/src/test/java/io/milvus/v2/service/vector/VectorTest.java line:520
Low ---- The happy-path test verifies the map expr (name/args/params), the limit param, and the three op names, but never asserts the sort op's wire shape (inputs ["$score","$id"] plus params column/desc/tie_break_col), the map op's outputs (["$score"]), the limit op's offset param, or a literal FunctionChainArg. Those are precisely the fields most likely to drift from the PyMilvus contract (sort.inputs and sort.params are the op with the most moving parts), and a serialization regression there would pass CI. Consider asserting inputs/params/outputs on ops 0-2 and adding one literal-arg case.

sdk-core/src/main/java/io/milvus/v2/service/vector/request/FunctionChainExpr.java
```diff
@@ -0,0 +1,106 @@
...
+
+        private FunctionChainExprBuilder() {
+        }
+
+        public FunctionChainExprBuilder name(String name) {
+            this.name = name;
+            return this;
+        }
+
+        public FunctionChainExprBuilder arg(FunctionChainArg arg) {
```
yhmo, line 84: sdk-core/src/main/java/io/milvus/v2/service/vector/request/FunctionChainExpr.java line:84
Low ---- `arg(FunctionChainArg arg)` accepts a null argument and defers the failure to an NPE inside `toGrpc()` (`arg.toGrpc()`), unlike the rest of this builder API which validates eagerly with a typed `MilvusClientException` (see `param`, `name`, and `FunctionChainArg.col`/`FunctionParamValue.from`). A null `arg` is caller-error input that should fail at `build()` time with a clear message; consider a null guard here (and on `FunctionChainOpBuilder.inputs/outputs`, which similarly accept lists whose elements are only dereferenced at serialization).

sdk-core/src/test/java/io/milvus/v2/service/vector/VectorTest.java
```diff
@@ -33,7 +33,12 @@
 import io.milvus.v2.common.IndexParam;
 import io.milvus.v2.exception.ErrorCode;
 import io.milvus.v2.exception.MilvusClientException;
+import io.milvus.v2.service.collection.request.CreateCollectionReq;
 import io.milvus.v2.service.vector.request.*;
+import io.milvus.v2.service.vector.request.FunctionChain;
+import io.milvus.v2.service.vector.request.FunctionChainArg;
+import io.milvus.v2.service.vector.request.FunctionChainExpr;
+import io.milvus.v2.service.vector.request.FunctionChainStage;
```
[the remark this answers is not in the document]
yhmo: The explicit imports are required, not redundant: io.milvus.v2.service.vector.request.* and io.milvus.grpc.* both declare FunctionChain/FunctionChainArg/FunctionChainExpr/FunctionChainStage, and the single-type imports are what disambiguate the two wildcard imports. Removing them makes the simple names ambiguous and fails compilation (verified). Kept as-is.

sdk-core/src/main/java/io/milvus/v2/utils/VectorUtils.java
```diff
@@ -420,6 +420,22 @@ public SearchRequest ConvertToGrpcSearchRequest(SearchReq request) {
             builder.setFunctionScore(convertOneFunction(ranker));
         }
 
+        // set function chains, mutually exclusive with ranker/functionScore
+        List<io.milvus.v2.service.vector.request.FunctionChain> functionChains = request.getFunctionChains();
+        if (functionChains != null && !functionChains.isEmpty()) {
+            if (ranker != null || functionScore != null) {
+                throw new MilvusClientException(ErrorCode.INVALID_PARAMS,
+                        "Not allow to set functionChains together with ranker/functionScore.");
```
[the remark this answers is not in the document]
yhmo: this issue is fixed, resolve this comment

sdk-core/src/test/java/io/milvus/v2/service/vector/VectorTest.java
```diff
@@ -477,6 +482,77 @@ void testSearch() {
...
+                .limit(10)
+                .functionChains(Collections.singletonList(FunctionChain.builder()
+                        .stage(FunctionChainStage.L2_RERANK)
+                        .build()))
+                .ranker(CreateCollectionReq.Function.builder()
+                        .name("rrf")
+                        .build())
+                .build();
+
+        Assertions.assertThrows(MilvusClientException.class, () -> client_v2.search(request));
```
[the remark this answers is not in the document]
yhmo: this issue is fixed, resolve this comment

sdk-core/src/main/java/io/milvus/v2/service/vector/request/FunctionChainOp.java
```diff
@@ -0,0 +1,180 @@
...
+            return this;
+        }
+
+        public FunctionChainOp build() {
+            if (op == null || op.isEmpty()) {
+                throw new MilvusClientException(ErrorCode.INVALID_PARAMS,
+                        "Function chain op name must be a non-empty string");
+            }
+            return new FunctionChainOp(this);
+        }
```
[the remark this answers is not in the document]
yhmo: The op is intentionally not restricted to map/sort/limit: schema.proto lists "map", "filter", "sort", "limit", and "merge" as valid operation names, so a hard allow-list would reject valid future server ops. The static map/sort/limit factories always set a valid op, and the raw builder still requires a non-empty string. The Javadoc now matches the proto instead of claiming only three ops.

sdk-core/src/main/java/io/milvus/v2/service/vector/request/FunctionChain.java
yhmo: this issue is fixed, resolve this comment

sdk-core/src/main/java/io/milvus/v2/service/vector/request/FunctionParamValue.java
yhmo: this issue is fixed, resolve this comment

sdk-core/src/test/java/io/milvus/v2/service/vector/VectorTest.java
yhmo: this issue is fixed, resolve this comment

sdk-core/src/main/java/io/milvus/v2/service/vector/request/FunctionChainExpr.java
yhmo: this issue is fixed, resolve this comment

sdk-core/src/main/java/io/milvus/v2/service/vector/request/FunctionChainOp.java
```diff
@@ -0,0 +1,191 @@
...
+            if (outputs == null) {
+                throw new MilvusClientException(ErrorCode.INVALID_PARAMS,
+                        "Function chain op outputs must not be null");
+            }
+            this.outputs.clear();
+            this.outputs.addAll(outputs);
+            return this;
+        }
+
+        public FunctionChainOpBuilder params(Map<String, FunctionParamValue> params) {
```
yhmo, line 177: sdk-core/src/main/java/io/milvus/v2/service/vector/request/FunctionChainOp.java line:177
Low ---- FunctionChainOpBuilder.params(Map) has no null guard: params(null) reaches this.params.putAll(null) and throws a raw NullPointerException, unlike the sibling inputs()/outputs() setters in this same class which reject null with a typed MilvusClientException(INVALID_PARAMS). Since FunctionChainOp.builder() is public, this inconsistent path is reachable by callers and produces a non-SDK-style failure. Consider the same null check as inputs()/outputs().

sdk-core/src/main/java/io/milvus/v2/service/vector/request/FunctionParamValue.java
```diff
@@ -0,0 +1,141 @@
...
+        if (value instanceof FunctionParamValue) {
+            return (FunctionParamValue) value;
+        }
+        if (value instanceof Boolean) {
+            return of((Boolean) value);
+        }
+        if (value instanceof Integer || value instanceof Long || value instanceof Short || value instanceof Byte) {
```
yhmo, line 105: sdk-core/src/main/java/io/milvus/v2/service/vector/request/FunctionParamValue.java line:101
Low ---- from() only accepts Integer/Long/Short/Byte, Float/Double, Boolean, String, byte[]/ByteString, List, and Map. Its javadoc claims parity with pymilvus _to_param_value, but common Java numeric/string types are rejected with INVALID_PARAMS where pymilvus accepts them: BigInteger (pymilvus int() handles arbitrary precision), Character (pymilvus treats a one-char string), and char[]. For SDK users passing a BigInteger or a char where the Python reference accepts the equivalent value, this is an avoidable parity gap. Consider mapping BigInteger (with range check) and Character/char[] to int64/string/bytes, or trimming the javadoc claim if intentional.

sdk-core/src/main/java/io/milvus/v2/service/vector/request/FunctionChainOp.java
yhmo, line 177: sdk-core/src/main/java/io/milvus/v2/service/vector/request/FunctionChainOp.java line:177
Low ---- The sibling setters `inputs()` and `outputs()` (lines 157-175) were given null guards that throw `MilvusClientException(INVALID_PARAMS)`, but `params()` still calls `this.params.putAll(params)` with no null check, so `.params(null)` surfaces a raw `NullPointerException` instead of the typed error. For consistency with the guards added on the lines above, add the same null check here.

sdk-core/src/main/java/io/milvus/v2/service/vector/request/FunctionChain.java
```diff
@@ -0,0 +1,123 @@
...
+
+    public static class FunctionChainBuilder {
+        private FunctionChainStage stage = FunctionChainStage.UNSPECIFIED;
+        private String name = "";
+        private final List<FunctionChainOp> ops = new ArrayList<>();
+
+        private FunctionChainBuilder() {
+        }
+
+        public FunctionChainBuilder stage(FunctionChainStage stage) {
```
yhmo, line 92: sdk-core/src/main/java/io/milvus/v2/service/vector/request/FunctionChain.java line:89
Low ---- `stage()` accepts `null` without a guard. `FunctionChainStage.UNSPECIFIED` is rejected by `VectorUtils.ConvertToGrpcSearchRequest`, but an explicit `null` stage slips past that check (`null == UNSPECIFIED` is false) and then throws an `NPE` inside `stage.toGrpc()` at serialization time, instead of a clear `MilvusClientException`. Since the sibling builders in this PR now reject null args/inputs/outputs, a null guard here (or a check in `build()`) would keep the failure eager and typed.

sdk-core/src/main/java/io/milvus/v2/service/vector/request/FunctionChainOp.java
```diff
@@ -0,0 +1,191 @@
...
+        if (expr != null) {
+            builder.setExpr(expr.toGrpc());
+        }
+        builder.addAllInputs(inputs);
+        builder.addAllOutputs(outputs);
+        params.forEach((k, v) -> builder.putParams(k, v.toGrpc()));
+        return builder.build();
+    }
+
+    static FunctionChainOp map(String output, FunctionChainExpr expr) {
```
yhmo, line 93: sdk-core/src/main/java/io/milvus/v2/service/vector/request/FunctionChainOp.java line:93
Low ---- The builder validation error branches are not directly covered by unit tests: `map()` with an empty `output` or null `expr`, `sort()` with an empty `by`, and `limit()` with `limit <= 0` or `offset < 0`. The new `testFunctionParamValueFrom` covers `FunctionParamValue.from()` well, and the happy-path test asserts the wire shape, but the factory validation paths are only exercised indirectly (or not at all). A small parameterized test for these error branches would raise the patch coverage on this file (codecov currently reports ~72% here and 34-44% on `FunctionChainArg`/`FunctionParamValue`) and lock in the intended error behavior.

sdk-core/src/main/java/io/milvus/v2/service/vector/request/FunctionChainOp.java
yhmo: this issue is fixed, resolve this comment

sdk-core/src/main/java/io/milvus/v2/service/vector/request/FunctionParamValue.java
yhmo: this issue is fixed, resolve this comment

sdk-core/src/main/java/io/milvus/v2/service/vector/request/FunctionChainOp.java
yhmo: this issue is fixed, resolve this comment

sdk-core/src/main/java/io/milvus/v2/service/vector/request/FunctionChain.java
yhmo: this issue is fixed, resolve this comment

sdk-core/src/main/java/io/milvus/v2/service/vector/request/FunctionChainOp.java
yhmo: this issue is fixed, resolve this comment

sdk-core/src/main/java/io/milvus/v2/service/vector/request/FunctionParamValue.java
```diff
@@ -0,0 +1,157 @@
...
+import io.milvus.v2.exception.MilvusClientException;
+
+import java.util.List;
+import java.util.Map;
+
+/**
+ * A typed value used as a function-chain parameter or literal argument.
+ *
+ * <p>Values are converted from ordinary Java objects into the protobuf
+ * {@code FunctionParamValue} message: {@code Boolean} -&gt; bool, integral numbers -&gt; int64,
```
yhmo: sdk-core/src/main/java/io/milvus/v2/service/vector/request/FunctionParamValue.java line:33
Low ---- The class-level Javadoc still lists the supported conversions as Boolean/integral/Float/Double/String/byte[]+ByteString/List/Map, but the latest commit added BigInteger (range-checked to int64), Character (single-char string) and char[] (UTF-8 bytes) to `from()` without updating this summary. `from()`'s own Javadoc (lines 84-92) documents the new types, so the class summary and the method Javadoc now disagree about the supported input surface. Please extend the class Javadoc to match, so the public contract is consistent.

sdk-core/src/test/java/io/milvus/v2/service/vector/VectorTest.java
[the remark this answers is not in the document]
yhmo: no need fix

sdk-core/src/main/java/io/milvus/v2/utils/VectorUtils.java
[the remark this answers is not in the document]
yhmo: no need fix

sdk-core/src/main/java/io/milvus/v2/service/vector/request/FunctionChainOp.java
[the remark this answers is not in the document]
yhmo: no need fix

sdk-core/src/test/java/io/milvus/v2/service/vector/VectorTest.java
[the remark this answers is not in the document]
yhmo: already fix

sdk-core/src/main/java/io/milvus/v2/service/vector/request/FunctionChain.java
yhmo: already fix

sdk-core/src/main/java/io/milvus/v2/service/vector/request/FunctionParamValue.java
yhmo: already fix

sdk-core/src/test/java/io/milvus/v2/service/vector/VectorTest.java
yhmo: already fix

sdk-core/src/main/java/io/milvus/v2/service/vector/request/FunctionChainExpr.java
yhmo: already fix

sdk-core/src/main/java/io/milvus/v2/service/vector/request/FunctionChainOp.java
yhmo: already fix

sdk-core/src/main/java/io/milvus/v2/service/vector/request/FunctionParamValue.java
yhmo: already fix

sdk-core/src/main/java/io/milvus/v2/service/vector/request/FunctionChainOp.java
yhmo: already fix

sdk-core/src/main/java/io/milvus/v2/service/vector/request/FunctionChain.java
yhmo: already fix

sdk-core/src/main/java/io/milvus/v2/service/vector/request/FunctionChainOp.java
yhmo: already fix

sdk-core/src/main/java/io/milvus/v2/service/vector/request/FunctionParamValue.java
yhmo: fixed

sdk-core/src/main/java/io/milvus/v2/utils/VectorUtils.java
```diff
@@ -420,6 +420,22 @@ public SearchRequest ConvertToGrpcSearchRequest(SearchReq request) {
...
 
+        // set function chains, mutually exclusive with ranker/functionScore
+        List<io.milvus.v2.service.vector.request.FunctionChain> functionChains = request.getFunctionChains();
+        if (functionChains != null && !functionChains.isEmpty()) {
+            if (ranker != null || functionScore != null) {
+                throw new MilvusClientException(ErrorCode.INVALID_PARAMS,
+                        "Not allowed to set functionChains together with ranker/functionScore.");
+            }
+            for (io.milvus.v2.service.vector.request.FunctionChain chain : functionChains) {
+                if (chain.getStage() == io.milvus.v2.service.vector.request.FunctionChainStage.UNSPECIFIED) {
```
yhmo, line 431: sdk-core/src/main/java/io/milvus/v2/utils/VectorUtils.java line:431
Medium ---- A null element inside `functionChains` (e.g. `SearchReq.builder().functionChains(Arrays.asList(chain, null))` or `addFunctionChain(null)`) hits `chain.getStage()` here and throws a raw NPE instead of the typed `MilvusClientException(INVALID_PARAMS)` used for every other null input in this feature. In the sync search path the NPE propagates to the caller untyped. Suggest rejecting null elements here (or guarding `addFunctionChain`/`functionChains`) with the same INVALID_PARAMS error so misuse fails cleanly in both sync and async search.

sdk-core/src/main/java/io/milvus/v2/service/vector/request/FunctionChainOp.java
```diff
@@ -0,0 +1,195 @@
...
+                        "Function chain op params must not be null");
+            }
+            this.params.clear();
+            this.params.putAll(params);
+            return this;
+        }
+
+        public FunctionChainOp build() {
+            if (op == null || op.isEmpty()) {
+                throw new MilvusClientException(ErrorCode.INVALID_PARAMS,
```
yhmo, line 189: sdk-core/src/main/java/io/milvus/v2/service/vector/request/FunctionChainOp.java line:189
Low ---- The new tests cover the `map`/`sort`/`limit` factory error paths (`testFunctionChainFactoryValidation`) and `stage(null)`/`params(null)` (`testFunctionChainBuilderNullGuards`), but these raw-builder branches remain untested: `build()` with an empty/null `op` here, `FunctionChainExprBuilder.build()` with an empty `name`, `arg(null)`, and `FunctionChainArg.col("")`. These branches implement the typed INVALID_PARAMS contract and could silently regress without any test; a few assertions would close the gap.

sdk-core/src/main/java/io/milvus/v2/utils/VectorUtils.java
yhmo: fixed

sdk-core/src/main/java/io/milvus/v2/service/vector/request/FunctionChainOp.java
yhmo: fixed

PR #2040, opened 2026-08-24
feat: add client telemetry support

xiaofan-luan:
## Summary

- add a client telemetry manager, configuration object, and gRPC interceptor
- report one outcome per logical operation across legacy/V1, V2, async, retry, validation, and result processing paths
- preserve telemetry state across global-cluster reconnects with failure-safe candidate handoff and stale-heartbeat fencing
- close shared telemetry with fluent wrappers, preserve arbitrary legacy wire request IDs while keeping telemetry trace IDs strict, and omit an unconfigured database
- keep equal-timestamp command IDs until the timestamp cursor advances so repeated delivery stays idempotent
- keep the command heartbeat alive after a server-side dynamic disable and carry that activation across runtime-state handoff, while an initial user opt-out starts no worker
- align strict atomic commands, deterministic sampling, collection filtering, RFC3339 validation, acknowledgement, and unsupported backoff with the Go SDK
- use a 10-second default heartbeat interval and add focused unit plus opt-in end-to-end coverage

## Related work

- Design: https://github.com/milvus-io/milvus/blob/master/docs/design-docs/design_docs/20260131-client_side_telemetry.md
- Server and Go SDK correctness follow-up: https://github.com/milvus-io/milvus/pull/52597
- Go SDK conformance follow-up: https://github.com/milvus-io/milvus/pull/52804

## Verification

- full sdk-core suite before the final control-plane follow-up: 633 tests, 0 failures, 0 errors, 1 skipped
- latest telemetry manager and GlobalStub handoff suites: 29 passed
- local diff coverage: 92.63%
- git diff --check

### DIFF
--- a/sdk-core/src/main/java/io/milvus/client/MilvusServiceClient.java
+++ b/sdk-core/src/main/java/io/milvus/client/MilvusServiceClient.java
@@ -91,7 +91,7 @@ public class MilvusServiceClient extends AbstractMilvusGrpcClient {
     private final ClientTelemetryManager telemetry;
     private final boolean ownsTelemetry;
     private final ThreadLocal<String> clientRequestId;
-    private static final ThreadLocal<Integer> LOGICAL_OPERATION_DEPTH =
+    private final ThreadLocal<Integer> logicalOperationDepth =
             ThreadLocal.withInitial(() -> 0);
 
     public MilvusServiceClient(ConnectParam connectParam) {
@@ -480,12 +480,12 @@ public class MilvusServiceClient extends AbstractMilvusGrpcClient {
 
     private <T> R<T> recordLogicalOperation(
             String operation, String collection, java.util.function.Supplier<R<T>> supplier) {
-        int depth = LOGICAL_OPERATION_DEPTH.get();
-        LOGICAL_OPERATION_DEPTH.set(depth + 1);
+        int depth = logicalOperationDepth.get();
+        logicalOperationDepth.set(depth + 1);
         long startNanos = System.nanoTime();
         String requestId = captureClientRequestId();
-        try (TelemetryInterceptor.LogicalOperationScope ignored =
-                     TelemetryInterceptor.beginLogicalOperation()) {
+        try (ClientTelemetryManager.LogicalOperationScope ignored =
+                     telemetry.beginLogicalOperation()) {
             R<T> result = supplier.get();
             if (depth == 0) {
                 recordLogicalResult(operation, collection, startNanos, resultError(result), requestId);
@@ -504,13 +504,13 @@ public class MilvusServiceClient extends AbstractMilvusGrpcClient {
     private <T> ListenableFuture<R<T>> recordLogicalOperationAsync(
             String operation, String collection,
             java.util.function.Supplier<ListenableFuture<R<T>>> supplier) {
-        int depth = LOGICAL_OPERATION_DEPTH.get();
-        LOGICAL_OPERATION_DEPTH.set(depth + 1);
+        int depth = logicalOperationDepth.get();
+        logicalOperationDepth.set(depth + 1);
         long startNanos = System.nanoTime();
         String requestId = captureClientRequestId();
         final ListenableFuture<R<T>> future;
-        try (TelemetryInterceptor.LogicalOperationScope ignored =
-                     TelemetryInterceptor.beginLogicalOperation()) {
+        try (ClientTelemetryManager.LogicalOperationScope ignored =
+                     telemetry.beginLogicalOperation()) {
             future = supplier.get();
         } catch (RuntimeException | Error throwable) {
             if (depth == 0) {
@@ -536,11 +536,11 @@ public class MilvusServiceClient extends AbstractMilvusGrpcClient {
         return future;
     }
 
-    private static void resetLogicalOperationDepth(int depth) {
+    private void resetLogicalOperationDepth(int depth) {
         if (depth == 0) {
-            LOGICAL_OPERATION_DEPTH.remove();
+            logicalOperationDepth.remove();
         } else {
-            LOGICAL_OPERATION_DEPTH.set(depth);
+            logicalOperationDepth.set(depth);
         }
     }
 
--- a/sdk-core/src/main/java/io/milvus/client/MilvusServiceClient.java
+++ b/sdk-core/src/main/java/io/milvus/client/MilvusServiceClient.java
@@ -102,7 +102,7 @@ public class MilvusServiceClient extends AbstractMilvusGrpcClient {
                 connectParam.getTelemetryConfig(),
                 connectParam.getUserName(),
                 getSDKVersion(),
-                () -> currentDatabaseName == null ? "default" : currentDatabaseName,
+                () -> currentDatabaseName,
                 () -> telemetryUserConfig(connectParam));
         this.ownsTelemetry = true;
         this.clientRequestId = connectParam.getClientRequestId();
@@ -243,7 +243,9 @@ public class MilvusServiceClient extends AbstractMilvusGrpcClient {
         this.currentDatabaseName = src.currentDatabaseName;
         this.endpoint = src.endpoint;
         this.telemetry = src.telemetry;
-        this.ownsTelemetry = false;
+        // Fluent wrappers share the channel and therefore share its lifecycle. Closing any
+        // wrapper closes that channel, so it must close the shared telemetry worker as well.
+        this.ownsTelemetry = true;
         this.clientRequestId = src.clientRequestId;
     }
 
--- a/sdk-core/src/main/java/io/milvus/client/MilvusServiceClient.java
+++ b/sdk-core/src/main/java/io/milvus/client/MilvusServiceClient.java
@@ -56,6 +56,8 @@ import io.milvus.param.index.*;
 import io.milvus.param.partition.*;
 import io.milvus.param.resourcegroup.*;
 import io.milvus.param.role.*;
+import io.milvus.telemetry.ClientTelemetryManager;
+import io.milvus.telemetry.TelemetryInterceptor;
 import io.milvus.v2.utils.ClientUtils;
 import org.apache.commons.lang3.StringUtils;
 
@@ -65,7 +67,9 @@ import java.net.InetAddress;
 import java.net.UnknownHostException;
 import java.time.LocalDateTime;
 import java.util.ArrayList;
+import java.util.LinkedHashMap;
 import java.util.List;
+import java.util.Map;
 import java.util.concurrent.Callable;
 import java.util.concurrent.TimeUnit;
 
@@ -79,11 +83,20 @@ public class MilvusServiceClient extends AbstractMilvusGrpcClient {
     private RetryParam retryParam = RetryParam.newBuilder().build();
     private String currentDatabaseName;
     private final String endpoint;
+    private final ClientTelemetryManager telemetry;
+    private final boolean ownsTelemetry;
 
     public MilvusServiceClient(ConnectParam connectParam) {
         ExceptionUtils.checkNotNull(connectParam, connectParam.getClass().getSimpleName());
         this.rpcDeadlineMs = connectParam.getRpcDeadlineMs();
         this.endpoint = connectParam.getHost() + ":" + connectParam.getPort();
+        this.telemetry = new ClientTelemetryManager(
+                connectParam.getTelemetryConfig(),
+                connectParam.getUserName(),
+                getSDKVersion(),
+                () -> currentDatabaseName == null ? "default" : currentDatabaseName,
+                () -> telemetryUserConfig(connectParam));
+        this.ownsTelemetry = true;
 
         Metadata metadata = new Metadata();
         metadata.put(Metadata.Key.of("authorization", Metadata.ASCII_STRING_MARSHALLER), connectParam.getAuthorization());
@@ -110,6 +123,7 @@ public class MilvusServiceClient extends AbstractMilvusGrpcClient {
                 };
             }
         });
+        clientInterceptors.add(new TelemetryInterceptor(telemetry, connectParam.getClientRequestId()));
 
         try {
             if (StringUtils.isNotEmpty(connectParam.getServerPemPath())) {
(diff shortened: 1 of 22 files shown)


sdk-core/src/main/java/io/milvus/common/interceptor/ClientRequestInterceptor.java
```diff
@@ -56,11 +56,27 @@ public void start(Listener<RespT> responseListener, Metadata headers) {
...
+                if (isValidClientRequestId(requestId)) {
                     headers.put(CLIENT_REQUEST_ID_HEADER, requestId);
                 }
                 super.start(responseListener, headers);
             }
         };
     }
+
+    /** Mirrors the server's OpenTelemetry TraceID parser. */
+    public static boolean isValidClientRequestId(String requestId) {
```
yhmo, line 71: sdk-core/src/main/java/io/milvus/common/interceptor/ClientRequestInterceptor.java line:68
Medium ---- This validator silently drops any client_request_id that is not exactly 32 lowercase hex digits, which changes the documented contract (ConnectConfig says callers "can assign a specific id" for access-log correlation) and loses the header with no warning, exception, or migration path. It is also stricter than the server: milvus parses the header with OTel TraceIDFromHex, which is case-insensitive and accepts e.g. "4BF92F3577B34DA6A3CE929D0E0E4736", while this check rejects it -- note the existing VectorTest cases had to be rewritten from arbitrary ids like "query-request" to 32-hex strings to keep passing. Consider at least logging when an id is dropped, or normalizing valid hex (lowercasing) instead of rejecting it, so existing callers are not silently cut off from correlation.

sdk-core/src/main/java/io/milvus/client/MilvusServiceClient.java
```diff
@@ -79,11 +88,24 @@ public class MilvusServiceClient extends AbstractMilvusGrpcClient {
...
 
     public MilvusServiceClient(ConnectParam connectParam) {
         ExceptionUtils.checkNotNull(connectParam, connectParam.getClass().getSimpleName());
         this.rpcDeadlineMs = connectParam.getRpcDeadlineMs();
         this.endpoint = connectParam.getHost() + ":" + connectParam.getPort();
+        this.telemetry = new ClientTelemetryManager(
+                connectParam.getTelemetryConfig(),
+                connectParam.getUserName(),
+                getSDKVersion(),
+                () -> currentDatabaseName == null ? "default" : currentDatabaseName,
```
yhmo: sdk-core/src/main/java/io/milvus/client/MilvusServiceClient.java line:105
Low ---- This database provider always returns a non-empty name ("default" when none is configured), and MilvusClientV2.currentUsedDatabase() does the same, so the telemetry heartbeat always sends Reserved["db_name"]="default". The Go SDK omits db_name entirely when no database is configured, and the design explicitly warns against synthesizing a "default" database for clients that never selected one (serverless). Because the server matches database:<name> command scoping on this field, a database:default push would reach Java clients that never used any database but not Go clients in the same state. Is always reporting "default" intentional, or should the field be omitted when no database is configured?

sdk-core/src/main/java/io/milvus/telemetry/ClientTelemetryManager.java
```diff
@@ -0,0 +1,1462 @@
...
+            return state;
+        }
+    }
+
+    /**
+     * Transfers the final runtime state to a connected replacement without ever running two
+     * telemetry workers for the same client ID. Operations that finish after the handoff are
+     * forwarded to the replacement instead of being dropped by the closed manager.
+     */
+    public void handoffRuntimeStateTo(ClientTelemetryManager replacement) {
```
yhmo: sdk-core/src/main/java/io/milvus/telemetry/ClientTelemetryManager.java line:357
Low ---- This one-argument handoffRuntimeStateTo overload appears to be unused: no production caller or test invokes it (GlobalStub and the tests use the retirement-token overload). If it is intentional public API for future callers, that is fine; otherwise it is a second state-handoff path that can drift from the fenced, token-versioned one and should be removed.

sdk-core/src/main/java/io/milvus/client/MilvusServiceClient.java
```diff
@@ -232,6 +242,9 @@ protected MilvusServiceClient(MilvusServiceClient src) {
         this.retryParam = src.retryParam;
         this.currentDatabaseName = src.currentDatabaseName;
         this.endpoint = src.endpoint;
+        this.telemetry = src.telemetry;
+        this.ownsTelemetry = false;
```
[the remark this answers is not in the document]
yhmo: The issue is confirmed fixed. Resolving this conversation.

sdk-core/src/main/java/io/milvus/common/interceptor/ClientRequestInterceptor.java
yhmo: The issue is confirmed fixed. Resolving this conversation.

sdk-core/src/main/java/io/milvus/client/MilvusServiceClient.java
yhmo: The issue is confirmed fixed. Resolving this conversation.

sdk-core/src/main/java/io/milvus/telemetry/ClientTelemetryManager.java
yhmo: The issue is confirmed fixed. Resolving this conversation.

sdk-core/src/main/java/io/milvus/v2/client/MilvusClientV2.java
```diff
@@ -286,14 +322,70 @@ private MilvusServiceGrpc.MilvusServiceFutureStub getFutureRpcStub() {
...
+                TelemetryInterceptor.LOGICAL_OPERATION_OPTION, true);
     }
 
     private String captureClientRequestId() {
         if (connectConfig == null || connectConfig.getClientRequestId() == null) {
-            return null;
+            return "";
+        }
+        String requestId = connectConfig.getClientRequestId().get();
+        return ClientRequestInterceptor.isValidClientRequestId(requestId) ? requestId : "";
```
yhmo, line 339: sdk-core/src/main/java/io/milvus/v2/client/MilvusClientV2.java line:334
Medium ---- captureClientRequestId() now applies the strict OTel trace-ID check (exactly 32 lowercase hex, non-zero) before the value is set as CLIENT_REQUEST_ID_OPTION on the future stub, so on queryAsync/searchAsync/hybridSearchAsync/getAsync an arbitrary caller-supplied client_request_id becomes "" and is dropped from the wire metadata, while sync V2 and legacy V1 still transmit it. Before this PR the raw thread-local value reached the wire on the async path too, so callers using ConnectConfig.clientRequestId with a non-OTel correlation ID silently lose access-log correlation on async calls. The strict check is already applied exactly where telemetry needs it (TelemetryInterceptor.requestId), so the raw value can be kept in the option and filtered only when recording error details. This is a follow-up to the resolved thread on ClientRequestInterceptor.java:68 -- that fix restored wire preservation at the interceptor level, but the async path never reaches the interceptor with the raw id, so the drop persists for async V2.

sdk-core/src/main/java/io/milvus/v2/client/MilvusClientV2.java
```diff
@@ -286,14 +322,70 @@ private MilvusServiceGrpc.MilvusServiceFutureStub getFutureRpcStub() {
...
+        ClientTelemetryManager manager = getTelemetry();
+        if (manager == null) {
+            return supplier.get();
+        }
+        long startNanos = System.nanoTime();
+        String requestId = captureClientRequestId();
+        try (TelemetryInterceptor.LogicalOperationScope ignored =
+                     TelemetryInterceptor.beginLogicalOperation()) {
+            T result = supplier.get();
+            manager.recordOperation(operation, collection, startNanos, "", requestId);
```
yhmo: sdk-core/src/main/java/io/milvus/v2/client/MilvusClientV2.java line:348
Medium ---- The happy-path `manager.recordOperation(...)` here is not wrapped in a best-effort guard, unlike the V1 legacy path: `MilvusServiceClient.recordLogicalResult` calls `telemetry.recordOperation` inside `try { ... } catch (RuntimeException ignored) { // Telemetry is best-effort and must never replace the operation result. }`. Here, if `recordOperation` throws on the success path, control jumps to the `catch (RuntimeException | Error)` below, which calls `recordOperation` again (error path) and then rethrows — so a telemetry-side failure would turn an otherwise successful Milvus operation into an exception for the caller. Consider wrapping the success-path recording in the same try/catch the V1 path uses so the telemetry contract is identical across both client surfaces.

sdk-core/src/main/java/io/milvus/telemetry/TelemetryInterceptor.java
```diff
@@ -0,0 +1,199 @@
...
+import java.util.HashSet;
+import java.util.Set;
+
+/** Records operation telemetry at the common unary gRPC boundary. */
+public final class TelemetryInterceptor implements ClientInterceptor {
+    private static final Set<String> OPERATIONS = new HashSet<>(Arrays.asList(
+            "Insert", "Delete", "Upsert", "Search", "HybridSearch", "Query", "RunAnalyzer"));
+    public static final CallOptions.Key<Boolean> LOGICAL_OPERATION_OPTION =
+            CallOptions.Key.create("milvus-telemetry-logical-operation");
+    private static final ThreadLocal<Integer> LOGICAL_OPERATION_DEPTH =
```
yhmo: sdk-core/src/main/java/io/milvus/telemetry/TelemetryInterceptor.java line:45
Low ---- `LOGICAL_OPERATION_DEPTH` is a static ThreadLocal, and `MilvusServiceClient` keeps a second, separate static ThreadLocal of the same name (MilvusServiceClient.java line:94). Because both are process-global per thread, any two clients running interleaved or nested logical operations on the same thread share the same depth counter (e.g., a user callback issued inside client A's operation that triggers client B's search). The scopes are balanced by try-with-resources, so normal nesting is fine, but the shared static state means one client's scope can suppress or mis-record telemetry for an unrelated client's operation on the same thread. Consider instance-scoping the depth counters to the owning client.

sdk-core/src/main/java/io/milvus/v2/client/MilvusClientV2.java
```diff
@@ -948,11 +1086,13 @@ CompletableFuture<QueryResp> queryAsync(QueryReq request, String clusterId) {
...
      */
     public SearchResp search(SearchReq request) {
-        return rpcUtils.retry(() -> vectorService.search(this.getRpcStub(), request));
+        return search(request, null);
     }
 
     SearchResp search(SearchReq request, String clusterId) {
-        return rpcUtils.retry(() -> vectorService.search(this.getRpcStub(), request, clusterId));
+        String collection = request == null ? "" : request.getCollectionName();
+        return recordLogicalOperation("Search", collection,
```
yhmo, line 1250: sdk-core/src/main/java/io/milvus/v2/client/MilvusClientV2.java line:1094
Medium ---- Iterators are the one V2 DQL path that is not wrapped in a logical-operation scope: `searchIteratorV2`/`queryIteratorV2` build their stub via `createIteratorRpcStub` and issue `search`/`query` RPCs per page-fetch without any `recordLogicalOperation` wrapper. Those RPCs still traverse the `TelemetryInterceptor`, so each page becomes its own per-attempt Search/Query metric. The PR summary states the goal is "report one outcome per logical operation across legacy/V1, V2, async, retry, validation, and result processing paths"; an iterator over N pages will instead report N Search/Query entries. Please confirm whether per-page recording for iterators is intended, or whether iterator construction should be wrapped so one logical iteration produces one metric.

sdk-core/src/main/java/io/milvus/telemetry/ClientTelemetryManager.java
```diff
@@ -0,0 +1,1451 @@
...
+
+        OperationSnapshot(String operation, MetricSnapshot global, Map<String, MetricSnapshot> collections) {
+            this.operation = operation;
+            this.global = global;
+            this.collection_metrics = collections;
+            this.collections = collections;
+        }
+    }
+
+    public static final class MetricSnapshot {
```
yhmo, line 1476: sdk-core/src/main/java/io/milvus/telemetry/ClientTelemetryManager.java line:1317
Low ---- These three snapshot value classes each store the same value twice: public snake_case fields (request_count, end_time, collection_metrics) and private camelCase copies (requestCount, endTime, collections), kept in sync by hand in the constructors. Any future metric field has to be added and assigned in two places, and a one-sided edit would silently diverge the JSON replies (show_errors / show_latency_history build their payloads from the camelCase copies) from what external callers of getMetricsSnapshots()/snapshotRuntimeState() read via the snake_case fields, with no test catching it unless the duplicate is asserted. Consider a single field set with a Gson field-naming strategy (@SerializedName or FieldNamingPolicy) so there is one source of truth to keep in sync.

sdk-core/src/main/java/io/milvus/v2/client/MilvusClientV2.java
yhmo: The issue is confirmed fixed. Resolving this conversation.

sdk-core/src/main/java/io/milvus/v2/client/MilvusClientV2.java
yhmo: The issue is confirmed fixed. Resolving this conversation.

sdk-core/src/main/java/io/milvus/telemetry/TelemetryInterceptor.java
yhmo: The issue is confirmed fixed. Resolving this conversation.

sdk-core/src/main/java/io/milvus/v2/client/MilvusClientV2.java
yhmo: The issue is confirmed fixed. Resolving this conversation.

sdk-core/src/main/java/io/milvus/telemetry/ClientTelemetryManager.java
yhmo: The issue is confirmed fixed. Resolving this conversation.

PR #2042, opened 2026-08-24
Support element-level query

yhmo:


sdk-core/src/main/java/io/milvus/v2/service/vector/response/QueryResp.java
```diff
@@ -116,21 +122,51 @@ public void setEntity(Map<String, Object> entity) {
...
+
+        public void setOriginalIdx(Long originalIdx) {
+            this.originalIdx = originalIdx;
+        }
+
         @Override
         public String toString() {
             return "QueryResult{" +
                     "entity=" + entity +
+                    ", elementOffset=" + elementOffset +
```
yhmo: sdk-core/src/main/java/io/milvus/v2/service/vector/response/QueryResp.java line:145
Low ---- For every normal (non-element-level) query these two fields are null, so toString() now renders "elementOffset=null, originalIdx=null" on every result, adding noise to existing query-result logging. SearchResp.SearchResult.toString() (SearchResp.java:264) already guards elementOffset with a null check; consider the same conditional rendering here so only element-level results print the extra fields.

sdk-core/src/main/java/io/milvus/v2/utils/ConvertUtils.java
```diff
@@ -85,12 +85,32 @@ public List<QueryResp.QueryResult> getEntities(QueryResults response) {
...
-            entities.add(queryResult);
-        });
+        List<QueryResultsWrapper.RowRecord> rowRecords = queryResultsWrapper.getRowRecords();
+        List<io.milvus.grpc.ElementIndices> elementIndices = response.getElementIndicesList();
+        if (elementIndices != null && !elementIndices.isEmpty()) {
+            // struct-array element-level query: expand one entity into one row per matched element
+            for (int i = 0; i < rowRecords.size(); i++) {
+                Map<String, Object> fieldValues = rowRecords.get(i).getFieldValues();
+                if (i >= elementIndices.size()) {
+                    continue;
```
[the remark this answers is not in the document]
yhmo: fixed

sdk-core/src/main/java/io/milvus/v2/utils/ConvertUtils.java
```diff
@@ -85,12 +85,32 @@ public List<QueryResp.QueryResult> getEntities(QueryResults response) {
...
+                }
+                for (Long offset : elementIndices.get(i).getIndices().getDataList()) {
+                    QueryResp.QueryResult queryResult = QueryResp.QueryResult.builder()
+                            .entity(fieldValues)
```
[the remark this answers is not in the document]
yhmo: fixed

sdk-core/src/main/java/io/milvus/v2/service/vector/response/QueryResp.java
yhmo: fixed

sdk-core/src/main/java/io/milvus/v2/utils/ConvertUtils.java
```diff
@@ -85,12 +85,34 @@ public List<QueryResp.QueryResult> getEntities(QueryResults response) {
...
+        List<QueryResultsWrapper.RowRecord> rowRecords = queryResultsWrapper.getRowRecords();
+        List<io.milvus.grpc.ElementIndices> elementIndices = response.getElementIndicesList();
+        if (elementIndices != null && !elementIndices.isEmpty()) {
+            if (elementIndices.size() != rowRecords.size()) {
+                throw new MilvusClientException(ErrorCode.SERVER_ERROR, String.format(
+                        "The element_indices count (%d) does not match the row count (%d)",
+                        elementIndices.size(), rowRecords.size()));
+            }
+            // struct-array element-level query: expand one entity into one row per matched element
+            for (int i = 0; i < rowRecords.size(); i++) {
```
yhmo, line 99: sdk-core/src/main/java/io/milvus/v2/utils/ConvertUtils.java line:97
Medium ---- This expansion duplicates the element-level logic already implemented in QueryIterator.extractRowRecords() (sdk-core/src/main/java/io/milvus/orm/iterator/QueryIterator.java), but the two paths expose the result differently: queryIterator() embeds the matched offset as an "offset" key inside the entity map (Constant.OFFSET), while this query() path adds elementOffset/originalIdx as separate QueryResult accessors (pymilvus likewise exposes offset/_original_idx as row keys). The same server feature now surfaces with different client-visible shapes depending on which V2 API is used, so the two implementations can drift. Consider extracting the shared expansion and/or documenting the divergence on QueryResult so callers of query() know the entity map intentionally omits the offset key.

sdk-core/src/test/java/io/milvus/v2/service/vector/VectorTest.java
```diff
@@ -249,6 +249,47 @@ void testQueryCostDefaultsToZeroWhenReportValueMissingOrInvalid() {
         Assertions.assertEquals(0L, respInvalid.getCost());
     }
 
+    @Test
+    void testQueryElementLevelExpandsResultsWithOffset() {
```
yhmo, line 253: sdk-core/src/test/java/io/milvus/v2/service/vector/VectorTest.java line:253
Low ---- The new error path introduced here (element_indices count != row count -> MilvusClientException) has no test; only the happy expansion path is covered. pymilvus covers this exact case with test_query_element_indices_length_mismatch. Consider adding a test that asserts the exception is thrown on a mismatch so the guard stays exercised.

sdk-core/src/main/java/io/milvus/v2/utils/ConvertUtils.java
yhmo: fixed

sdk-core/src/test/java/io/milvus/v2/service/vector/VectorTest.java
yhmo: fixed

sdk-core/src/main/java/io/milvus/v2/service/vector/response/QueryResp.java
```diff
@@ -100,8 +100,23 @@ public QueryResp build() {
     public static class QueryResult {
         private Map<String, Object> entity;
 
+        /**
+         * For struct-array element-level queries (via {@code element_filter}): the matched
+         * element's index within the array, and the original entity index before row expansion.
```
yhmo: sdk-core/src/main/java/io/milvus/v2/service/vector/response/QueryResp.java line:105
Low ---- The Javadoc describes originalIdx as "the original entity index before row expansion", but the value is the source row's position within the current response page (the loop index i), not a stable entity index. For a paginated element-level query (QueryReq.offset + limit), originalIdx restarts at 0 on each page, so it cannot identify the source entity across pages; the primary key in the entity map is the reliable identifier. Consider clarifying in the Javadoc that originalIdx is the page-local row position.

yhmo: no need

PR #2045, opened 2026-08-25
Support query iterator cursor

yhmo:


sdk-core/src/main/java/io/milvus/orm/iterator/QueryIteratorCursor.java
```diff
@@ -0,0 +1,142 @@
...
+        private Long lastElementOffset;
+
+        public QueryIteratorCursorBuilder sessionTs(long sessionTs) {
+            this.sessionTs = sessionTs;
+            return this;
+        }
+
+        public QueryIteratorCursorBuilder intPk(Long intPk) {
+            this.intPk = intPk;
+            return this;
```
yhmo, line 137: sdk-core/src/main/java/io/milvus/orm/iterator/QueryIteratorCursor.java line:125
Medium ---- toProto()/fromProto() drop lastElementOffset, so an element-filter cursor that is serialized and then deserialized resumes without the element offset: hasElementCursor() becomes false, the resumed iterator uses "pk >" (strict) instead of "pk >=" + QUERY_ITER_LAST_ELEMENT_OFFSET, and the remaining elements under the last matched pk are silently skipped. This contradicts the class javadoc ("the cursor holds ... the last matched element offset, so pagination can continue in a new iterator") and is a data-loss surface in the new public serialization API (in-memory resume through QueryIteratorReq.cursor(...) is unaffected). pymilvus has the same proto limitation, but its cp-file resume path preserves the offset in JSON, so this Java path is new. Consider rejecting toProto() when lastElementOffset is set, documenting the limitation, or extending QueryCursor with a last_element_offset field.

sdk-core/src/main/java/io/milvus/orm/iterator/QueryIterator.java
```diff
@@ -189,6 +195,34 @@ public void close() {
...
+            builder.strPk(String.valueOf(nextId));
+        } else {
+            builder.intPk(((Number) nextId).longValue());
+        }
+        return builder.build();
+    }
+
+    private void restoreFromCursor(QueryIteratorCursor cursor) {
+        this.sessionTs = cursor.getSessionTs();
+        this.nextId = cursor.getStrPk() != null ? cursor.getStrPk() : cursor.getIntPk();
```
yhmo, line 221: sdk-core/src/main/java/io/milvus/orm/iterator/QueryIterator.java line:221
Medium ---- The QueryIteratorReq.cursor field comment promises "offset is ignored in that case", but restoreFromCursor() never zeroes this.offset: the constructor reads it from the request before the cursor branch, seek() is skipped, and next() passes offset as the RPC offset on every resumed page (QueryIterator.java:180). The normal path consumes the request offset once inside seek() and then sets offset = 0; the resume path re-applies a non-zero request offset on each page, silently skipping rows at and after the resume boundary. Zeroing offset in restoreFromCursor() (or in the cursor branch) would make the code honor the documented contract.

sdk-core/src/main/java/io/milvus/v2/service/vector/request/QueryIteratorReq.java
```diff
@@ -172,6 +179,14 @@ public Map<String, Object> getFilterTemplateValues() {
         return filterTemplateValues;
     }
 
+    public QueryIteratorCursor getCursor() {
+        return cursor;
```
yhmo, line 183: sdk-core/src/main/java/io/milvus/v2/service/vector/request/QueryIteratorReq.java line:183
Low ---- The new cursor field is omitted from toString(), so a dumped request will not reveal whether a resume cursor is attached, unlike every other field. Consider appending ", cursor=" + cursor to help debugging resume flows.

sdk-core/src/main/java/io/milvus/orm/iterator/QueryIteratorCursor.java
yhmo: The issue is confirmed fixed. Resolving this conversation.

sdk-core/src/main/java/io/milvus/orm/iterator/QueryIterator.java
yhmo: The issue is confirmed fixed. Resolving this conversation.

sdk-core/src/main/java/io/milvus/v2/service/vector/request/QueryIteratorReq.java
yhmo: The issue is confirmed fixed. Resolving this conversation.

examples/src/main/java/io/milvus/v2/IteratorExample.java
```diff
@@ -199,6 +200,61 @@ private static void queryIteratorWithTemplate(int batchSize) {
...
+        System.out.printf("%d query results returned before cursor capture, cursor=%s%n",
+                firstPage.size(), cursor);
+
+        // resume pagination from the captured cursor in a brand new iterator
+        QueryIterator resumed = client.queryIterator(QueryIteratorReq.builder()
+                .collectionName(COLLECTION_NAME)
+                .expr(expr)
+                .outputFields(Lists.newArrayList(ID_FIELD, AGE_FIELD))
+                .batchSize(batchSize)
+                .limit(limit)
```
yhmo: examples/src/main/java/io/milvus/v2/IteratorExample.java line:235
Low ---- The resumed iterator re-applies the original .limit(limit) (100), so when the collection holds more than limit matching rows the combined output becomes firstPage.size() + limit (up to 102) instead of the advertised 100 - the cursor carries no remaining-limit state. Consider resuming with limit - firstPage.size(), or add a comment noting the caller must account for rows already consumed.

sdk-core/src/main/java/io/milvus/orm/iterator/QueryIterator.java
```diff
@@ -189,6 +195,35 @@ public void close() {
...
```
yhmo, line 221: sdk-core/src/main/java/io/milvus/orm/iterator/QueryIterator.java line:221
Low ---- The varchar primary-key resume path is only covered up to capture (queryIteratorGetCursorVarcharPk); an actual resume never runs through this strPk branch together with the quoted pk > "..." expression in setupNextExpr(). A regression there would go unnoticed - consider a queryIteratorResumesVarcharPkFromCursor test asserting the resumed expr and request.

examples/src/main/java/io/milvus/v2/IteratorExample.java
yhmo: The example now resumes with `limit - firstPage.size()`, so the combined total stays at `limit`; a comment documents the per-iterator limit semantics. Fixed.

sdk-core/src/main/java/io/milvus/orm/iterator/QueryIterator.java
yhmo: Added `queryIteratorResumesVarcharPkFromCursor`, which resumes from a strPk cursor and asserts the quoted `pk > \"...\"` expression plus the reused guarantee timestamp. Fixed.

PR #2047, opened 2026-08-26
Fix three PyMilvus parity gaps

yhmo:
- Expose deleted primary keys in DeleteResp
  - Preserve field index params in DescribeCollectionResp
  - Support compact target size units
  - Add a V2 function chain example

### DIFF
--- a/sdk-core/src/main/java/io/milvus/v2/service/utility/request/CompactReq.java
+++ b/sdk-core/src/main/java/io/milvus/v2/service/utility/request/CompactReq.java
@@ -19,12 +19,17 @@
 
 package io.milvus.v2.service.utility.request;
 
+import io.milvus.v2.exception.ErrorCode;
+import io.milvus.v2.exception.MilvusClientException;
+import io.milvus.v2.service.utility.OptimizeTask;
+
 public class CompactReq {
     private String databaseName;
     private String collectionName;
     private Boolean isClustering = Boolean.FALSE;
     private Boolean isL0 = Boolean.FALSE;
-    private Long targetSize; // in MB, null means server default
+    private Long targetSize; // value in targetSizeUnit, null means server default
+    private String targetSizeUnit = "mb";
 
     private CompactReq(CompactReqBuilder builder) {
         this.databaseName = builder.databaseName;
@@ -32,6 +37,7 @@ public class CompactReq {
         this.isClustering = builder.isClustering;
         this.isL0 = builder.isL0;
         this.targetSize = builder.targetSize;
+        this.targetSizeUnit = builder.targetSizeUnit;
     }
 
     public static CompactReqBuilder builder() {
@@ -78,6 +84,18 @@ public class CompactReq {
         this.targetSize = targetSize;
     }
 
+    public String getTargetSizeUnit() {
+        return targetSizeUnit;
+    }
+
+    public void setTargetSizeUnit(String targetSizeUnit) {
+        this.targetSizeUnit = targetSizeUnit;
+    }
+
+    public Long getTargetSizeInMB() {
+        return convertTargetSizeToMB(targetSize, targetSizeUnit);
+    }
+
     @Override
     public String toString() {
         return "CompactReq{" +
@@ -86,7 +104,28 @@ public class CompactReq {
                 ", isClustering=" + isClustering +
                 ", isL0=" + isL0 +
                 ", targetSize=" + targetSize +
+                ", targetSizeUnit='" + targetSizeUnit + '\'' +
                 '}';
+    }
+
+    private static Long convertTargetSizeToMB(Long targetSize, String targetSizeUnit) {
+        // null and 0 both mean "server default": the server reads an unset target_size as 0
+        // and skips force-merge, so mapping 0 to null preserves the wire behavior that existed
+        // before this change (0 was previously sent verbatim and treated as server default).
+        if (targetSize == null || targetSize == 0) {
+            return null;
+        }
+        if (targetSize < 0) {
+            throw new MilvusClientException(ErrorCode.INVALID_PARAMS,
+                    "targetSize must be a positive integer: " + targetSize);
+        }
+        if (targetSizeUnit == null) {
+            throw new MilvusClientException(ErrorCode.INVALID_PARAMS,
+                    "targetSizeUnit cannot be null when targetSize is set");
+        }
+        // Reuse the shared size-to-MB parser from OptimizeTask so the unit table and the
+        // 1MB floor live in a single place instead of two sibling parsers that can drift.
+        return OptimizeTask.parseTargetSize(targetSize + targetSizeUnit);
     }
 
     public static class CompactReqBuilder {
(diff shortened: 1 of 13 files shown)


sdk-core/src/main/java/io/milvus/v2/service/vector/VectorService.java
```diff
@@ -678,8 +678,16 @@ public DeleteResp delete(MilvusServiceGrpc.MilvusServiceBlockingStub blockingStu
         // update the last write timestamp for SESSION consistency
         updateTsCache(dbName, collectionName, response.getTimestamp());
 
+        List<Object> primaryKeys = new ArrayList<>();
+        if (response.getIDs().hasIntId()) {
```
yhmo: sdk-core/src/main/java/io/milvus/v2/service/vector/VectorService.java line:682
Medium ---- Modern Milvus servers (>= 2.3.2) no longer return the deleted primary keys in the delete response: the proxy initializes `MutationResult.IDs` to an empty `IDs` and never fills it (the echo was removed in "support complex delete expression" for 2.3.2, see `internal/proxy/task_delete.go`), which is also why PyMilvus's `delete()` docstring notes that keys are only returned by servers older than 2.3.2.
As a result `DeleteResp.getPrimaryKeys()` will always be an empty list against a current server, and the new `VectorTest.testDeleteExposesCost` asserts against a mocked response the real server never produces. Consider documenting this legacy-only behavior on `DeleteResp`/`delete()`, and guarding the read with `response.hasIDs()` since `getIDs()` can be null when a server returns success without setting the field.

sdk-core/src/main/java/io/milvus/v2/utils/SchemaUtils.java
```diff
@@ -289,6 +289,22 @@ public static CreateCollectionReq.FieldSchema convertFromGrpcFieldSchema(FieldSc
             typeParams.put(keyValuePair.getKey(), keyValuePair.getValue());
         }
         schema.setTypeParams(typeParams);
+
+        Map<String, Object> indexParams = new HashMap<>();
+        for (KeyValuePair keyValuePair : fieldSchema.getIndexParamsList()) {
+            if (io.milvus.param.Constant.PARAMS.equals(keyValuePair.getKey())) {
+                indexParams.put(keyValuePair.getKey(), JsonUtils.fromJson(keyValuePair.getValue(),
```
yhmo: sdk-core/src/main/java/io/milvus/v2/utils/SchemaUtils.java line:296
Medium ---- This unguarded `JsonUtils.fromJson(...)` can throw `JsonSyntaxException` if the server returns an index `params` value that is not a valid JSON object, and it runs on the describeCollection path whose output feeds the schema cache used by search/query/insert. The typeParams block directly above wraps the same kind of parse in a try/catch because "the kernel does not enforce validation on the input typeParams". Consider guarding this parse the same way (log and skip the key on failure) so a malformed `params` value cannot break `describeCollection`.

sdk-core/src/main/java/io/milvus/v2/service/utility/request/CompactReq.java
```diff
@@ -86,15 +112,61 @@ public String toString() {
...
                 ", targetSize=" + targetSize +
+                ", targetSizeUnit='" + targetSizeUnit + '\'' +
                 '}';
     }
 
+    private static Long convertTargetSizeToMB(Long targetSize, String targetSizeUnit) {
+        if (targetSize == null) {
+            return null;
+        }
+        if (targetSize <= 0) {
```
yhmo: sdk-core/src/main/java/io/milvus/v2/service/utility/request/CompactReq.java line:123
Low ---- Backward-compat tightening: before this PR a `targetSize` of 0 was sent verbatim and the server treated it as "use the default/auto target size" (manualCompaction skips force-merge when `target_size==0`), so existing callers passing `0L` now get `INVALID_PARAMS` instead. The only way to express "server default" is now `null`. If any caller may pass 0, either handle it as "server default" or call the change out as intentional.

sdk-core/src/main/java/io/milvus/v2/service/vector/VectorService.java
yhmo: The issue is confirmed fixed. Resolving this conversation.

sdk-core/src/main/java/io/milvus/v2/utils/SchemaUtils.java
yhmo: The issue is confirmed fixed. Resolving this conversation.

sdk-core/src/main/java/io/milvus/v2/service/utility/request/CompactReq.java
yhmo: The issue is confirmed fixed. Resolving this conversation.

sdk-core/src/main/java/io/milvus/v2/service/utility/request/CompactReq.java
```diff
@@ -86,15 +112,63 @@ public String toString() {
                 ", isClustering=" + isClustering +
                 ", isL0=" + isL0 +
                 ", targetSize=" + targetSize +
+                ", targetSizeUnit='" + targetSizeUnit + '\'' +
                 '}';
     }
 
+    private static Long convertTargetSizeToMB(Long targetSize, String targetSizeUnit) {
```
yhmo, line 111: sdk-core/src/main/java/io/milvus/v2/service/utility/request/CompactReq.java line:119
Medium ---- This adds a second size-unit to MB parser in the same io.milvus.v2.service.utility package: OptimizeTask.parseTargetSize(String) (OptimizeTask.java:290) already converts B/KB/MB/GB/TB/PB to MB, and the new convertTargetSizeToMB(Long, String) re-implements the same unit table and 1MB floor with different semantics (OptimizeTask accepts decimals like "1.5gb" and treats a missing unit as bytes; CompactReq is integer-only and defaults to MB). Two sibling parsers for the same mapping will drift on the next unit change (adding a unit, decimal support, or rounding rule). Consider extracting one shared parser and reusing it from both paths, e.g. CompactReq could call OptimizeTask.parseTargetSize(targetSize + unit).

sdk-core/src/main/java/io/milvus/v2/service/utility/request/CompactReq.java
```diff
@@ -86,15 +112,63 @@ public String toString() {
...
+        // null and 0 both mean "server default": the server reads an unset target_size as 0
+        // and skips force-merge, so mapping 0 to null keeps the pre-2.x wire behavior intact.
```
yhmo: sdk-core/src/main/java/io/milvus/v2/service/utility/request/CompactReq.java line:121
Low ---- The wording "keeps the pre-2.x wire behavior intact" is ambiguous: nothing in this change is tied to a Milvus 2.x version, so a future reader may read it as behavior specific to servers before 2.x. The intent is "preserves the wire behavior that existed before this change" (0 remains 0 on the wire through the proto3 default). Consider rewording to avoid the version-specific phrasing.

sdk-core/src/main/java/io/milvus/v2/service/utility/request/CompactReq.java
yhmo: The issue is confirmed fixed. Resolving this conversation.

sdk-core/src/main/java/io/milvus/v2/service/utility/request/CompactReq.java
yhmo: The issue is confirmed fixed. Resolving this conversation.

PR #2055, opened 2026-08-31
Support nullable struct field

yhmo:


sdk-core/src/main/java/io/milvus/v2/service/collection/request/CreateCollectionReq.java
```diff
@@ -2072,6 +2072,10 @@ public StructFieldSchema addField(AddFieldReq addFieldReq) {
             if (addFieldReq.getDataType() == DataType.Array || addFieldReq.getElementType() == DataType.Struct) {
                 throw new ParamException("Struct field schema does not support Array, ArrayOfVector or Struct");
             }
+            if (Boolean.TRUE.equals(addFieldReq.getIsNullable())) {
```
yhmo: sdk-core/src/main/java/io/milvus/v2/service/collection/request/CreateCollectionReq.java line:2075
Medium ---- This new rejection is a public API behavior change: existing callers that previously added a struct sub-field with isNullable=true via addField() now get a ParamException at collection-creation time. Also note the guard only covers the addField() path — sub-fields added through StructFieldSchemaBuilder.fields(...) or getFields().add(...) still accept isNullable=true and are silently overridden to the struct nullable in SchemaUtils.convertToGrpcStructFieldSchema. Consider documenting this restriction in the addField Javadoc, and either enforcing it consistently on all construction paths or explicitly documenting the silent override, so the two paths cannot drift.

sdk-core/src/main/java/io/milvus/response/FieldDataWrapper.java
```diff
@@ -199,6 +199,12 @@ public long getRowCount() throws IllegalResponseException {
                 return fieldData.getScalars().getArrayData().getDataCount();
             case ArrayOfStruct: {
                 List<FieldData> structData = fieldData.getStructArrays().getFieldsList();
+                for (FieldData fd : structData) {
+                    List<Boolean> subValid = fd.getValidDataList();
+                    if (subValid != null && !subValid.isEmpty()) {
+                        return subValid.size();
```
yhmo, line 205: sdk-core/src/main/java/io/milvus/response/FieldDataWrapper.java line:205
Low ---- getRowCount() for ArrayOfStruct returns the validData size of the FIRST sub-field that has non-empty validData, while getStructData() computes rowCount as the max over all sub-fields (validData size, array dataCount, vector dataCount). If sub-fields ever carry inconsistent lengths, the two methods disagree on the total row count, which can desync callers that use getRowCount() to size the decoded data. Consider computing the row count the same way in both methods to keep them consistent.

sdk-core/src/main/java/io/milvus/v2/utils/DataUtils.java
```diff
@@ -547,9 +562,24 @@ private static FieldData genStructSubFieldData(CreateCollectionReq.FieldSchema f
...
+            List<?> nonNullRows = objects;
+            if (isNullable) {
+                List<Object> filtered = new ArrayList<>();
+                for (Object obj : objects) {
+                    builder.addValidData(obj != null);
+                    if (obj != null) {
+                        filtered.add(obj);
+                    }
+                }
+                if (filtered.isEmpty()) {
```
yhmo, line 583: sdk-core/src/main/java/io/milvus/v2/utils/DataUtils.java line:577
Low ---- When every row of a nullable vector sub-field is null, this early return builds an ArrayOfVector FieldData with validData (all false) but no vectors and no setDim, whereas the non-null path always sets VectorField.setDim(schema dimension). The all-null scalar path (line 610) has a unit test (testInsertNullableStructAllNullRowsLeavePayloadUnset), but there is no equivalent test for an all-null vector sub-field, and the DockerTest only mixes null and non-null rows. Please add a unit/integration case for the all-null vector sub-field and confirm the server accepts a dimension-less ArrayOfVector payload.

sdk-core/src/main/java/io/milvus/v2/utils/DataUtils.java
```diff
@@ -458,12 +458,18 @@ private boolean processStructFieldValues(JsonObject row, CreateCollectionReq.Str
                                                  Map<String, InsertDataInfo> nameInsertInfo, boolean partialUpdate) {
             String structName = structField.getName();
             JsonElement rowFieldData = row.get(structName);
-            if (rowFieldData == null) {
+            if (rowFieldData == null || rowFieldData.isJsonNull()) {
```
yhmo: sdk-core/src/main/java/io/milvus/v2/utils/DataUtils.java line:461
Low ---- In a partial update, an explicit JsonNull for a nullable struct field is silently skipped via `return false`, so callers cannot null out a nullable struct during partial update. This differs from nullable scalar fields, which accept JsonNull and write a null value (processNormalFieldValues -> checkRowFieldValue). Please confirm skipping is intended; if not, route JsonNull through addNullStructRow so the struct can be nulled in partial update like a scalar field.

sdk-core/src/main/java/io/milvus/v2/service/collection/request/AddCollectionStructFieldReq.java
```diff
@@ -202,12 +200,11 @@ public void setTypeParams(Map<String, String> typeParams) {
...
      *
      * @return the struct field schema
-     * @throws MilvusClientException if the field is not nullable or the schema conversion fails
+     * @throws ParamException if the field is not nullable or the schema conversion fails
      */
     public CreateCollectionReq.StructFieldSchema toStructFieldSchema() {
         if (Boolean.FALSE.equals(nullable)) {
-            throw new MilvusClientException(ErrorCode.INVALID_PARAMS,
-                    "Adding struct field to existing collection requires nullable=true");
+            throw new ParamException("Adding struct field to existing collection requires nullable=true");
```
yhmo, line 207: sdk-core/src/main/java/io/milvus/v2/service/collection/request/AddCollectionStructFieldReq.java line:207
Low ---- `toStructFieldSchema()` is a public method on a public request class, and this PR changes its thrown exception type from `io.milvus.v2.exception.MilvusClientException` (which carries `getErrorCode()`) to `io.milvus.exception.ParamException`. The RPC path is preserved because `CollectionService.addCollectionStructField` re-wraps `ParamException` into `MilvusClientException(INVALID_PARAMS)`, so end users calling `addCollectionStructField` still see the same exception. However, any direct caller that catches the v2 `MilvusClientException` around `toStructFieldSchema()` will no longer catch these validation failures. Please confirm this public contract change is intentional (e.g., only internal callers use the method), or keep the v2 exception for direct callers so the request class remains source-compatible.

sdk-core/src/test/java/io/milvus/v2/client/MilvusClientV2DockerTest.java
```diff
@@ -1371,6 +1373,12 @@ void testStruct() {
         Assertions.assertEquals(DataType.Float16Vector, structSchema.getFields().get(3).getDataType());
         Assertions.assertEquals(DataType.BFloat16Vector, structSchema.getFields().get(4).getDataType());
         Assertions.assertEquals(DataType.Int8Vector, structSchema.getFields().get(5).getDataType());
+        // nullable struct field: nullable must be true on the struct and inherited by every sub-field
+        Assertions.assertTrue(structSchema.getNullable());
+        for (CreateCollectionReq.FieldSchema subField : structSchema.getFields()) {
+            Assertions.assertTrue(subField.getIsNullable(), "sub-field " + subField.getName() + " should inherit nullable");
+        }
+        Assertions.assertFalse(descSchema.getStructFields().get(1).getNullable());
```
yhmo: sdk-core/src/test/java/io/milvus/v2/client/MilvusClientV2DockerTest.java line:1381
Low ---- This assertion assumes the describe response preserves struct-field creation order (index 1 == st2). If the server ever returns struct fields in a different order, the assertion would silently test the wrong field. Consider locating st2 by name (e.g. via getStructField("st2")) so the check is order-independent and still validates that a non-nullable struct does not inherit nullable.

sdk-core/src/test/java/io/milvus/docker-compose.yml
```diff
@@ -52,8 +52,8 @@ services:
       timeout: 20s
       retries: 3
     ports:
-      - "19530:19530"
-      - "9091:9091"
+      - "29530:19530"
```
yhmo, line 55: sdk-core/src/test/java/io/milvus/docker-compose.yml line:55
Low ---- The docker-compose/TestUtils port renumbering (standalone 19530->29530, etcd 2379->12379, minio 9000->19000, health 9091/9092->19091/19092, and the matching cache keys in the DockerTests) is unrelated to nullable-struct support and is bundled into this feature PR. If the renumbering was needed to avoid conflicts with a locally running Milvus, consider splitting it into a separate test-infra PR so the feature change stays reviewable; also confirm every docker test sharing these compose files (e.g. MilvusClientV2SessionDockerTest, NullableVectorTest) is updated consistently, since they rely on the same MilvusStandaloneUri and compose file.

sdk-core/src/main/java/io/milvus/v2/service/collection/request/AddCollectionStructFieldReq.java
yhmo: This is intentional. The change aligns the validation exception with pymilvus, which throws ParamError for struct sub-field validation; to keep the end-user RPC contract unchanged, CollectionService.addCollectionStructField re-wraps ParamException into MilvusClientException(INVALID_PARAMS), so client.addCollectionStructField() still surfaces the same error code. Direct callers of toStructFieldSchema() are only the internal CollectionService and tests (no external usage), so dropping the v2 exception here does not break public API callers, and it makes the request layer consistent with createCollection, which validates struct sub-fields via SchemaUtils.convertFieldReqToStructFieldSchema throwing ParamException.

sdk-core/src/test/java/io/milvus/v2/client/MilvusClientV2DockerTest.java
yhmo: Fixed. The assertion now locates st2 by name via descSchema.getStructField("st2") instead of positional index 1, so the non-nullable check is independent of describe ordering. Verified by running MilvusClientV2DockerTest#testStruct (BUILD SUCCESS).

sdk-core/src/test/java/io/milvus/docker-compose.yml
yhmo: The port renumbering (standalone 29530, etcd 12379, minio 19000/19001, health 19091/19092, plus matching cache keys in MilvusClientV2DockerTest/MilvusClientDockerTest and MultiConnectParam ports in MilvusMultiClientDockerTest) is intentional to avoid clashing with a locally running Milvus, and is bundled here as part of the same nullable-struct test infrastructure work. Consistency verified: all five DockerTest classes sharing these compose files connect via TestUtils.MilvusStandaloneUri (now 29530) and no stale 19530 references remain; only the two DockerTests using CollectionTsCache keys had explicit endpoint strings, and those were updated.

PR #2056, opened 2026-09-01
Align V2 API parity with PyMilvus 2.6

yhmo:


sdk-core/src/main/java/io/milvus/v2/service/collection/CollectionService.java
```diff
@@ -463,6 +463,9 @@ public Void loadCollection(MilvusServiceGrpc.MilvusServiceBlockingStub blockingS
         if (StringUtils.isNotEmpty(dbName)) {
             builder.setDbName(dbName);
         }
+        if (StringUtils.isNotEmpty(request.getPriority())) {
+            builder.putLoadParams("load_priority", request.getPriority());
```
yhmo: sdk-core/src/main/java/io/milvus/v2/service/collection/CollectionService.java line:467
Low ---- `load_priority` is passed through verbatim, but the proxy only maps the exact lowercase string "low" to LOW priority (task.go: `if ok && loadPriorityStr == "low"`); every other value, including capitalized "Low"/"High", becomes HIGH. The new tests (CollectionTest/PartitionTest) use capitalized "High"/"Low", so a user copying that pattern would silently get HIGH priority on a real server. Consider documenting the accepted values (lowercase "low"/"high") or normalizing the case before `putLoadParams`, and apply the same to PartitionService.java loadPartitions.

sdk-core/src/main/java/io/milvus/v2/service/utility/UtilityService.java
```diff
@@ -215,6 +215,46 @@ public CompactResp compact(MilvusServiceGrpc.MilvusServiceBlockingStub blockingS
...
+            case "pb":
+                unitBytes = 1024L * 1024L * 1024L * 1024L * 1024L;
+                break;
+            default:
+                throw new MilvusClientException(ErrorCode.INVALID_PARAMS,
+                        "Invalid targetSizeUnit: '" + request.getTargetSizeUnit()
+                                + "'. Supported units: b, kb, mb, gb, tb, pb");
+        }
+        return (targetSize * unitBytes) / (1024L * 1024L);
+    }
```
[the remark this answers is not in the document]
yhmo: Fixed in 52f504ef. Conversion now uses BigDecimal for overflow-safe arithmetic, and values that convert to <= 0 MB (e.g. 1kb) are rejected with INVALID_PARAMS before being sent to the server.

sdk-core/src/main/java/io/milvus/v2/service/utility/UtilityService.java
```diff
@@ -215,6 +215,46 @@ public CompactResp compact(MilvusServiceGrpc.MilvusServiceBlockingStub blockingS
...
+     * it is sent to the server, and rejects non-positive values. Supported units: b, kb, mb, gb, tb, pb.
+     * This mirrors the PyMilvus {@code parse_target_size} behavior.
+     */
+    private long convertTargetSizeToMB(CompactReq request) {
+        Long targetSize = request.getTargetSize();
+        if (targetSize <= 0) {
+            throw new MilvusClientException(ErrorCode.INVALID_PARAMS,
+                    "targetSize must be a positive integer, got " + targetSize);
+        }
+        String unit = request.getTargetSizeUnit() == null ? "mb" : request.getTargetSizeUnit().trim().toLowerCase();
```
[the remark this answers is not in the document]
yhmo: Fixed in 52f504ef. Blank/whitespace-only targetSizeUnit is now treated as unset and defaults to MB, no longer throwing INVALID_PARAMS.

sdk-core/src/main/java/io/milvus/v2/service/collection/CollectionService.java
yhmo: fixed

sdk-core/src/main/java/io/milvus/v2/service/utility/UtilityService.java
```diff
@@ -215,6 +217,59 @@ public CompactResp compact(MilvusServiceGrpc.MilvusServiceBlockingStub blockingS
...
+            default:
+                throw new MilvusClientException(ErrorCode.INVALID_PARAMS,
+                        "Invalid targetSizeUnit: '" + request.getTargetSizeUnit()
+                                + "'. Supported units: b, kb, mb, gb, tb, pb");
+        }
+        // Overflow-safe conversion: use BigDecimal so large values (e.g. tb/pb) do not overflow long.
+        long sizeMb = BigDecimal.valueOf(targetSize)
+                .multiply(BigDecimal.valueOf(unitBytes))
+                .divide(BigDecimal.valueOf(1024L * 1024L), 0, RoundingMode.FLOOR)
+                .longValue();
```
yhmo: sdk-core/src/main/java/io/milvus/v2/service/utility/UtilityService.java line:264
Medium ---- The overflow-safe claim is only half true: BigDecimal prevents the intermediate multiply from overflowing, but the final longValue() back-conversion silently returns the low-order 64 bits (no exception) when sizeMb exceeds Long.MAX_VALUE, so a type-valid but extreme input (e.g. targetSize = Long.MAX_VALUE with unit "pb") wraps to a wrong positive value that is sent to the server instead of being rejected. PyMilvus parse_target_size explicitly rejects size_mb > (1<<63)-1 with a ParamError; consider adding the same upper-bound check before the longValue() call.

sdk-core/src/main/java/io/milvus/v2/service/vector/VectorService.java
```diff
@@ -364,6 +385,10 @@ public CompletableFuture<SearchResp> searchAsync(
             SearchReq request, RpcUtils retryUtils) {
         final SearchRequest searchRequest;
         try {
+            if (request.getLimit() <= 0) {
```
yhmo, line 388: sdk-core/src/main/java/io/milvus/v2/service/vector/VectorService.java line:388
Low ---- The new limit<=0 validation is applied in both sync and async variants (search, searchAsync, hybridSearch, hybridSearchAsync), but the added tests (testSearchInvalidLimitRejected, testHybridSearchInvalidLimitRejected) only cover the sync paths. The async variants route the thrown MilvusClientException through catch(Throwable) -> failedFuture; a test asserting that searchAsync/hybridSearchAsync with limit<=0 returns a failed future (rather than throwing synchronously) would prevent that behavior from regressing.

sdk-core/src/main/java/io/milvus/v2/service/utility/UtilityService.java
yhmo: Fixed in 85d7fee6. convertTargetSizeToMB now checks the BigDecimal result against Long.MAX_VALUE before longValue() and throws INVALID_PARAMS when the converted MB exceeds the int64 range, matching PyMilvus parse_target_size. Added testCompactOversizeTargetSizeRejected.

sdk-core/src/main/java/io/milvus/v2/service/vector/VectorService.java
yhmo: Fixed in 85d7fee6. Added testSearchAsyncInvalidLimitReturnsFailedFuture and testHybridSearchAsyncInvalidLimitReturnsFailedFuture asserting the async variants return a failed future (ExecutionException with INVALID_PARAMS) rather than throwing synchronously, and that no RPC is issued.

PR #2058, opened 2026-09-01
feat: support refreshable workload-identity credentials in BulkWriter

alohaha22:
## Problem

`RemoteBulkWriter` hands static one-shot credentials to the storage client at construction time:

- **AWS**: a session triple minted once via STS `AssumeRoleWithWebIdentity` with a fixed `DurationSeconds`. If the requested duration exceeds the role's `MaxSessionDuration` (AWS default 3600s), STS rejects the call outright; if the job outlives the session, all subsequent S3 writes fail with expired credentials.
- **GCP**: an OAuth2 access token fetched once from the metadata server (fixed 1h lifetime), sent as a static `Bearer` header. Jobs longer than 1 hour fail mid-write.
- **Azure**: already fine (`DefaultAzureCredential` refreshes itself).

Any long-running bulk write (e.g. large vector migrations) hits one of these two failure modes, and there is no duration value that works for both short-lived roles and long-running jobs.

## Solution

Let the storage client refresh credentials instead of freezing them:

- `S3ConnectParam` gains an optional `credentialsProvider` (minio `io.minio.credentials.Provider`) channel. When set, it takes precedence over the static accessKey/secretKey/sessionToken.
- **AWS**: callers pass minio's `WebIdentityProvider` (reads `AWS_ROLE_ARN`/`AWS_WEB_IDENTITY_TOKEN_FILE`, omits `DurationSeconds` so the role default applies, and re-assumes the role before each session expires).
- **GCP**: new `GcpMetadataServerCredentialsProvider` fetches tokens from the GCE metadata server, caches them per `expires_in`, and refetches before expiry. The bearer token rides in `Credentials.sessionToken()`, matching the existing convention of the GCP static-token path.
- The GCP bearer interceptor now calls the provider per request instead of stamping one frozen token.

## Compatibility

- Fully additive: when no provider is set, behavior is byte-for-byte identical to before (static ak/sk/sessionToken via `StaticProvider`, static bearer token for GCP).
- The existing 7-arg `getStorageClient` signature is preserved and now delegates to the new `S3ConnectParam`-based factory.
- Works with any S3-compatible storage; no changes to Tencent COS / Alibaba OSS / MinIO paths.

## Testing

- `GcpMetadataServerCredentialsProviderTest`: token fetch, caching while valid, refetch near expiry, fail-fast on metadata server errors (local stub HTTP server).
- `MinioStorageClientTest`: static keys, external provider, GCP bearer header sourced from the provider's session token, GCP static token fallback.
- Full `sdk-bulkwriter` suite: 60/60 passing.

### DIFF
--- a/sdk-bulkwriter/src/main/java/io/milvus/bulkwriter/connect/GcpMetadataServerCredentialsProvider.java
+++ b/sdk-bulkwriter/src/main/java/io/milvus/bulkwriter/connect/GcpMetadataServerCredentialsProvider.java
@@ -1,0 +1,136 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package io.milvus.bulkwriter.connect;
+
+import com.google.gson.JsonObject;
+import com.google.gson.JsonParser;
+import io.minio.credentials.Credentials;
+import io.minio.credentials.Provider;
+import io.minio.messages.ResponseDate;
+import okhttp3.OkHttpClient;
+import okhttp3.Request;
+import okhttp3.Response;
+
+import java.io.IOException;
+import java.time.ZonedDateTime;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * A minio {@link Provider} that fetches OAuth2 access tokens from the GCE metadata
+ * server, for workloads running on GCP with a service-account-attached identity (GKE
+ * workload identity, GCE service account).
+ *
+ * <p>The GCS XML API authenticates with an {@code Authorization: Bearer} header instead
+ * of request signing, so the access token is carried in {@link Credentials#sessionToken()}
+ * (matching the established convention of passing the GCS bearer token as the session
+ * token); accessKey/secretKey hold a sentinel value and are never used.</p>
+ *
+ * <p>Tokens are cached and refetched shortly before the server-reported expiry, so a
+ * long-running writer never sends an expired token. If a refetch fails while the cached
+ * token is still valid, the cached token is served until it truly expires; the cached
+ * fast path is lock-free, since {@code fetch()} is invoked per storage request.</p>
+ */
+public class GcpMetadataServerCredentialsProvider implements Provider {
+    private static final String TOKEN_URL =
+            "http://metadata.google.internal/computeMetadata/v1/instance/service-accounts/default/token";
+    // refetch this far ahead of the server-reported expiry to absorb clock skew and
+    // in-flight requests
+    private static final long REFRESH_MARGIN_MS = 60_000;
+    // minio Credentials rejects empty accessKey/secretKey; the bearer path never reads
+    // them, so a self-describing sentinel fills the slots
+    private static final String BEARER_ONLY = "gcp-bearer-only";
+
+    private final String tokenUrl;
+    private final OkHttpClient httpClient;
+
+    private volatile Credentials cachedCredentials;
+    private volatile long expiresAtMillis;
+
+    public GcpMetadataServerCredentialsProvider() {
+        this(TOKEN_URL, new OkHttpClient.Builder()
+                .connectTimeout(5, TimeUnit.SECONDS)
+                .readTimeout(10, TimeUnit.SECONDS)
+                .build());
+    }
+
+    // visible for testing
+    GcpMetadataServerCredentialsProvider(String tokenUrl, OkHttpClient httpClient) {
+        this.tokenUrl = tokenUrl;
+        this.httpClient = httpClient;
+    }
+
+    @Override
+    public Credentials fetch() {
+        if (cachedCredentials != null && System.currentTimeMillis() < expiresAtMillis - REFRESH_MARGIN_MS) {
+            return cachedCredentials;
+        }
+        synchronized (this) {
+            long now = System.currentTimeMillis();
+            if (cachedCredentials != null && now < expiresAtMillis - REFRESH_MARGIN_MS) {
+                return cachedCredentials;
+            }
+            try {
+                return fetchToken();
+            } catch (IllegalStateException e) {
+                if (cachedCredentials != null && now < expiresAtMillis) {
+                    // the cached token is still valid; serve it while the metadata
+                    // server is unreachable instead of failing in-flight uploads
+                    return cachedCredentials;
+                }
+                throw e;
+            }
+        }
+    }
+
+    private Credentials fetchToken() {
+        Request request = new Request.Builder()
+                .url(tokenUrl)
+                .header("Metadata-Flavor", "Google")
+                .get()
+                .build();
+        String body;
+        try (Response response = httpClient.newCall(request).execute()) {
+            if (!response.isSuccessful() || response.body() == null) {
+                throw tokenFetchFailure("http " + response.code(), null);
+            }
+            body = response.body().string();
+        } catch (IOException e) {
+            throw tokenFetchFailure(e.getMessage(), e);
+        }
+        try {
+            JsonObject json = JsonParser.parseString(body).getAsJsonObject();
+            String token = json.get("access_token").getAsString();
+            long expiresInSeconds = json.get("expires_in").getAsLong();
+            cachedCredentials = new Credentials(BEARER_ONLY, BEARER_ONLY, token,
+                    new ResponseDate(ZonedDateTime.now().plusSeconds(expiresInSeconds)));
+            expiresAtMillis = System.currentTimeMillis() + expiresInSeconds * 1000;
+            return cachedCredentials;
+        } catch (RuntimeException e) {
+            // malformed JSON or missing access_token/expires_in fields
+            throw tokenFetchFailure("unparseable token response: " + e.getMessage(), e);
+        }
+    }
+
+    private static IllegalStateException tokenFetchFailure(String detail, Throwable cause) {
+        String message = "failed to fetch GCP access token from metadata server, " + detail
+                + "; no GCP service account identity is available in this environment";
+        return cause == null ? new IllegalStateException(message) : new IllegalStateException(message, cause);
+    }
+}
--- a/sdk-bulkwriter/src/main/java/io/milvus/bulkwriter/connect/S3ConnectParam.java
+++ b/sdk-bulkwriter/src/main/java/io/milvus/bulkwriter/connect/S3ConnectParam.java
@@ -21,7 +21,9 @@ package io.milvus.bulkwriter.connect;
 
 import io.milvus.exception.ParamException;
 import io.milvus.param.ParamUtils;
+import io.minio.credentials.Provider;
 import okhttp3.OkHttpClient;
+import org.apache.commons.lang3.StringUtils;
 import org.jetbrains.annotations.NotNull;
 
 /**
@@ -36,6 +38,7 @@ public class S3ConnectParam extends StorageConnectParam {
     private final String region;
     private final OkHttpClient httpClient;
     private final String cloudName;
+    private final Provider credentialsProvider;
 
     private S3ConnectParam(@NotNull Builder builder) {
         this.bucketName = builder.bucketName;
@@ -46,6 +49,7 @@ public class S3ConnectParam extends StorageConnectParam {
         this.region = builder.region;
         this.httpClient = builder.httpClient;
         this.cloudName = builder.cloudName;
+        this.credentialsProvider = builder.credentialsProvider;
     }
 
     public String getBucketName() {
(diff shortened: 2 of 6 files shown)


alohaha22:
Thanks for the thorough review — all points addressed in the updated commit:

**7-arg overload regression (critical)**: Confirmed — routing through `S3ConnectParam.build()` made every call fail on the `bucketName` check, breaking `VolumeFileManager.createStorageClient()`. Fixed by extracting the shared logic into a private `newStorageClient(...)` (taking an explicit `credentialsProvider` param); both public overloads delegate to it, and the 7-arg overload no longer goes through `build()` validation. Added `sevenArgOverloadSignsWithStaticKeys` and `sevenArgOverloadGcpStaticTokenSendsBearerHeader` regression tests that would have caught this.

**Mutual exclusivity**: Now enforced in `S3ConnectParam.build()` — setting `credentialsProvider` together with any of accessKey/secretKey/sessionToken throws `ParamException` (covered by `credentialsProviderRejectsStaticKeys`).

**GCP provider exception handling**: JSON parse failures (`JsonSyntaxException`) and missing `access_token`/`expires_in` fields are now wrapped in the same fail-fast `IllegalStateException` with the service-account hint (covered by `failsFastWhenResponseIsNotJson` / `failsFastWhenTokenFieldMissing`).

**Import order**: Fixed (`java.util.concurrent.*` before `java.util.function.Supplier`).

**Per-request fetch note**: Documented in the `getStorageClient(S3ConnectParam)` and `withCredentialsProvider` javadoc that for GCP the provider's `fetch()` is invoked per HTTP request and must cache internally per the minio Provider contract.

**Weak assertions**: `buildsClientWithStaticKeys` / `buildsClientWithExternalCredentialsProvider` now drive a real request against a stub HTTP server and assert the AWS4-HMAC-SHA256 Authorization header carries the expected access key, verifying which provider actually signed.

Full sdk-bulkwriter suite: 65/65 green.

alohaha22:
Small structural follow-up on the fix above: instead of a private helper, the shared logic now lives in a new public 8-arg `getStorageClient(..., OkHttpClient, Provider)` overload — the original 7-arg overload simply delegates with a `null` provider, so its existing signature/semantics stay byte-for-byte intact for current callers like `VolumeFileManager`. Tests re-run: 65/65 green.

alohaha22:
Three nits addressed:

1. The 8-arg `getStorageClient` overload is now `private` — it's purely internal delegation, the public surface stays at the `S3ConnectParam` overload + the original 7-arg overload.
2. `requestCount` removed from `MinioStorageClientTest` (the `authorizationHeader` assertions already prove a request happened).
3. Mutual-exclusion check now uses blank semantics (`isNotBlank`), consistent with `CheckNullEmptyString`.

Tests still 65/65 green.

alohaha22:
Good catch on the Java 8 target — replaced `String.isBlank()` (Java 11+) with commons-lang3 `StringUtils.isNotBlank`, which the module already depends on. Also trimmed the javadoc on the private 8-arg overload down to a short comment. Tests still 65/65 green.

sdk-bulkwriter/src/main/java/io/milvus/bulkwriter/connect/S3ConnectParam.java
```diff
@@ -179,8 +207,17 @@ public Builder withHttpClient(@NotNull OkHttpClient httpClient) {
...
+                    || isNotBlank(sessionToken))) {
+                throw new ParamException("credentialsProvider is mutually exclusive with the static"
+                        + " accessKey/secretKey/sessionToken fields");
+            }
 
             return new S3ConnectParam(this);
         }
+
+        private static boolean isNotBlank(String value) {
+            return value != null && !value.isBlank();
```
yhmo: sdk-bulkwriter/src/main/java/io/milvus/bulkwriter/connect/S3ConnectParam.java line:220
Medium ---- String.isBlank() is a Java 11 API, but this project targets Java 8 (pom.xml maven.compiler.source/target=8, README "Java 8 or higher"). CI compiles on JDK 11 with -source 8 -target 8 and no --release, so this passes CI yet emits bytecode that calls String.isBlank(), which does not exist on a Java 8 runtime -> NoSuchMethodError the first time this new mutual-exclusivity check runs for a Java 8 user. The codebase already uses commons-lang3 StringUtils (ParamUtils.CheckNullEmptyString, MinioStorageClient), so StringUtils.isNotBlank(...) is the consistent, Java 8-safe fix and would let you drop this private helper.

sdk-bulkwriter/src/main/java/io/milvus/bulkwriter/storage/client/MinioStorageClient.java
```diff
@@ -134,11 +194,18 @@ public static MinioStorageClient getStorageClient(String cloudName,
...
+        }
+        return () -> sessionToken;
+    }
+
+    private static OkHttpClient buildAuthorizedClient(OkHttpClient httpClient, Supplier<String> tokenSupplier) {
         Interceptor authInterceptor = chain -> {
             Request original = chain.request();
             Request requestWithAuth = original.newBuilder()
-                    .header("Authorization", "Bearer " + sessionToken)
+                    .header("Authorization", "Bearer " + tokenSupplier.get())
```
yhmo, line 197: sdk-bulkwriter/src/main/java/io/milvus/bulkwriter/storage/client/MinioStorageClient.java line:208
Low ---- `gcpBearerTokenSource` returns `credentialsProvider.fetch().sessionToken()` and this interceptor builds `"Authorization: Bearer " + tokenSupplier.get()`. `withCredentialsProvider` accepts any minio `Provider`, and a user-supplied provider that returns a null/empty sessionToken (or throws) would silently emit `Bearer null` and fail auth with no diagnostic. Consider guarding for a null/empty token and throwing a clear error so a misconfigured provider fails loudly instead of producing an opaque 403 from the storage endpoint.

sdk-bulkwriter/src/main/java/io/milvus/bulkwriter/connect/GcpMetadataServerCredentialsProvider.java
```diff
@@ -0,0 +1,119 @@
...
+
+    // visible for testing
+    GcpMetadataServerCredentialsProvider(String tokenUrl, OkHttpClient httpClient) {
+        this.tokenUrl = tokenUrl;
+        this.httpClient = httpClient;
+    }
+
+    @Override
+    public synchronized Credentials fetch() {
+        if (cachedCredentials != null && System.currentTimeMillis() < expiresAtMillis - REFRESH_MARGIN_MS) {
```
yhmo, line 81: sdk-bulkwriter/src/main/java/io/milvus/bulkwriter/connect/GcpMetadataServerCredentialsProvider.java line:79
Low ---- When the cached token is inside the 60s refresh margin but not yet expired, a transient metadata-server failure makes every fetch() throw; because the GCP bearer interceptor calls fetch() per storage request, in-flight uploads fail even though the still-valid cached token could serve them for up to REFRESH_MARGIN_MS. For the long-running writer this PR targets, consider returning the cached token on refetch failure and only failing once it is truly expired. The current fail-fast behavior is documented and tested, so treat this as a resilience suggestion rather than a bug.

alohaha22:
Latest round of review comments addressed:

**GCP bearer token guard (yhmo)**: `gcpBearerTokenSource` now throws a clear `IllegalStateException` when the provider returns a null/empty sessionToken, instead of silently emitting `Bearer null` and failing with an opaque 403 (`gcpProviderWithoutSessionTokenFailsLoudly` test).

**Resilience on refetch failure (yhmo)**: `fetch()` now serves the cached token when a refetch fails but the token is still valid (i.e. inside the refresh margin but not yet expired), and only fails once the token has truly expired — in-flight uploads no longer break on a transient metadata-server error (`servesCachedTokenWhenRefetchFailsButTokenStillValid` / `failsWhenTokenExpiredAndRefetchFails` tests).

**Lock-free fast path (Copilot)**: cache fields are `volatile` with double-checked refresh, so the common cached case no longer contends on the monitor even though `fetch()` is invoked per storage request.

**Environment-agnostic message (Copilot)**: the failure hint no longer hard-codes "writer pod" — now reads "no GCP service account identity is available in this environment".

Also added coverage for the mutual-exclusion branches (secretKey/sessionToken-only conflicts) and the provided-httpClient path. Suite: 69/69 green.

sdk-bulkwriter/src/main/java/io/milvus/bulkwriter/connect/S3ConnectParam.java
yhmo: The issue is confirmed fixed. Resolving this conversation.

sdk-bulkwriter/src/main/java/io/milvus/bulkwriter/storage/client/MinioStorageClient.java
yhmo: The issue is confirmed fixed. Resolving this conversation.

sdk-bulkwriter/src/main/java/io/milvus/bulkwriter/connect/GcpMetadataServerCredentialsProvider.java
yhmo: The issue is confirmed fixed. Resolving this conversation.

sdk-bulkwriter/src/main/java/io/milvus/bulkwriter/connect/GcpMetadataServerCredentialsProvider.java
```diff
@@ -0,0 +1,119 @@
...
+            return cachedCredentials;
+        }
+        return fetchToken();
+    }
```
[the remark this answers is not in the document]
yhmo: The issue is confirmed fixed. Resolving this conversation.

sdk-bulkwriter/src/main/java/io/milvus/bulkwriter/connect/GcpMetadataServerCredentialsProvider.java
```diff
@@ -0,0 +1,119 @@
...
+            // malformed JSON or missing access_token/expires_in fields
+            throw tokenFetchFailure("unparseable token response: " + e.getMessage(), e);
+        }
+    }
+
+    private static IllegalStateException tokenFetchFailure(String detail, Throwable cause) {
+        String message = "failed to fetch GCP access token from metadata server, " + detail
+                + "; the writer pod is not running with a GCP service account identity";
+        return cause == null ? new IllegalStateException(message) : new IllegalStateException(message, cause);
+    }
```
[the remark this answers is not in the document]
yhmo: The issue is confirmed fixed. Resolving this conversation.

yhmo:
Thank you! I will release a new version later (hopefully this Friday).
/lgtm
/approve

sdk-bulkwriter/src/main/java/io/milvus/bulkwriter/connect/GcpMetadataServerCredentialsProvider.java
```diff
@@ -0,0 +1,136 @@
...
+        this.tokenUrl = tokenUrl;
+        this.httpClient = httpClient;
+    }
+
+    @Override
+    public Credentials fetch() {
+        if (cachedCredentials != null && System.currentTimeMillis() < expiresAtMillis - REFRESH_MARGIN_MS) {
+            return cachedCredentials;
+        }
+        synchronized (this) {
```
yhmo, line 84: sdk-bulkwriter/src/main/java/io/milvus/bulkwriter/connect/GcpMetadataServerCredentialsProvider.java line:84
Medium ---- The refresh path still holds the monitor across the blocking fetchToken() network call (up to the 10s read timeout), and a failed refetch adds no backoff/cooldown. When the metadata server is briefly unreachable inside the refresh margin, every concurrent storage request (the GCP interceptor calls fetch() per request) serializes on this lock and each re-attempts a full network fetch before the still-valid cached token is served - turning a short metadata blip into a thundering herd of serialized, timeout-length stalls on the exact long-running-writer path this PR targets. The lock-free fast path is good; consider doing the network fetch outside the monitor (e.g. single-flight refresh with a separate lock) or adding a short failure cooldown so concurrent callers reuse the cached token instead of all retrying.

sdk-bulkwriter/src/main/java/io/milvus/bulkwriter/connect/GcpMetadataServerCredentialsProvider.java
```diff
@@ -0,0 +1,136 @@
...
+            long now = System.currentTimeMillis();
+            if (cachedCredentials != null && now < expiresAtMillis - REFRESH_MARGIN_MS) {
+                return cachedCredentials;
+            }
+            try {
+                return fetchToken();
+            } catch (IllegalStateException e) {
+                if (cachedCredentials != null && now < expiresAtMillis) {
```
yhmo, line 92: sdk-bulkwriter/src/main/java/io/milvus/bulkwriter/connect/GcpMetadataServerCredentialsProvider.java line:92
Low ---- The fallback checks the cached token against `now`, which was captured before the blocking fetchToken() call (line 85). If the refetch fails slowly (up to the 10s read timeout) and the cached token was already within a few seconds of expiry, by the time this catch block runs the token may have actually expired, yet the stale `now` still satisfies `now < expiresAtMillis`, so an already-expired token is served to in-flight uploads. Consider re-reading `System.currentTimeMillis()` after the failed fetch before deciding whether to serve the cached token.

sdk-bulkwriter/src/main/java/io/milvus/bulkwriter/connect/GcpMetadataServerCredentialsProvider.java
```diff
@@ -0,0 +1,136 @@
...
+    private static final String BEARER_ONLY = "gcp-bearer-only";
+
+    private final String tokenUrl;
+    private final OkHttpClient httpClient;
+
+    private volatile Credentials cachedCredentials;
+    private volatile long expiresAtMillis;
+
+    public GcpMetadataServerCredentialsProvider() {
+        this(TOKEN_URL, new OkHttpClient.Builder()
```
yhmo, line 67: sdk-bulkwriter/src/main/java/io/milvus/bulkwriter/connect/GcpMetadataServerCredentialsProvider.java line:67
Low ---- Each provider instance creates its own OkHttpClient here, but the provider exposes no close()/shutdown and MinioStorageClient.close() only closes the minio httpClient, not this internal one. If callers create a provider per writer (or per test), the metadata client threads and connection pool are never released. Consider reusing a caller-supplied/shared OkHttpClient (S3ConnectParam already supports withHttpClient) or wiring a close() through the storage-client lifecycle.

PR #2060, opened 2026-09-02
Align V2 API parity gaps with PyMilvus 2.6

yhmo:
- decode cost from status.extra_info for insert/upsert/delete
- send num_partitions/properties on fast-path createCollection
- reject null index fieldName and empty index_params with checked errors
- support guaranteeTimestamp override for query/hybridSearch
- surface describeCollection aliases/consistencyLevelName/enableNamespace/ schemaVersion/updateTimestamp and field/function id metadata
- enforce createCollection numeric bounds; guard alterIndexProperties null properties; set delete consistency level

### DIFF
--- a/sdk-core/src/main/java/io/milvus/v2/service/collection/CollectionService.java
+++ b/sdk-core/src/main/java/io/milvus/v2/service/collection/CollectionService.java
@@ -93,6 +93,14 @@ public class CollectionService extends BaseService {
                 .setShardsNum(request.getNumShards())
                 .setConsistencyLevelValue(request.getConsistencyLevel().getCode());
 
+        List<KeyValuePair> propertiesList = ParamUtils.AssembleKvPair(request.getProperties());
+        if (CollectionUtils.isNotEmpty(propertiesList)) {
+            propertiesList.forEach(builder::addProperties);
+        }
+        if (request.getNumPartitions() != null) {
+            builder.setNumPartitions(request.getNumPartitions());
+        }
+
         if (StringUtils.isNotEmpty(dbName)) {
             builder.setDbName(dbName);
         }
--- a/sdk-core/src/main/java/io/milvus/v2/service/index/IndexService.java
+++ b/sdk-core/src/main/java/io/milvus/v2/service/index/IndexService.java
@@ -40,10 +40,17 @@ import java.util.stream.Collectors;
 public class IndexService extends BaseService {
 
     public Void createIndex(MilvusServiceGrpc.MilvusServiceBlockingStub blockingStub, CreateIndexReq request) {
+        if (CollectionUtils.isEmpty(request.getIndexParams())) {
+            throw new MilvusClientException(ErrorCode.INVALID_PARAMS, "IndexParams is empty, no index can be created");
+        }
+
         String dbName = request.getDatabaseName();
         String collectionName = request.getCollectionName();
         for (IndexParam indexParam : request.getIndexParams()) {
             String fieldName = indexParam.getFieldName();
+            if (StringUtils.isEmpty(fieldName)) {
+                throw new MilvusClientException(ErrorCode.INVALID_PARAMS, "Field name cannot be null or empty");
+            }
             String indexName = indexParam.getIndexName();
             String title = String.format("Create index for field: '%s' in collection: '%s' in database: '%s'",
                     fieldName, collectionName, dbName);
@@ -71,9 +78,16 @@ public class IndexService extends BaseService {
             Map<String, Object> extraParams = indexParam.getExtraParams();
             if (extraParams != null && !extraParams.isEmpty()) {
                 for (String key : extraParams.keySet()) {
+                    Object value = extraParams.get(key);
+                    // the dimension must be an integer, keep consistent with PyMilvus create_index_request
+                    // which requires isinstance(dim, int); floats and numeric strings are rejected
+                    if (Constant.VECTOR_DIM.equals(key) && !isLegalDimensionValue(value)) {
+                        throw new MilvusClientException(ErrorCode.INVALID_PARAMS,
+                                "Dimension must be an integer, got: " + value);
+                    }
                     builder.addExtraParams(KeyValuePair.newBuilder()
                             .setKey(key)
-                            .setValue(extraParams.get(key).toString())
+                            .setValue(String.valueOf(value))
                             .build());
                 }
             }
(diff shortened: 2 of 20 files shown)


sdk-core/src/main/java/io/milvus/v2/service/vector/VectorService.java
```diff
@@ -477,6 +479,15 @@ private void fillSearchRespFromExtraInfo(SearchResp.SearchRespBuilder respBuilde
...
     }
 
+    // decode the report value from the status.extra_info, returns null when it is absent
+    private Long getCost(io.milvus.grpc.Status status) {
+        java.util.Map<String, String> extraInfo = status.getExtraInfoMap();
+        if (extraInfo != null && extraInfo.containsKey("report_value")) {
+            return Long.parseLong(extraInfo.get("report_value"));
+        }
+        return null;
+    }
```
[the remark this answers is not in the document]
yhmo: The issue is confirmed fixed. Resolving this conversation.

sdk-core/src/test/java/io/milvus/v2/service/vector/VectorTest.java
```diff
@@ -550,6 +551,120 @@ void testHybridSearchAsync() throws Exception {
...
+        DeleteReq request = DeleteReq.builder()
+                .collectionName("test")
+                .filter("id > 0")
+                .consistencyLevel(ConsistencyLevel.STRONG)
+                .build();
+        client_v2.delete(request);
+
+        ArgumentCaptor<DeleteRequest> captor = ArgumentCaptor.forClass(DeleteRequest.class);
+        verify(blockingStub).delete(captor.capture());
+        Assertions.assertEquals(0, captor.getValue().getConsistencyLevelValue());
```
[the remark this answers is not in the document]
yhmo: The issue is confirmed fixed. Resolving this conversation.

sdk-core/src/main/java/io/milvus/v2/utils/VectorUtils.java
```diff
@@ -70,7 +70,9 @@ public QueryRequest ConvertToGrpcQueryRequest(QueryReq request) {
         }
         String dbName = request.getDatabaseName();
         String collectionName = request.getCollectionName();
-        long guaranteeTimestamp = getGuaranteeTimestamp(request.getConsistencyLevel(), dbName, collectionName);
+        long guaranteeTimestamp = request.getGuaranteeTimestamp() > 0
```
yhmo: sdk-core/src/main/java/io/milvus/v2/utils/VectorUtils.java line:73
Medium ---- The guaranteeTimestamp override is silently dropped by the 2.6 server when consistencyLevel is left null. If the caller sets guaranteeTimestamp>0 but keeps the default consistencyLevel=null, this line sends guarantee_timestamp=<override>, yet the code below still calls builder.setUseDefaultConsistency(true) (VectorUtils.java:95). On the server (internal/proxy/task_query.go, 2.6), use_default_consistency=true makes the proxy recompute guaranteeTs from the collection's own consistency level via parseGuaranteeTsFromConsistency, so for the common Bounded/Strong/Eventually collections the client's explicit timestamp is discarded (it only survives for Session/Customized collections). The new tests only assert the wire value (getGuaranteeTimestamp()==12345) so they do not catch this. Consider setting useDefaultConsistency=false when an explicit guaranteeTimestamp is provided (or requiring/deriving consistencyLevel), and note the same pattern exists at line 676 for hybridSearch.

yhmo: The issue is confirmed fixed. Resolving this conversation.

sdk-core/src/main/java/io/milvus/v2/service/collection/CollectionService.java
```diff
@@ -93,6 +93,14 @@ public Void createCollection(MilvusServiceGrpc.MilvusServiceBlockingStub blockin
                 .setShardsNum(request.getNumShards())
                 .setConsistencyLevelValue(request.getConsistencyLevel().getCode());
 
+        List<KeyValuePair> propertiesList = ParamUtils.AssembleKvPair(request.getProperties());
+        if (CollectionUtils.isNotEmpty(propertiesList)) {
+            propertiesList.forEach(builder::addProperties);
+        }
+        if (request.getNumPartitions() != null) {
+            builder.setNumPartitions(request.getNumPartitions());
```
yhmo, line 101: sdk-core/src/main/java/io/milvus/v2/service/collection/CollectionService.java line:101
Medium ---- Fast-path createCollection now forwards numPartitions, but this path only builds a schema with a vector field and the primary-key field, so it never declares a partition-key field. The 2.6 proxy rejects any CreateCollectionRequest with num_partitions != 0 when no partition key field exists ("num_partitions should only be specified with partition key field enabled", internal/proxy/task.go:314-317). Callers who previously set numPartitions on the fast path had it silently ignored and the collection created; after this change the same request fails server-side, and the new unit test only asserts the mock wire value so it cannot catch that. Consider sending numPartitions only when a partition-key field is present (the schema path already supports it), or dropping it from the fast path.

sdk-core/src/main/java/io/milvus/v2/service/index/IndexService.java
```diff
@@ -71,9 +78,15 @@ public Void createIndex(MilvusServiceGrpc.MilvusServiceBlockingStub blockingStub
             Map<String, Object> extraParams = indexParam.getExtraParams();
             if (extraParams != null && !extraParams.isEmpty()) {
                 for (String key : extraParams.keySet()) {
+                    Object value = extraParams.get(key);
+                    // the dimension must be an integer, keep consistent with other SDKs
+                    if (Constant.VECTOR_DIM.equals(key) && !(value instanceof Integer) && !(value instanceof Long)) {
```
yhmo: sdk-core/src/main/java/io/milvus/v2/service/index/IndexService.java line:83
Low ---- The new dim type check only accepts Integer/Long, which is stricter than the stated parity target: PyMilvus is_legal_dimension() does int(dim) and accepts numeric strings and float values (e.g. "128" or 128.0), which commonly arrive as Double from Gson/JSON parsing. Such callers previously had the value serialized via toString() and sent to the server; they now get a hard INVALID_PARAMS error. Consider accepting any Number that parses to an integer, or numeric strings, to avoid a silent compatibility break.

sdk-core/src/main/java/io/milvus/v2/service/collection/CollectionService.java
yhmo: The issue is confirmed fixed. Resolving this conversation.

sdk-core/src/main/java/io/milvus/v2/service/index/IndexService.java
yhmo: The issue is confirmed fixed. Resolving this conversation.

PR #2070, opened 2026-09-04
Reorganize sdk-core tests into unit/integration/system layers

yhmo:
Restructure the sdk-core test tree by test type (unit/integration/system), Maven version (v1/v2/common), then business domain, and split the oversized monolithic test classes into per-domain files.

- Move shared test infrastructure (TestUtils, BaseTest, MockMilvusServer) under io.milvus.support, and docker-compose files under test resources.
- Split 13 monolithic test classes (MilvusClientV2DockerTest, MilvusClientDockerTest, MilvusServiceClientTest, VectorTest, CollectionTest, MilvusClientV2Test, ClientTelemetryManagerTest, DataUtilsTest, RpcUtilsTest, SchemaCacheTest, IteratorTest, RoaringBitmapUtilsTest, MilvusMultiClientDockerTest) into focused domain-scoped test classes.
- Tag every test class with @Tag(unit|integration|system) and add surefire integration/system profiles so `mvn test` runs unit tests only by default, -Pintegration adds integration tests, and -Psystem runs the full suite.
- Add Docker test base classes with a reference-counted Milvus lifecycle so the split system-test classes share a single server start/stop.


pom.xml
```diff
@@ -357,6 +370,7 @@
                 <version>${maven.surefire.plugin.version}</version>
                 <configuration>
                     <skipTests>false</skipTests>
+                    <groups>${test.groups}</groups>
```
yhmo, line 373: pom.xml line:373
High ---- The new `groups` filter defaults to `unit` and is inherited by every module, including `sdk-bulkwriter`, whose test classes carry no `@Tag`. Under JUnit5 tag filtering, untagged tests are excluded whenever `groups` is set, so `mvn verify` (CI runs `mvn -Pcoverage verify`) silently runs ZERO sdk-bulkwriter tests under every profile — the coverage CI uploads from `sdk-bulkwriter/target/site/jacoco/jacoco.xml` will collapse. Either tag the bulkwriter tests (e.g. `@Tag("unit")`) or scope the `groups` filter so it does not disable that module's suite.

pom.xml
```diff
@@ -271,6 +272,18 @@
...
             </build>
         </profile>
+        <profile>
+            <id>integration</id>
+            <properties>
+                <test.groups>unit,integration</test.groups>
+            </properties>
+        </profile>
+        <profile>
+            <id>system</id>
```
yhmo, line 282: pom.xml line:282
Medium ---- CI (`maven.yml`) still runs `mvn -Pcoverage verify` without `-Psystem`/`-Pintegration`, so with the new default `test.groups=unit` the docker-based system tests and the mock-server integration tests will no longer execute in CI at all; previously `mvn verify` ran everything. Please update the workflow to activate the new profiles (e.g. `-Psystem`) or the reorganized suites silently drop out of CI coverage.

sdk-core/src/test/java/io/milvus/support/v2/MilvusV2DockerTestBase.java
```diff
@@ -0,0 +1,347 @@
...
+        ByteBuffer vector = ByteBuffer.allocate(dimension);
+        for (int k = 0; k < dimension; ++k) {
+            vector.put((byte) (RANDOM.nextInt(256) - 128));
+        }
+        return vector;
+    }
+    @BeforeAll
+    public static void setUp() {
+        synchronized (MilvusV2DockerTestBase.class) {
+            if (lifecycleRefCount.getAndIncrement() == 0) {
```
yhmo: sdk-core/src/test/java/io/milvus/support/v2/MilvusV2DockerTestBase.java line:118
Medium ---- The reference-counted lifecycle does not actually share one server start/stop across the split system-test classes under the default surefire execution model. Without `junit.jupiter.execution.parallel` enabled, JUnit5 runs test classes sequentially in a single JVM, so each subclass runs its inherited static `@BeforeAll` (refcount 0→1, starts docker) followed by its `@AfterAll` (refcount 1→0, stops docker) before the next class begins. The refcount therefore never accumulates beyond 1 and docker is torn down and restarted for every one of the ~28 system classes — the opposite of the stated goal. If the intent is a shared server, class-level parallel execution (or a suite-level `@TestInstance` lifecycle) is needed; otherwise this indirection should be removed or documented as per-class. Also note a docker-unavailable `@BeforeAll` assumption failure would leak the refcount and make later classes skip the server start while `client` stays null.

sdk-core/src/test/java/io/milvus/telemetry/TelemetryConfigPushTest.java
```diff
@@ -0,0 +1,291 @@
...
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicInteger;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+@Tag("integration")
```
yhmo: sdk-core/src/test/java/io/milvus/telemetry/TelemetryConfigPushTest.java line:45
Low ---- The split telemetry tests (e.g. this class, TelemetryHeartbeatTest, TelemetryMetricsTest) exercise an in-process gRPC server and direct ClientTelemetryManager construction — they do not need a real Milvus — yet they are tagged `integration`, so they are excluded from the default `mvn test` (unit only) run. Previously the monolith ClientTelemetryManagerTest was untagged and ran under the default `mvn test`/CI. Combined with CI not activating `-Pintegration`, these tests silently drop out of the default/CI runs. If they are meant to run by default, tag them `unit`; otherwise confirm the coverage shift is intentional.

pom.xml
```diff
@@ -357,6 +370,7 @@
                 <version>${maven.surefire.plugin.version}</version>
                 <configuration>
                     <skipTests>false</skipTests>
+                    <groups>${test.groups}</groups>
                 </configuration>
```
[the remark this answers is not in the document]
yhmo: Not changing. I verified the tag filter works on surefire 2.19.1 + junit-platform 1.1.0: default `mvn test` runs only unit (329 in sdk-core), `-Pintegration` adds integration (636), and `-Psystem` includes system tests. The concern about tags being ignored did not reproduce.

sdk-core/src/test/java/io/milvus/system/v2/client/ClientPoolDockerTest.java
```diff
@@ -0,0 +1,189 @@
...
+            int threadCount = 20;
+            int requestCount = 50000;
+            ExecutorService executor = Executors.newFixedThreadPool(threadCount);
+            for (int i = 0; i < requestCount; i++) {
+                Runnable worker = new Worker(i);
+                executor.execute(worker);
+            }
+            executor.shutdown();
+            if (!executor.awaitTermination(100, TimeUnit.SECONDS)) {
+                System.err.println("Executor did not terminate in the specified time.");
```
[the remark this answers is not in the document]
yhmo: Not changing. This is an opt-in stress test that only runs under `-Psystem`; the default and CI workflows do not run it. Leaving the load configuration as-is.

sdk-core/src/test/java/io/milvus/system/v2/client/AnalyzerDockerTest.java
```diff
@@ -0,0 +1,166 @@
...
+        texts.add("2.5 to take advantage of enhancements and fixes!");
+
+        Map<String, Object> analyzerParams = new HashMap<>();
+        analyzerParams.put("tokenizer", "standard");
+        analyzerParams.put("filter",
+                Arrays.asList("lowercase",
+                        new HashMap<String, Object>() {{
+                            put("type", "stop");
+                            put("stop_words", Arrays.asList("to", "of", "for", "the"));
+                        }}));
```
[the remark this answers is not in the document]
yhmo: Fixed in ef8feedd: replaced the double-brace HashMap initializer in AnalyzerDockerTest with a plain local map.

sdk-core/src/test/java/io/milvus/system/v2/client/RbacDockerTest.java
```diff
@@ -0,0 +1,56 @@
...
+
+        ListPrivilegeGroupsResp resp = client.listPrivilegeGroups(ListPrivilegeGroupsReq.builder().build());
+        List<PrivilegeGroup> groups = resp.getPrivilegeGroups();
+        Map<String, List<String>> groupsPlivileges = new HashMap<>();
+        for (PrivilegeGroup group : groups) {
+            groupsPlivileges.put(group.getGroupName(), group.getPrivileges());
+        }
+        Assertions.assertTrue(groupsPlivileges.containsKey("dummy"));
+        Assertions.assertEquals(1, groupsPlivileges.get("dummy").size());
+        Assertions.assertEquals("CreateCollection", groupsPlivileges.get("dummy").get(0));
```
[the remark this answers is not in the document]
yhmo: Fixed in ef8feedd: corrected `groupsPlivileges` to `groupsPrivileges` in RbacDockerTest.

pom.xml
yhmo: Fixed in ef8feedd: tagged all six sdk-bulkwriter test classes with @Tag("unit"), so the root surefire `groups=unit` filter no longer drops them. `mvn -pl sdk-core,sdk-bulkwriter test` now runs 329 + 69 tests, all green.

pom.xml
yhmo: Fixed in ef8feedd: the CI maven workflow now runs `mvn -Pcoverage -Pintegration verify`, so mock-server integration tests stay in CI. Docker-backed system tests remain opt-in via `-Psystem` (not run in the standard PR workflow to avoid the Docker dependency).

sdk-core/src/test/java/io/milvus/support/v2/MilvusV2DockerTestBase.java
yhmo: Fixed in ef8feedd: removed the reference-counted lifecycle from the Docker test base classes. Under sequential surefire execution the counter never exceeded 1, so each concrete system-test class now starts/stops its own standalone, documented in the base class comment. This also removes the refcount leak on a docker-unavailable @BeforeAll.

sdk-core/src/test/java/io/milvus/telemetry/TelemetryConfigPushTest.java
yhmo: Fixed in ef8feedd: retagged the six split telemetry tests from integration to unit (they use in-process gRPC servers and need no real Milvus), so they run again under the default `mvn test` and CI coverage.

pom.xml
yhmo: Follow-up in aa71a744: CI now runs `mvn -Pcoverage -Psystem verify`, restoring the Docker-backed system tests to the standard workflow as before the reorg.

sdk-core/src/test/java/io/milvus/support/v2/MilvusV2DockerTestBase.java
yhmo: Follow-up in aa71a744: instead of per-class start/stop, TestUtils.startMilvusStandalone is now idempotent per compose file and registers a JVM shutdown hook, so the shared Milvus standalone is started once and torn down once at JVM exit across all split system-test classes. Verified: running VersionDockerTest + IndexDockerTest in one JVM starts compose once (first class 33s) and reuses it (second class 2.4s), with containers cleaned up after exit.

sdk-core/src/test/java/io/milvus/system/v2/client/MilvusClientV2SessionDockerTest.java
```diff
@@ -49,6 +53,7 @@
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.TimeUnit;
 
+@Tag("system")
```
yhmo, line 56: sdk-core/src/test/java/io/milvus/system/v2/client/MilvusClientV2SessionDockerTest.java line:56
Medium ---- This class is tagged `@Tag("system")` and now runs in the same JVM as the split system-test classes that share one Milvus standalone via `TestUtils.startMilvusStandalone`, but its `@AfterAll` still calls `TestUtils.stopMilvusStandalone()` (line 87). If it executes between two split classes, it tears down the shared stack mid-suite and forces an expensive full restart (compose down -v + up + wait) for whichever class runs next, contradicting the new shared-lifecycle design ("torn down once at JVM exit"). Consider extending `MilvusV2DockerTestBase` or removing the explicit stop.

sdk-core/src/test/java/io/milvus/system/v2/service/vector/NullableVectorTest.java
```diff
@@ -40,6 +40,7 @@
 import java.util.*;
 
 @TestMethodOrder(MethodOrderer.OrderAnnotation.class)
+@Tag("system")
```
yhmo, line 43: sdk-core/src/test/java/io/milvus/system/v2/service/vector/NullableVectorTest.java line:43
Medium ---- Same shared-lifecycle concern as `MilvusClientV2SessionDockerTest`: this class is tagged `@Tag("system")` and runs in the same JVM as the split classes sharing one Milvus standalone, but its `@AfterAll` still calls `TestUtils.stopMilvusStandalone()` (line 103). If it runs between split classes, the shared stack is torn down mid-suite and restarted (wiping data) for the next class, defeating the new shared-lifecycle design. Consider extending `MilvusV2DockerTestBase` or removing the explicit stop.

sdk-core/src/test/java/io/milvus/system/v2/client/CollectionDockerTest.java
```diff
@@ -0,0 +1,845 @@
...
+        // check the schema
+        DescribeCollectionResp descResp = client.describeCollection(DescribeCollectionReq.builder()
+                .collectionName(randomCollectionName)
+                .build());
+
+        CreateCollectionReq.CollectionSchema collSchema = descResp.getCollectionSchema();
+        CreateCollectionReq.FieldSchema fieldSchema = collSchema.getField("text");
+        Assertions.assertNotNull(fieldSchema);
+        Assertions.assertTrue(fieldSchema.getEnableAnalyzer());
+        Assertions.assertTrue(fieldSchema.getEnableAnalyzer());
```
yhmo, line 498: sdk-core/src/test/java/io/milvus/system/v2/client/CollectionDockerTest.java line:498
Low ---- Duplicate assertion: `fieldSchema.getEnableAnalyzer()` is checked twice (lines 497-498) while the schema sets both `enableAnalyzer(true)` and `enableMatch(true)`, so the second assert was almost certainly meant to be `getEnableMatch()`. Pre-existing in the original monolith but preserved by the split; worth fixing while this file is being reorganized.

sdk-core/src/test/java/io/milvus/system/v2/client/DatabaseDockerTest.java
```diff
@@ -0,0 +1,148 @@
...
+import org.junit.jupiter.api.Test;
+
+@Tag("system")
+class DatabaseDockerTest extends MilvusV2DockerTestBase {
+    @Test
+    void testDatabase() {
+        // get current database
+        ListDatabasesResp listDatabasesResp = client.listDatabases();
+        List<String> dbNames = listDatabasesResp.getDatabaseNames();
+        Assertions.assertEquals(1, dbNames.size());
```
yhmo: sdk-core/src/test/java/io/milvus/system/v2/client/DatabaseDockerTest.java line:43
High ---- This assertion assumes a fresh Milvus with only the `default` database, but the split system-test classes now share one standalone instance in the same JVM (via the idempotent `TestUtils.startMilvusStandalone`). `CollectionDockerTest.testOperationsAcrossDB` (CollectionDockerTest.java line:226) creates database `db_temp` and never drops it; because CollectionDockerTest runs before DatabaseDockerTest in surefire class order, `db_temp` still exists here, so `testDatabase` fails on CI ("Build and test" run: expected: <1> but was: <4>). Please use a unique database name per test and/or drop `db_temp` in a finally block.

sdk-core/src/test/java/io/milvus/support/v2/MilvusV2DockerTestBase.java
```diff
@@ -0,0 +1,337 @@
...
+    protected static final File DockerComposeVolumeDirectory = new File("target/milvus-compose");
+    protected static final List<String> DockerComposeContainerNames = Arrays.asList("milvus-javasdk-etcd", "milvus-javasdk-minio", "milvus-javasdk-standalone");
+    protected static ByteBuffer generateStructInt8Vector(int dimension) {
+        ByteBuffer vector = ByteBuffer.allocate(dimension);
+        for (int k = 0; k < dimension; ++k) {
+            vector.put((byte) (RANDOM.nextInt(256) - 128));
+        }
+        return vector;
+    }
+    // All split system-test classes in one JVM share a single Milvus standalone.
```
yhmo, line 113: sdk-core/src/test/java/io/milvus/support/v2/MilvusV2DockerTestBase.java line:113
Medium ---- The shared-lifecycle contract documented here (one Milvus standalone per JVM, torn down once at JVM exit by a shutdown hook) is violated by two split classes that did not migrate to these base classes: MilvusClientV2SessionDockerTest.java:87 and NullableVectorTest.java:103 call TestUtils.stopMilvusStandalone() in @AfterAll, tearing the shared stack down mid-suite. Since surefire runs all ~28 system classes sequentially in one JVM, whichever class runs immediately after one of these pays a full docker compose down/up plus an up-to-300s health wait in its @BeforeAll, and `down -v` wipes volumes mid-suite. The suite therefore performs up to 2 extra full stack restarts per run, order-dependent. Consider migrating these two classes onto the shared base classes (or removing the explicit stop) so the JVM shutdown hook remains the single teardown point.

sdk-core/src/test/java/io/milvus/integration/v1/client/MilvusServiceClientConnectTest.java
```diff
@@ -0,0 +1,325 @@
...
+import java.lang.reflect.Method;
+import java.util.*;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeUnit;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+@Tag("integration")
+public class MilvusServiceClientConnectTest {
+    private final int testPort = 53019;
```
yhmo, line 45: sdk-core/src/test/java/io/milvus/integration/v1/client/MilvusServiceClientConnectTest.java line:45
Medium ---- Both integration classes hard-code the same mock-server port 53019 (ConnectTest and MilvusServiceClientRpcDispatchTest:57) and surefire runs them sequentially in the same JVM. Several tests in both classes (e.g. ConnectTest testConnect, RpcDispatchTest loadCollection/loadPartitions/createIndex/dropIndex) call server.stop() only at the end of the happy path; if an assertion fails before that, the gRPC server leaks on 53019. MockMilvusServer.start() swallows a bind exception (logs and returns with rpcServer == null), so the next test silently connects to the leaked server, which still holds the previous test's MockMilvusServerImpl — the new test's setXxxResponse values are never seen and failures cascade in confusing ways. Consider a per-test random port or always stopping the server in a finally/@AfterEach.

sdk-core/src/test/java/io/milvus/system/v1/client/DatabaseDockerTest.java
```diff
@@ -0,0 +1,243 @@
...
+        R<RpcStatus> dropResponse = client.dropDatabase(dropDatabaseParam);
+        Assertions.assertEquals(R.Status.Success.getCode(), dropResponse.getStatus().intValue());
+    }
+
+    @Test
+    void testCacheCollectionSchema() throws InterruptedException {
+        String randomCollectionName = generator.generate(10);
+
+        // create a new db
+        String testDbName = "test_database";
```
yhmo, line 89: sdk-core/src/test/java/io/milvus/system/v1/client/DatabaseDockerTest.java line:89
Medium ---- Both testDatabase() and testCacheCollectionSchema() create a fixed-name database "test_database"; testDatabase() drops it, but testCacheCollectionSchema() never does. JUnit 5 does not guarantee method execution order (default MethodOrderer is deterministic but not source order), so if testCacheCollectionSchema() runs first, testDatabase()'s createDatabase("test_database") fails with "database already exists". Use a unique per-test database name (or drop it in a finally) to remove the order dependence.

sdk-core/src/test/java/io/milvus/system/v1/client/ClientPoolDockerTest.java
```diff
@@ -0,0 +1,106 @@
...
+                        }
+                    }
+                    System.out.printf("Thread %s finished%n", Thread.currentThread().getName());
+                });
+                t.start();
+                threadList.add(t);
+            }
+
+            for (Thread t : threadList) {
+                t.join();
```
yhmo, line 86: sdk-core/src/test/java/io/milvus/system/v1/client/ClientPoolDockerTest.java line:86
Low ---- This v1 pool test runs 20 threads x 10,000 requests (200k getVersion calls) and then waits with a bare t.join() (no timeout). The v2 counterpart (system/v2/client/ClientPoolDockerTest) bounds the same wait with awaitTermination(100s). If any worker thread stalls (e.g. pool.getClient blocks when the pool is exhausted), this test can hang the whole -Psystem suite indefinitely with no way to time out. Mirror the v2 bounded-wait pattern (awaitTermination with a limit) so a stalled pool degrades to a timeout instead of an infinite hang.

sdk-core/src/test/java/io/milvus/support/v2/MilvusV2DockerTestBase.java
```diff
@@ -0,0 +1,337 @@
...
+import io.milvus.v2.service.vector.request.ranker.RRFRanker;
+import io.milvus.v2.service.vector.request.ranker.WeightedRanker;
+import io.milvus.v2.service.vector.response.*;
+import io.milvus.v2.utils.DataUtils;
+import org.apache.commons.lang3.StringUtils;
+import org.apache.commons.text.RandomStringGenerator;
+import org.junit.jupiter.api.AfterAll;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Tag;
```
yhmo, line 85: sdk-core/src/test/java/io/milvus/support/v2/MilvusV2DockerTestBase.java line:85
Low ---- org.junit.jupiter.api.Tag and org.junit.jupiter.api.Test are imported here but never used in this abstract base class (same in MilvusV1DockerTestBase and MilvusMultiDockerTestBase). The concrete subclasses carry @Tag("system"); the base classes carry no test methods. Remove the unused imports.

.github/workflows/maven.yml
```diff
@@ -34,8 +34,12 @@ jobs:
         uses: actions/checkout@v4
       - name: Update submodule
         run: git submodule update --init
-      - name: Build with Maven
+      - name: Unit tests (ut)
         run: mvn -Pcoverage --batch-mode --update-snapshots verify
+      - name: Integration tests (it)
+        run: mvn -Pintegration -Dtest.groups=integration --batch-mode --update-snapshots verify
```
yhmo, line 40: .github/workflows/maven.yml line:40
Medium ---- Coverage signal now reflects unit tests only. Only the "Unit tests (ut)" step passes -Pcoverage; the integration step (-Pintegration -Dtest.groups=integration) and the system step (-Psystem -Dtest.groups=system) run without the coverage profile, so the jacoco report uploaded by the codecov step is generated purely from unit-test execution. The previous workflow ran a single `mvn -Pcoverage verify` over all groups, so integration and system test paths used to contribute to codecov. If the coverage gate is meant to keep tracking RPC-dispatch and docker-backed paths, add -Pcoverage to the it/st steps (or keep one combined coverage run); otherwise the project/patch coverage numbers will silently drop.

sdk-core/src/test/java/io/milvus/system/v2/client/CollectionDockerTest.java
```diff
@@ -0,0 +1,845 @@
...
+            if (tempClient != null) {
+                tempClient.close();
+            }
+        }
+    }
+
+    @Test
+    void testOperationsAcrossDB() {
+        // create a temp database
+        String tempDatabaseName = "db_temp";
```
yhmo, line 226: sdk-core/src/test/java/io/milvus/system/v2/client/CollectionDockerTest.java line:226
Medium ---- With the shared-server lifecycle (one Milvus standalone for all split system-test classes in the same JVM, torn down only at JVM exit), this fixed-name db_temp database created here is never dropped by this test. DatabaseDockerTest had to add a defensive dropDatabase guard (see its "the shared standalone may still hold a db_temp left by another test class" comment), which only works if DatabaseDockerTest happens to run after this class and after any mid-test failure. JUnit does not guarantee class execution order, so the suite is now order-dependent: if DatabaseDockerTest fails partway and leaves db_temp behind, this createDatabase("db_temp") will fail. Recommend dropping db_temp in a finally/@AfterAll here (or using a unique/random database name) so each class is self-contained on the shared server.

sdk-core/src/test/java/io/milvus/system/v1/client/AliasDockerTest.java
```diff
@@ -0,0 +1,111 @@
...
+        Assertions.assertEquals(R.Status.Success.getCode(), listResp.getStatus().intValue());
+        Assertions.assertEquals(listResp.getData().getAliases(0), "alias_A");
+        Assertions.assertEquals(listResp.getData().getCollectionName(), "coll_A");
+
+        // alter alias
+        response = client.alterAlias(AlterAliasParam.newBuilder()
+                .withAlias("alias_A")
+                .withCollectionName("coll_B")
+                .build());
+        Assertions.assertEquals(R.Status.Success.getCode(), listResp.getStatus().intValue());
```
yhmo, line 84: sdk-core/src/test/java/io/milvus/system/v1/client/AliasDockerTest.java line:84
Low ---- After alterAlias, the assertion still checks listResp.getStatus() (the previous listAliases response) instead of the just-assigned response. Both currently succeed, so the test passes, but a failed alterAlias would go unnoticed. This was carried over from the original MilvusClientDockerTest; while the test is being split, consider asserting on response.

sdk-core/src/test/java/io/milvus/system/v2/client/MilvusClientV2SessionDockerTest.java
yhmo: Fixed: removed TestUtils.stopMilvusStandalone() from @AfterAll in MilvusClientV2SessionDockerTest. The class now shares the single Milvus standalone lifecycle, torn down once at JVM exit by the shutdown hook.

sdk-core/src/test/java/io/milvus/system/v2/service/vector/NullableVectorTest.java
yhmo: Fixed: removed TestUtils.stopMilvusStandalone() from @AfterAll in NullableVectorTest. It now reuses the shared standalone instead of tearing the stack down mid-suite.

sdk-core/src/test/java/io/milvus/system/v2/client/CollectionDockerTest.java
yhmo: Fixed: changed the duplicate getEnableAnalyzer() assertion to getEnableMatch(), matching the enableAnalyzer/enableMatch fields set on the schema.

sdk-core/src/test/java/io/milvus/system/v2/client/DatabaseDockerTest.java
yhmo: Fixed: testDatabase no longer assumes a pristine server (asserts the default database exists instead of an exact count) and drops a leftover db_temp defensively. Verified green in CI.

sdk-core/src/test/java/io/milvus/support/v2/MilvusV2DockerTestBase.java
yhmo: Fixed: migrated both MilvusClientV2SessionDockerTest and NullableVectorTest off their explicit stopMilvusStandalone calls so the JVM shutdown hook is the single teardown point, honoring the base-class shared-lifecycle contract.

sdk-core/src/test/java/io/milvus/integration/v1/client/MilvusServiceClientConnectTest.java
yhmo: Fixed: both mock-server integration classes now allocate a per-instance random free port instead of sharing hard-coded 53019, so a leaked server from a failed assertion cannot bleed into the next test.

sdk-core/src/test/java/io/milvus/system/v1/client/DatabaseDockerTest.java
yhmo: Fixed: V1 DatabaseDockerTest now uses unique per-test database names (generator-based) instead of the shared fixed name, removing the testDatabase/testCacheCollectionSchema order dependence.

sdk-core/src/test/java/io/milvus/system/v1/client/ClientPoolDockerTest.java
yhmo: Fixed: the v1 pool stress test now uses a bounded t.join(100s) plus an assertion that all worker threads finished, mirroring the v2 awaitTermination pattern so a stalled pool degrades to a timeout.

sdk-core/src/test/java/io/milvus/support/v2/MilvusV2DockerTestBase.java
yhmo: Fixed: removed the unused org.junit.jupiter.api.Tag and org.junit.jupiter.api.Test imports from MilvusV2DockerTestBase, MilvusV1DockerTestBase, and MilvusMultiDockerTestBase.

.github/workflows/maven.yml
yhmo: Fixed: the maven workflow now runs -Pcoverage on all three steps with per-step jacoco destFile/reportDir, and codecov uploads the ut/it/st reports so integration and system paths still contribute to coverage.

sdk-core/src/test/java/io/milvus/system/v2/client/CollectionDockerTest.java
yhmo: Fixed: V2 CollectionDockerTest.testOperationsAcrossDB now drops db_temp at the end, making it self-contained on the shared server and removing the order dependence with DatabaseDockerTest.

sdk-core/src/test/java/io/milvus/system/v1/client/AliasDockerTest.java
yhmo: Fixed: AliasDockerTest now asserts response.getStatus() right after alterAlias instead of the stale listResp status.

.github/workflows/maven.yml
```diff
@@ -30,16 +30,42 @@ jobs:
...
-      - name: Checkout
-        uses: actions/checkout@v4
       - name: Update submodule
         run: git submodule update --init
-      - name: Build with Maven
-        run: mvn -Pcoverage --batch-mode --update-snapshots verify
+      - name: BulkWriter unit tests (ut)
+        run: mvn -pl sdk-bulkwriter -Pcoverage "-Djacoco.destFile=target/jacoco-ut.exec" "-Djacoco.reportDir=target/site/jacoco-ut" --batch-mode --update-snapshots verify
+      - name: System tests (st)
+        run: mvn -Pcoverage -Psystem "-Dtest.groups=system" "-Djacoco.destFile=target/jacoco-st.exec" "-Djacoco.reportDir=target/site/jacoco-st" --batch-mode --update-snapshots verify
```
yhmo, line 40: .github/workflows/maven.yml line:38
Medium ---- With `-Dtest.groups=system` here (which overrides the `-Psystem` profile property `test.groups=unit,integration,system`), and the BulkWriter step above scoped to `-pl sdk-bulkwriter`, the ubuntu job no longer executes any sdk-core unit (~2180) or integration (~333) tests — those suites now run only on the windows-2022 job. The previous single-job workflow ran the whole sdk-core suite on ubuntu (Linux). Since Milvus is primarily deployed on Linux, consider keeping the sdk-core unit suite on the ubuntu runner as well (or explicitly documenting that the main unit/integration suites are now Windows-only), so a Linux-specific failure in those suites cannot slip through CI undetected.

sdk-core/src/test/java/io/milvus/integration/v2/client/ConnectConfigTest.java
```diff
@@ -0,0 +1,62 @@
...
+
+package io.milvus.integration.v2.client;
+
+import io.milvus.support.v2.BaseTest;
+import io.milvus.v2.client.ConnectConfig;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Tag;
+import org.junit.jupiter.api.Test;
+
+@Tag("integration")
```
yhmo: sdk-core/src/test/java/io/milvus/integration/v2/client/ConnectConfigTest.java line:28
Low ---- This class (and BuilderConventionTest) are pure unit tests — ConnectConfigTest only checks builder defaults and toString redaction, no server and no mock stubs are exercised — yet they are tagged `integration`. Under the new `test.groups=unit` default they drop out of `mvn test`, and because the ubuntu CI job now runs only `-Dtest.groups=system`, they execute nowhere on the Linux runner (they run only in the windows integration step). This is the same situation the telemetry split tests were in before they were retagged to `unit`; consider tagging these two as `unit` so they stay in the default and Linux unit runs.

.github/workflows/maven.yml
yhmo: Fixed in 92a05904: the ubuntu job now runs a dedicated sdk-core unit tests step (`-pl sdk-core`, default unit group) alongside the BulkWriter unit step and the system step, so the sdk-core unit/integration suites run on both the Linux (ubuntu) and Windows runners. jacoco destFile/reportDir are separated per step (jacoco-ut-core / jacoco-ut-bulk / jacoco-st) so coverage reports do not overwrite each other, and codecov uploads all three.

sdk-core/src/test/java/io/milvus/integration/v2/client/ConnectConfigTest.java
yhmo: Fixed in 92a05904: the integration-side ConnectConfigTest (2 tests) was redundant with the existing 17-test unit version and has been deleted; BuilderConventionTest was moved to unit/v2/client, retagged @Tag("unit"), and no longer extends BaseTest (it uses no mock stubs). Both now run in the default `mvn test` and in the Linux unit run. Verified: unit suite is 2182 tests, all green.

yhmo:
@Mergifyio refresh

PR #2075, opened 2026-09-08
Add beginner tutorials consuming the published milvus-sdk-java artifact

yhmo:
Introduce `tutorial/` as eight standalone Maven projects that depend on the **published** `milvus-sdk-java` from Maven Central instead of the local source tree, mirroring the C++/Rust SDK tutorial layout.

## Changes

- `tutorial/1_quickstart` — connect, create a collection, index, insert, search, clean up.
- `tutorial/2_collection` — full collection lifecycle (create/describe/properties/load/release/rename/truncate/drop).
- `tutorial/3_schema` — scalar, vector, and sparse/struct field data types.
- `tutorial/4_index` — vector and scalar index management (`createIndex`/`listIndexes`/`describeIndex`/`dropIndex`).
- `tutorial/5_dml` — insert, upsert (incl. partial update), delete by ids/filter, query.
- `tutorial/6_dql` — query, search, hybrid search, query/search iterators.
- `tutorial/7_database` — create, configure, select, and drop databases.
- `tutorial/8_rbac` — users, roles, privilege groups, and grants.
- Parent `tutorial/pom.xml` is the single source of the SDK version (`milvus.sdk.version`) and the `slf4j` runtime needed by the shaded artifact; each module is runnable standalone via `mvn exec:java`.
- `AGENTS.md`: document the tutorial layout and the release-time version bump.

Each tutorial is verified end-to-end against a Milvus standalone, and each ships its own `README.md` (What you learn / Prerequisites / Build and run).

DCO: commit carries a `Signed-off-by` trailer.

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -32,8 +32,10 @@ jobs:
           cache: maven
       - name: Update submodule
         run: git submodule update --init
-      - name: sdk-core unit tests (ut)
-        run: mvn -pl sdk-core -Pcoverage "-Djacoco.destFile=target/jacoco-ut-core.exec" "-Djacoco.reportDir=target/site/jacoco-ut-core" --batch-mode --update-snapshots verify
+      # BulkWriter must be tested on linux: several unit tests write Parquet files locally and
+      # require Hadoop's native WinUtils, which is only bundled for linux. On windows these tests
+      # fail with "HADOOP_HOME and hadoop.home.dir are unset" (see PR #2075), so keep this step on
+      # the ubuntu job.
       - name: BulkWriter unit tests (ut)
         run: mvn -pl sdk-bulkwriter -Pcoverage "-Djacoco.destFile=target/jacoco-ut-bulk.exec" "-Djacoco.reportDir=target/site/jacoco-ut-bulk" --batch-mode --update-snapshots verify
       - name: System tests (st)
@@ -41,7 +43,7 @@ jobs:
       - name: Upload coverage
         uses: codecov/codecov-action@v5
         with:
-          files: ./sdk-core/target/site/jacoco-ut-core/jacoco.xml,./sdk-core/target/site/jacoco-st/jacoco.xml,./sdk-bulkwriter/target/site/jacoco-ut-bulk/jacoco.xml
+          files: ./sdk-core/target/site/jacoco-st/jacoco.xml,./sdk-bulkwriter/target/site/jacoco-ut-bulk/jacoco.xml
           name: ubuntu-jacoco-coverage
           disable_safe_directory: true
           verbose: true
@@ -60,6 +62,8 @@ jobs:
           cache: maven
       - name: Update submodule
         run: git submodule update --init
+      - name: Tutorials compile (published SDK)
+        run: mvn -f tutorial/pom.xml compile --batch-mode
       - name: Unit tests (ut)
         run: mvn -pl sdk-core -Pcoverage "-Djacoco.destFile=target/jacoco-ut.exec" "-Djacoco.reportDir=target/site/jacoco-ut" --batch-mode --update-snapshots verify
       - name: Integration tests (it)
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -46,6 +46,36 @@ mvn exec:java -Dexec.mainClass="io.milvus.v2.SimpleExample"
   header comment of each class for its prerequisites.
 - Prefer `io.milvus.v2.*` for new example code, matching the V2-first development policy below.
 
+## Tutorials
+
+`tutorial/` is a set of beginner-oriented, standalone Maven projects for learning the SDK, mirroring the
+`tutorial/` layout of the C++ and Rust SDKs. Unlike `examples/`, each tutorial depends on the **published**
+`milvus-sdk-java` artifact (from Maven Central), not the local source tree, so the learner's experience
+matches an application developer installing the SDK.
+
+```text
+tutorial/
+├── pom.xml                    # parent POM; single source of the milvus-sdk-java version
+├── 1_quickstart/
+├── 2_collection/
+├── 3_schema/
+├── 4_index/
+├── 5_dml/
+├── 6_dql/
+├── 7_database/
+└── 8_rbac/
+```
+
+- `tutorial/pom.xml` is the aggregator/parent POM. The `milvus-sdk-java` version is declared once there
+  (e.g. `<milvus.sdk.version>3.0.9</milvus.sdk.version>` in `<dependencyManagement>`) and inherited by the
+  eight child modules — it must **not** be duplicated in each child `pom.xml`.
+- **When releasing a new SDK version, update the `milvus-sdk-java` version in `tutorial/pom.xml`
+  (`milvus.sdk.version`) so tutorials keep consuming the latest published release.**
+- Each child module is runnable standalone: `cd tutorial/1_quickstart && mvn exec:java`.
+- Connection settings use `MILVUS_URI` / `MILVUS_TOKEN` env vars, defaulting to
+  `http://localhost:19530` / `root:Milvus` (same convention as the C++/Rust SDK tutorials).
+- The source-linked examples used for SDK development and testing remain under `examples/`.
+
 ## SDK generations
 
 Two API generations coexist inside `sdk-core`:
@@ -88,13 +118,20 @@ mvn install                          # builds sdk-core + sdk-bulkwriter
 
 ## Testing
 
-Three layers under `sdk-core/src/test/java/io/milvus/`, tagged with JUnit 5 `@Tag`:
+Since the sdk-core test reorganization on `master` (#2070) the test tree is split into three layers
+under `sdk-core/src/test/java/io/milvus/`, tagged with JUnit 5 `@Tag`:
 
 | Layer | Tag | Depends on | Examples |
 |---|---|---|---|
 | Unit | `@Tag("unit")` | none | request/response DTO builders, utils, caches, enums, interceptors |
 | Integration | `@Tag("integration")` | mock gRPC stub (`support/v2/BaseTest` for V2, in-process `MockMilvusServer` for V1) | facade method forwarding, error mapping, ts/schema caches |
 | System | `@Tag("system")` | real Milvus standalone via Docker | end-to-end |
+
+Before the v3.0.9-era rework the tests lived in a single tree mirroring the `io.milvus.*` production
+packages (`v2/`, `client/`, `param/`, ...) with no type separation and no tag-driven
+filtering. The rework instead groups tests by layer (`unit/`, `integration/`,
+`system/`), then by SDK generation (`v1/`, `v2/`, `common/`), and makes each layer
+selectable via JUnit tags.
 
 Run a layer with:
 
@@ -119,7 +156,8 @@ mvn -pl sdk-core -Psystem test       # unit + integration + system (Docker)
 
 - `.github/workflows/maven.yml` splits into two jobs:
   - `Build and test on ubuntu` (`ubuntu-latest`): bulkwriter unit tests + system tests (`-Psystem`).
-  - `Build and test on windows` (`windows-2022`): sdk-core unit + integration tests (`-Pintegration`).
+  - `Build and test on windows` (`windows-2022`): tutorials compile + sdk-core unit + integration
+    tests (`-Pintegration`).
 - Coverage is collected with JaCoCo (`-Pcoverage`); each step writes a per-step exec/report via
   `-Djacoco.destFile` / `-Djacoco.reportDir`, and codecov uploads the ut/it/st reports.
 - The JaCoCo report excludes proto-generated packages (`io/milvus/grpc/**`, `milvus/proto/**`) so codecov
--- a/tutorial/1_quickstart/src/main/java/io/milvus/tutorial/quickstart/App.java
+++ b/tutorial/1_quickstart/src/main/java/io/milvus/tutorial/quickstart/App.java
@@ -1,0 +1,148 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package io.milvus.tutorial.quickstart;
+
+import com.google.gson.Gson;
+import com.google.gson.JsonObject;
+import io.milvus.v2.client.ConnectConfig;
+import io.milvus.v2.client.MilvusClientV2;
+import io.milvus.v2.common.DataType;
+import io.milvus.v2.common.IndexParam;
+import io.milvus.v2.service.collection.request.AddFieldReq;
+import io.milvus.v2.service.collection.request.CreateCollectionReq;
+import io.milvus.v2.service.collection.request.DropCollectionReq;
+import io.milvus.v2.service.collection.request.LoadCollectionReq;
+import io.milvus.v2.service.index.request.CreateIndexReq;
+import io.milvus.v2.service.vector.request.InsertReq;
+import io.milvus.v2.service.vector.request.SearchReq;
+import io.milvus.v2.service.vector.request.data.FloatVec;
+import io.milvus.v2.service.vector.response.InsertResp;
+import io.milvus.v2.service.vector.response.SearchResp;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
+/**
+ * Tutorial 1: Quick start.
+ *
+ * <p>The smallest complete MilvusClientV2 application. It connects to Milvus, creates a collection
+ * with an integer primary key, a text field, and a float vector, inserts two rows, performs a
+ * vector search, and removes the collection.
+ *
+ * <p>Connection defaults to {@code MILVUS_URI=http://localhost:19530} and
+ * {@code MILVUS_TOKEN=root:Milvus}. Override them with environment variables when needed.
+ */
+public class App {
+    public static void main(String[] args) {
+        String uri = System.getenv().getOrDefault("MILVUS_URI", "http://localhost:19530");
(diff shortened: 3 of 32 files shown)


tutorial/8_rbac/src/main/java/io/milvus/tutorial/rbac/App.java
```diff
@@ -0,0 +1,205 @@
...
+            String role = "java_tutorial_role_" + suffix;
+            String group = "java_tutorial_group_" + suffix;
+
+            // createUser creates a login identity. The password is its secret and the description
+            // is optional administrator-facing metadata.
+            client.createUser(CreateUserReq.builder()
+                    .userName(username)
+                    .password("JavaTutorial!123")
+                    .description("temporary Java SDK tutorial user")
+                    .build());
```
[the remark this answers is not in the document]
yhmo: Fixed: the password is now generated per run from the unique suffix ("JavaTutorial!" + suffix) instead of a hard-coded literal, since the tutorial never logs in as this user.

tutorial/2_collection/src/main/java/io/milvus/tutorial/collection/App.java
```diff
@@ -0,0 +1,204 @@
...
+ * inspect with {@code hasCollection} / {@code describeCollection}, change properties, load and
+ * release, get load state and row count, rename, truncate, and finally drop.
+ *
+ * <p>Connection defaults to {@code MILVUS_URI=http://localhost:19530} and
+ * {@code MILVUS_TOKEN=root:Milvus}. Override them with environment variables when needed.
+ */
+public class App {
+    private static final String TTL_SECONDS = "collection.ttl.seconds";
+
+    public static void main(String[] args) throws InterruptedException {
```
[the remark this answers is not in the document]
yhmo: Removed the unnecessary "throws InterruptedException" declaration. This tutorial only calls the no-arg close() and no interruptible APIs.

tutorial/6_dql/src/main/java/io/milvus/tutorial/dql/App.java
```diff
@@ -0,0 +1,316 @@
...
+        // searchIterator paginates a search while preserving one search session.
+        SearchIterator iterator = client.searchIterator(SearchIteratorReq.builder()
+                .collectionName(collectionName)
+                .vectorFieldName("dense")
+                .metricType(IndexParam.MetricType.COSINE)
+                .vectors(Arrays.asList(new FloatVec(denseVector(2))))
+                .outputFields(Arrays.asList("title", "category"))
+                .topK(3)
+                .batchSize(3)
+                .limit(7)
```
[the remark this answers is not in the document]
yhmo: Fixed: removed the deprecated topK(3) call and kept only limit(7) with batchSize(3).

tutorial/1_quickstart/src/main/java/io/milvus/tutorial/quickstart/App.java
```diff
@@ -0,0 +1,148 @@
...
+                .build();
+        MilvusClientV2 client = new MilvusClientV2(config);
+        try {
+            String collectionName = "tutorial_quickstart";
+
+            // Drop the collection if it already exists from a previous run.
+            client.dropCollection(DropCollectionReq.builder()
+                    .collectionName(collectionName)
+                    .build());
+
```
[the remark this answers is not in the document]
yhmo: Verified against Milvus standalone: dropCollection on a non-existent collection returns success (no error thrown), so the first run is safe and the tutorial stays rerunnable. Keeping the unconditional drop also serves as cleanup for any leftover from a previous interrupted run.

tutorial/7_database/src/main/java/io/milvus/tutorial/database/App.java
```diff
@@ -0,0 +1,123 @@
...
+ * databases, create a database, inspect it, set and remove a property, switch the client's
+ * selected database with {@code useDatabase}, and finally drop the tutorial database.
+ *
+ * <p>Connection defaults to {@code MILVUS_URI=http://localhost:19530} and
+ * {@code MILVUS_TOKEN=root:Milvus}. Override them with environment variables when needed.
+ */
+public class App {
+    private static final String REPLICA_NUMBER = "database.replica.number";
+
+    public static void main(String[] args) throws InterruptedException {
```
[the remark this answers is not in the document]
yhmo: This tutorial does call an interruptible API: client.useDatabase(String) declares "throws InterruptedException" (MilvusClientV2.java:555), and this tutorial invokes it twice (select the tutorial database, then switch back to default). The declaration is required and kept.

tutorial/5_dml/src/main/java/io/milvus/tutorial/dml/App.java
```diff
@@ -0,0 +1,201 @@
...
+            DeleteResp deleteById = client.delete(DeleteReq.builder()
+                    .collectionName(collectionName)
+                    .ids(Arrays.asList(1))
+                    .build());
+            System.out.printf("Deleted %d rows by primary key%n", deleteById.getDeleteCnt());
+
+            // delete with filter removes all entities matching the Milvus boolean expression.
+            DeleteResp deleteByFilter = client.delete(DeleteReq.builder()
+                    .collectionName(collectionName)
+                    .filter("id >= 4")
```
yhmo: tutorial/5_dml/src/main/java/io/milvus/tutorial/dml/App.java line:120
Medium ---- The "delete by filter" step never deletes anything. After the insert (ids 1,2), upsert (2,3), partial update of id 3, and `delete(ids=[1])`, only ids 2 and 3 remain, so `filter("id >= 4")` matches zero rows and the tutorial prints "Deleted 0 rows by filter". Since the surrounding comment states that this step removes all entities matching the boolean expression, consider using a filter that actually matches existing rows (e.g. `id >= 2`) so the tutorial demonstrates the delete-by-filter path instead of a no-op.

tutorial/6_dql/src/main/java/io/milvus/tutorial/dql/App.java
```diff
@@ -0,0 +1,316 @@
...
+import io.milvus.v2.service.vector.request.SearchReq;
+import io.milvus.v2.service.vector.request.data.FloatVec;
+import io.milvus.v2.service.vector.request.data.SparseFloatVec;
+import io.milvus.v2.service.vector.request.ranker.WeightedRanker;
+import io.milvus.v2.service.vector.response.QueryResp;
+import io.milvus.v2.service.vector.response.SearchResp;
+
+import com.google.gson.Gson;
+import com.google.gson.JsonObject;
+import io.milvus.v2.service.vector.request.data.BaseVector;
```
yhmo: tutorial/6_dql/src/main/java/io/milvus/tutorial/dql/App.java line:50
Low ---- `BaseVector` is imported but never referenced in this file; `FloatVec` and `SparseFloatVec` are used directly. The unused import can be removed to keep the example tidy.

tutorial/pom.xml
```diff
@@ -0,0 +1,76 @@
...
+    </licenses>
+
+    <properties>
+        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+        <maven.compiler.source>8</maven.compiler.source>
+        <maven.compiler.target>8</maven.compiler.target>
+        <!-- Single source of the milvus-sdk-java version used by all tutorials.
+             Keep in sync with the latest published release; do not duplicate it
+             in the child pom.xml files. -->
+        <milvus.sdk.version>3.0.9</milvus.sdk.version>
```
yhmo, line 30: tutorial/pom.xml line:30
Medium ---- The tutorial modules are not built by any CI workflow: the root reactor only includes `sdk-core` and `sdk-bulkwriter`, and no workflow references `tutorial/`. Nothing verifies the tutorials still compile against the pinned published artifact when `milvus.sdk.version` is bumped. Consider adding a CI job that runs e.g. `mvn -f tutorial/pom.xml compile` so a future SDK release that changes or removes an API used here is caught before users hit it.

tutorial/5_dml/src/main/java/io/milvus/tutorial/dml/App.java
yhmo: Fixed: the filter is now id >= 3 so it deletes the partially-updated id=3 row. Verified output shows 1 row deleted by filter, with the query afterward printing the remaining id=2 row.

tutorial/6_dql/src/main/java/io/milvus/tutorial/dql/App.java
yhmo: Fixed: removed the unused BaseVector import; FloatVec and SparseFloatVec are imported directly.

tutorial/pom.xml
yhmo: Fixed: added a Tutorials compile step to the ubuntu CI job that runs mvn -f tutorial/pom.xml compile --batch-mode against the pinned published SDK, so a future release that changes or removes a used API is caught before users hit it.

AGENTS.md
```diff
@@ -88,14 +118,21 @@ mvn install                          # builds sdk-core + sdk-bulkwriter
 
 ## Testing
 
-Three layers under `sdk-core/src/test/java/io/milvus/`, tagged with JUnit 5 `@Tag`:
+Since v3.0.9 the sdk-core test tree is reorganized into three layers under
```
yhmo: AGENTS.md line:121
Low ---- This claim is incorrect: the v3.0.9 tag (cc37bfd0, released 2026-09-03) still ships the old single-tree test layout under sdk-core/src/test/java/io/milvus/{client,param,orm,...}, while the unit/integration/system reorganization only exists on master (added by #2070, merged 2026-09-08). As written, release-note writers and maintainers will believe the test rework shipped in v3.0.9. Consider rewording to something like "Since the sdk-core test reorganization on master" or dropping the version reference.

tutorial/8_rbac/src/main/java/io/milvus/tutorial/rbac/App.java
```diff
@@ -0,0 +1,208 @@
...
+
+        ConnectConfig config = ConnectConfig.builder()
+                .uri(uri)
+                .token(token)
+                .build();
+        MilvusClientV2 client = new MilvusClientV2(config);
+        try {
+            // Milvus 2.6 defaults maxUsernameLength to 32 bytes. Keep the username compact while
+            // retaining the process ID and millisecond timestamp for uniqueness across runs.
+            String suffix = System.currentTimeMillis() + "_" + Thread.currentThread().getId();
```
yhmo, line 76: tutorial/8_rbac/src/main/java/io/milvus/tutorial/rbac/App.java line:77
Low ---- The comment says the suffix retains "the process ID", but the code uses Thread.currentThread().getId(), which is not a process ID. A fresh JVM started by `mvn exec:java` gives the main thread a small, usually identical id (often 1), so two tutorial runs launched in the same millisecond can produce the same username/role/group and collide on createUser -- the "uniqueness across runs" guarantee in the comment does not hold across processes. Either base the suffix on something process-unique (e.g. a UUID prefix) or fix the comment.

tutorial/3_schema/src/main/java/io/milvus/tutorial/schema/App.java
```diff
@@ -0,0 +1,253 @@
...
+import io.milvus.v2.client.MilvusClientV2;
+import io.milvus.v2.common.DataType;
+import io.milvus.v2.service.collection.request.AddFieldReq;
+import io.milvus.v2.service.collection.request.CreateCollectionReq;
+import io.milvus.v2.service.collection.request.DescribeCollectionReq;
+import io.milvus.v2.service.collection.request.DropCollectionReq;
+import io.milvus.v2.service.collection.response.DescribeCollectionResp;
+
+import java.util.Arrays;
+import java.util.List;
```
yhmo: tutorial/3_schema/src/main/java/io/milvus/tutorial/schema/App.java line:32
Low ---- java.util.List is imported but never used in this class; removing it keeps the tutorial code clean.

tutorial/8_rbac/src/main/java/io/milvus/tutorial/rbac/App.java
```diff
@@ -0,0 +1,208 @@
...
+import io.milvus.v2.service.rbac.request.ListPrivilegeGroupsReq;
+import io.milvus.v2.service.rbac.request.RemovePrivilegesFromGroupReq;
+import io.milvus.v2.service.rbac.request.RevokePrivilegeReqV2;
+import io.milvus.v2.service.rbac.request.RevokeRoleReq;
+import io.milvus.v2.service.rbac.response.DescribeRoleResp;
+import io.milvus.v2.service.rbac.response.DescribeUserResp;
+import io.milvus.v2.service.rbac.response.ListPrivilegeGroupsResp;
+
+import java.util.Arrays;
+import java.util.List;
```
yhmo: tutorial/8_rbac/src/main/java/io/milvus/tutorial/rbac/App.java line:45
Low ---- java.util.List is imported but never used in this class (collections come from Arrays.asList(...) and are only iterated); removing it keeps the tutorial code clean.

yhmo:
Reviewed at head b97db5af. Verified every SDK API used by the tutorials against the published 3.0.9 artifact (javap on the jar) and confirmed 3.0.9 is the latest published release; all tutorial modules compile against it, and the new CI "Tutorials compile" step will catch future API drift. The earlier copilot and review threads are all addressed or correctly rebutted. I left four low-severity, non-blocking polish notes inline: the AGENTS.md "Since v3.0.9" test-reorg claim is inaccurate (the unit/integration/system rework only exists on master, after v3.0.9 shipped), the RBAC suffix comment says "process ID" but uses the JVM thread id, and two unused java.util.List imports in 3_schema and 8_rbac.

tutorial/2_collection/src/main/java/io/milvus/tutorial/collection/App.java
```diff
@@ -0,0 +1,204 @@
...
+                    .build());
+            schema.addField(AddFieldReq.builder()
+                    .fieldName("embedding")
+                    .dataType(DataType.FloatVector)
+                    .dimension(4)
+                    .build());
+
+            // createCollection creates the collection with its schema, a vector index, and a
+            // default consistency level.
+            client.createCollection(CreateCollectionReq.builder()
```
yhmo, line 103: tutorial/2_collection/src/main/java/io/milvus/tutorial/collection/App.java line:94
Low ---- Tutorial 1 pre-drops its collection at startup to stay rerunnable after an interrupted run, but tutorials 2-7 do not. If a previous run left `tutorial_collection` / `tutorial_collection_renamed` behind (a crash or Ctrl-C during load), a rerun fails at createCollection with a confusing "collection already exists" error - exactly the scenario tutorial 1's pre-drop is designed to prevent. Consider pre-dropping the collection (and the renamed variant) at startup here and in tutorials 3-7, mirroring tutorial 1; dropCollection on a non-existent collection returns success, so no hasCollection guard is needed. Tutorial 8 already sidesteps this with timestamp-unique names.

AGENTS.md
yhmo: Fixed: reworded to reference the master test reorganization (PR 2070, merged 2026-09-08) and dropped the v3.0.9 version reference, since the v3.0.9 tag (2026-09-03) still ships the old layout.

tutorial/8_rbac/src/main/java/io/milvus/tutorial/rbac/App.java
yhmo: Fixed: the comment now says thread ID instead of process ID, matching Thread.currentThread().getId().

tutorial/3_schema/src/main/java/io/milvus/tutorial/schema/App.java
yhmo: Fixed: removed the unused java.util.List import.

tutorial/8_rbac/src/main/java/io/milvus/tutorial/rbac/App.java
yhmo: Fixed: removed the unused java.util.List import.

tutorial/2_collection/src/main/java/io/milvus/tutorial/collection/App.java
yhmo: Fixed: added startup pre-drops in tutorials 2-7 mirroring tutorial 1. Tutorial 2 drops both tutorial_collection and tutorial_collection_renamed (rename may have completed), tutorial 3 drops its three schema collections, tutorials 4-6 drop their single collection, and tutorial 7 drops the tutorial database. Verified that dropCollection and dropDatabase on non-existent resources both return success, and verified the interrupted-run scenario by seeding leftover collections before a rerun.

.github/workflows/maven.yml
```diff
@@ -60,6 +62,8 @@ jobs:
           cache: maven
       - name: Update submodule
         run: git submodule update --init
+      - name: Tutorials compile (published SDK)
```
yhmo, line 65: .github/workflows/maven.yml line:65
Low ---- The new Tutorials compile step validates that the tutorials still compile against the published SDK, but it does not exercise their runtime behavior. Several tutorial runtime bugs were found and fixed during this PR's review (for example the delete-by-filter step that originally deleted nothing and the deprecated iterator topK/limit conflict), and those fixes are currently protected only by manual end-to-end verification rather than by CI. Since the system-test job already starts a Milvus standalone on localhost:19530, consider also running the quickstart (or the other tutorials) against that instance so future SDK changes that break tutorial behavior at runtime are caught before users hit them. Non-blocking suggestion.

tutorial/7_database/src/main/java/io/milvus/tutorial/database/App.java
```diff
@@ -0,0 +1,129 @@
...
+                .uri(uri)
+                .token(token)
+                .build();
+        MilvusClientV2 client = new MilvusClientV2(config);
+        try {
+            String databaseName = "tutorial_database";
+
+            // Pre-drop any database left behind by an interrupted run; dropping a non-existent
+            // database succeeds, so this keeps the tutorial rerunnable.
+            client.dropDatabase(DropDatabaseReq.builder()
```
yhmo: tutorial/7_database/src/main/java/io/milvus/tutorial/database/App.java line:62
Medium ---- This new pre-drop assumes dropping a non-existent database succeeds, but on Milvus 2.6.0-2.6.5 (which the tutorial README explicitly supports: "Milvus 2.6 or later") DropDatabase on a missing database returns an error - the server's "drop a database that not found, ignore it" handling was only added in 2.6.6 (internal/rootcoord/root_coord.go), so the very first run of this tutorial aborts with a "database not found" failure on those versions. Collection pre-drops in tutorials 1-6 are unaffected because dropCollection has been idempotent since 2.6.0. Consider guarding the pre-drop (e.g. drop only when listDatabases() contains the name, or catch the not-found error) so the first run works on every 2.6.x.

yhmo: Fixed: the pre-drop now checks listDatabases() first and only drops when the tutorial database exists, so the first run works on Milvus 2.6.0-2.6.5 (where dropDatabase on a missing database errors) as well as 2.6.6+. Verified the tutorial still runs cleanly against a local standalone.

PR #2077, opened 2026-09-09
Auto release note

yhmo:


.github/workflows/publish-release.yaml
```diff
@@ -61,6 +61,8 @@ jobs:
             exit 1
           fi
 
+          ./scripts/extract_changelog.sh "${tag_version}" > /dev/null
```
yhmo, line 123: .github/workflows/publish-release.yaml line:64
Medium ---- This new changelog validation runs in the `publish` job before the Maven Central upload, so a tag whose CHANGELOG.md has no entry for the released version now blocks the entire artifact publish, not just the GitHub release note. If an urgent hotfix tag is pushed without a changelog update, no Maven artifacts can be released at all. Was blocking the Maven publish the intent, or should the missing-entry case only fail (or fall back in) the `release` job instead?

scripts/extract_changelog.sh
```diff
@@ -0,0 +1,138 @@
...
+      found = 0
+    }
+    {
+      if (substr($0, 1, 1) == "#") {
+        lvl = heading_level($0)
+        text = $0
+        sub(/^#+[[:space:]]*/, "", text)
+        sub(/^v/, "", text)
+        if (!found) {
+          if (text ~ pattern) {
```
yhmo, line 115: scripts/extract_changelog.sh line:115
Medium ---- The version token is matched against the text of any heading line, not specifically a version heading. If a changelog section ever gains a subheading that mentions the released version (e.g. `### Notes for 3.0.9 users` inside the 3.0.10 section, which sits above 3.0.9 in the file), the awk sets `found` at that subheading and emits the wrong section with exit 0, silently producing an incorrect release body. The current CHANGELOG.md only uses `### Feature`/`Improvement`/`Bug` subheadings, so this is latent today, but anchoring the match to a version-style heading would make the extraction robust.

.github/workflows/publish-release.yaml
yhmo: No change needed: this check mirrors the Rust SDK, whose publish job also fails before uploading when CHANGELOG.md lacks an entry for the released version. Blocking the immutable Maven Central upload is the intended fail-fast behavior so a release is never shipped without release notes.

scripts/extract_changelog.sh
yhmo: No change needed: the current CHANGELOG.md only uses ### Feature / ### Improvement / ### Bug subheadings, which never contain version numbers, and this script is a verbatim port of the Rust SDK's accepted script. The described scenario is latent; if a version-mentioning subheading is ever introduced, we can anchor the match to a version-style heading then.

PR #2079, opened 2026-09-10
Trim redundant content from published jar and make tests JDK 8 compatible

yhmo:
- Stop protobuf-maven-plugin from attaching .proto sources to the artifact (`attachProtoSources=false`), removing ~144KB of unused proto source text.
- Remove unused `log4j2.xml` and `milvus-client.properties` resources so they no longer ship in the published jar; the SDK only uses slf4j and reads the version from the jar manifest, and nothing in main or test code reads these files.
- Replace JDK 9+ APIs in sdk-core test sources (`Optional.orElseThrow()`, `List.of()`, `Map.of()`, `String.repeat()`) with JDK 8 equivalents so `mvn test` works with JDK 8 or later.
- Simplify RPC error logging in `RpcUtils` to varargs form (slf4j short-circuits formatting internally), log full exception stacks in `LocalBulkWriter`, and downgrade the per-file upload log in `VolumeFileManager` to debug to avoid INFO spam.
- Add a test-scope `slf4j-simple` binding to sdk-core and sdk-bulkwriter so guarded log lines execute under `mvn test` and patch coverage reflects them.


pom.xml
```diff
@@ -345,6 +345,7 @@
                 <version>${protobuf.maven.plugin.version}</version>
                 <configuration>
                     <protoSourceRoot>src/main/milvus-proto/proto</protoSourceRoot>
+                    <attachProtoSources>false</attachProtoSources>
```
yhmo, line 348: pom.xml line:348
Low ---- The protobuf-maven-plugin attachProtoSources default (true) exists so downstream Maven consumers can import the .proto definitions from this artifact and have them resolved automatically at build time (see the plugin compile-goal documentation). Turning it off is a reasonable jar trim, but it is a compatibility change for any consumer that imported protos from the published jar; confirming that no external consumer relies on that mechanism would fully close this out.

yhmo:
sdk-core/src/test/resources/milvus-client.properties line:3
Low ---- After the move, nothing in the test tree reads this file, and the root pom resource filtering (`<filtering>true</filtering>` on src/main/resources only) no longer resolves its `${project.groupId}/${project.artifactId}/${project.version}` placeholders, so they stay literal on the test classpath. Version is read from the jar manifest (`pkg.getImplementationVersion()`), so this is harmless today; consider deleting the dead file, or documenting that the placeholders are intentionally left unfiltered if it is meant to be used by tests.

yhmo:
sdk-core/src/test/resources/milvus-client.properties line:1
Low ---- This relocated file has no readers in either main or test code: a repo-wide search finds no reference to milvus-client.properties, and the SDK reports its version from the jar manifest (Package.getImplementationVersion(), enabled by addDefaultImplementationEntries in sdk-core/pom.xml) rather than from this file. One extra detail: main resources are filtered (filtering=true in the parent pom) but test resources are not, so the ${project.groupId}/${project.artifactId}/${project.version} placeholders now remain literal text. Since the PR goal is trimming redundant content, deleting this file (and log4j2.xml, which also has no reader given sdk-core pulls no log4j2 dependency) would be cleaner than relocating it; keeping it intentionally for a future test-time use is also fine.

sdk-core/src/main/java/io/milvus/v2/utils/RpcUtils.java
```diff
@@ -199,8 +201,10 @@ public <T> T retry(Callable<T> callable) {
                         || code == io.grpc.Status.ALREADY_EXISTS.getCode()
                         || code == io.grpc.Status.RESOURCE_EXHAUSTED.getCode()
                         || code == io.grpc.Status.UNIMPLEMENTED.getCode()) {
-                    String msg = String.format("Encounter rpc error that cannot be retried, reason: %s", e);
-                    logger.error(msg);
+                    if (logger.isErrorEnabled()) {
```
yhmo: sdk-core/src/main/java/io/milvus/v2/utils/RpcUtils.java line:204
Medium ---- These isErrorEnabled() guards skip the log statements whenever the level is disabled, and unit tests for both reactor modules run with no SLF4J binding (sdk-core and sdk-bulkwriter depend on slf4j-api only; sdk-bulkwriter explicitly excludes slf4j-reload4j), so isErrorEnabled() is false during `mvn test`. The non-retryable branch is exercised by RpcRetryTest.testRetryDoesNotRetryInvalidArgumentPreservesCause, but the newly guarded lines no longer execute, which is why codecov/patch now reports 0% patch coverage on RpcUtils and VolumeFileManager and is currently the only failing check (mergeable_state is blocked). Consider adding a test-scope SLF4J binding (e.g. slf4j-simple at debug level) so the guarded lines execute during tests and the coverage gate passes; also, the guard on the direct varargs call in handleResponse (line 151) is redundant because slf4j's varargs overload already short-circuits formatting internally.

sdk-core/src/main/resources/milvus-client.properties
```diff
@@ -1,3 +0,0 @@
-groupId=${project.groupId}
```
yhmo, line 1: sdk-core/src/main/resources/milvus-client.properties line:1
Low ---- The PR description says log4j2.xml and milvus-client.properties were "moved ... from main to test resources", but this diff deletes them entirely and neither file exists under sdk-core/src/test/resources at the PR head (the issue comments also reference sdk-core/src/test/resources/milvus-client.properties, which is a 404 path on this commit). If deletion is intended (consistent with the commit message, and fine since nothing reads these files), please update the PR description; if relocation for future test-time use was intended, the files are missing from test resources.

pom.xml
yhmo: Resolving: the underlying concern is addressed in the current head (codecov/patch now passes; PR description updated to reflect removal of the dead resource files).

sdk-core/src/main/java/io/milvus/v2/utils/RpcUtils.java
yhmo: Resolving: the underlying concern is addressed in the current head (codecov/patch now passes; PR description updated to reflect removal of the dead resource files).

sdk-core/src/main/resources/milvus-client.properties
yhmo: Resolving: the underlying concern is addressed in the current head (codecov/patch now passes; PR description updated to reflect removal of the dead resource files).

.github/workflows/publish-release.yaml
```diff
@@ -8,13 +8,13 @@ on:
...
-# was found" error.
+# CI check name(s) the PR that produced the tagged commit must have passed before
+# publishing. GitHub reports these on the PR object (statusCheckRollup); keep the
+# name in sync with what GitHub displays for the workflow in maven.yml (see gh pr
+# checks). Renaming the workflow job there must update the matching name here or
+# every release validation fails with a "No successful ... check was found" error.
 env:
-  UT_CHECK_UBUNTU: 'Build and test on ubuntu'
-  UT_CHECK_WINDOWS: 'Build and test on windows'
+  UT_CHECK: 'Build and test'
```
yhmo: .github/workflows/publish-release.yaml line:17
High ---- UT_CHECK ("Build and test") does not match the actual check run name, so this validation can never pass. GitHub records the Java CI check runs under the maven.yml JOB names -- "Build and test on ubuntu" and "Build and test on windows" (verified on the current PR head ed8931b7 via GET /commits/{sha}/check-runs, which returns exactly those job names). The jq filter `.name == $name` with "Build and test" therefore always yields zero matches, the step exits with "No successful Build and test check was found", and publishing is blocked on every release. Use the exact job name (e.g. "Build and test on ubuntu") or, as the previous code did, require one success from either of the two job names.

.github/workflows/publish-release.yaml
```diff
@@ -64,15 +67,35 @@ jobs:
...
-              echo "No successful ${check_name} check was found for ${TAG_SHA}." >&2
-              exit 1
-            fi
-          done
-          echo "Required post-merge checks passed for ${TAG_SHA}."
+          # CI checks run on the PR head commit (the repo squash-merges PRs, so the
+          # "Build and test" check is recorded on the PR head, not on the merged commit).
+          # Resolve the PR whose squash-merge produced the tagged commit, then verify the
+          # required check passed on that PR's head commit. Requires `checks: read`.
+          pr="$(gh api "repos/${GITHUB_REPOSITORY}/commits/${TAG_SHA}/pulls" \
```
yhmo, line 76: .github/workflows/publish-release.yaml line:74
Medium ---- GET /repos/{owner}/{repo}/commits/{sha}/pulls requires the Pull requests read permission for the GITHUB_TOKEN, but this job only declares `contents: read` and `checks: read` (lines 32-34). If the call returns 403, the `2>/dev/null || true` swallows the error, `pr` becomes empty, and the step fails with "No pull request found for commit ${TAG_SHA}", so the release can never be validated even after the check-name below is fixed. Add `pull-requests: read` to the job-level permissions block so the PR lookup can succeed.

.github/workflows/publish-release.yaml
yhmo: Fixed in c58ccc9c: check names restored to the actual maven.yml job names (Build and test on ubuntu / Build and test on windows); validate job now also declares pull-requests: read for the PR lookup.

.github/workflows/publish-release.yaml
yhmo: Fixed in c58ccc9c: check names restored to the actual maven.yml job names (Build and test on ubuntu / Build and test on windows); validate job now also declares pull-requests: read for the PR lookup.

sdk-core/src/test/java/io/milvus/system/v1/client/MultiClientAsyncTest.java
```diff
@@ -214,7 +217,12 @@ void testAsyncMethods() {
...
                 FieldDataWrapper wrapper = queryResultsWrapper.getFieldWrapper(fieldName);
                 System.out.println("Query data of " + fieldName + ", row count: " + wrapper.getRowCount());
-//                System.out.println(wrapper.getFieldData());
+                if (field1Name.equals(fieldName)) {
+                    System.out.println("distinctQueryIDs: " + distinctQueryIDs);
+                    System.out.println("query returned " + fieldName + " ids: " + wrapper.getFieldData());
+                }
+                // Strong consistency guarantees reads see the latest committed writes
+                // (guarantee ts = current max ts), so all flushed rows must be visible here.
                 assertEquals(distinctQueryIDs.size(), wrapper.getFieldData().size());
```
yhmo: High ---- The "Build and test on ubuntu" check is red at the current head (9a651ed2): MultiClientAsyncTest.testAsyncMethods fails at line 238 with expected:<10> but was:<7>. Even with this new 30s polling loop, the query `long_field in distinctQueryIDs` returns only 7 of the 10 distinct ids (the three 46900169883045xxxx ids from the first insert batches never appear in the result), so the poll times out and the count assertion still fails. Re-running the same query cannot help when the requested ids are not actually queryable in the collection — the multi-primary auto-id overlap described in the comment above means `wrapper.getLongIDs().get(0)` can return ids that are not the real stored primary keys. Since this is exactly the flaky assertion the PR intends to stabilize, please investigate the root cause (e.g., assert against ids actually read back from the collection) and confirm the ubuntu system-test step passes before merge.

yhmo: Resolved: this comment targeted the polling-based fix on 9a651ed2. The fix was replaced by Strong consistency (SearchParam + QueryParam .withConsistencyLevel(STRONG)), and the current head ce26562b passes the ubuntu system-test step (MultiClientAsyncTest now reads all rows on the first query). Note the 'multi-primary auto-id overlap' theory is not the root cause: Milvus allocates auto ids server-side and they are globally unique (insert returns the stored primary keys via AllocAutoID). The misleading test comment was corrected accordingly.

PR #2081, opened 2026-09-14
Build telemetry snapshots outside the collectors lock

yhmo:
Build telemetry snapshots outside the collectors lock

## Performance

Swap the accumulated collectors map out of the heartbeat lock in O(1) and build the per-bucket snapshot (copy + sort of up to 1000 samples per collection) outside the lock, so `createSnapshot` no longer blocks `recordOperation` on the hot path with many collections. The window boundary is captured at the swap instant so concurrent records start the next window. A dedicated `collectorsLock` monitor replaces locking on the mutable map reference so the swap stays race-free.

Benchmark (Java 8, 12 CPU, telemetry enabled, sampling 1.0): with 7 operations x 1000 collections x 1000 samples/bucket, the maximum time a concurrent `recordOperation` is blocked during `createSnapshot` drops from ~28-68ms to ~0.1ms (3000 collections: ~115ms to sub-ms).

## Telemetry hardening (following milvus-sdk-cpp#601)

- Reset the UNIMPLEMENTED backoff when a transport is attached or on runtime-state handoff, so a new endpoint is probed immediately instead of staying silent for up to thirty minutes.
- Truncate `show_errors` replies with UTF-8-safe prefixes across all truncatable fields instead of splitting a UTF-16 `error_msg` mid-surrogate-pair.
- Re-queue a command's original reply on duplicate redelivery so an already-reported failure is not contradicted by a success ACK.
- Cap the reported latency-history window start at the retained range so a maximum pushed interval cannot underflow the timestamp.
- Canonicalize ACK command IDs returned by custom handlers (the incoming command ID is the protocol correlation key).
- Per-operation try/catch in the snapshot loop so one failing collector does not discard the whole window or take down the heartbeat scheduling loop.

## Tests

- `TelemetrySnapshotSwapTest`: sequential separate-window semantics + a deterministic concurrency test that observes the collectors-map swap while `createSnapshot` is in flight.
- `TelemetryHardeningTest`: handoff backoff reset, UTF-8-safe truncation branches, longest-field truncation, duplicate redelivery replay (failed reply + null-reply fallback).

## Test infra

Bump the test docker-compose MinIO images to `quay.io/minio/minio:RELEASE.2025-09-07T16-13-09Z`. The Docker Hub `minio/minio` tags are no longer anonymously pullable in CI (`pull access denied`), which broke the "Deploy milvus server,build and test" job.


sdk-core/src/main/java/io/milvus/telemetry/ClientTelemetryManager.java
```diff
@@ -748,13 +749,20 @@ private void createSnapshot() {
             return;
         }
         CollectionScope collectionScope = snapshotCollectionScope();
+        // Swap the accumulated collectors out in O(1) so the per-bucket snapshot work
+        // below (copy + sort of up to 1000 samples per collection) runs without holding
+        // the lock. Concurrent recordOperation calls meanwhile start a fresh next-window
+        // bucket, so heartbeat work no longer blocks the hot path with many collections.
+        Map<String, Collector> working;
```
yhmo, line 773: sdk-core/src/main/java/io/milvus/telemetry/ClientTelemetryManager.java line:756
Low ---- This PR is a performance change (recordOperation is no longer blocked while createSnapshot copies and sorts up to 1000 samples per collection), and CODE_REVIEW.md asks that a performance PR list a benchmark result. Please add measured before/after evidence (for example recordOperation latency or heartbeat lock-hold time with many collections) so the claimed hot-path win is verifiable.

sdk-core/src/main/java/io/milvus/telemetry/ClientTelemetryManager.java
```diff
@@ -748,13 +749,20 @@ private void createSnapshot() {
...
-                OperationSnapshot snapshot = entry.getValue().snapshot(entry.getKey());
-                if (snapshot != null) {
-                    metrics.add(filterCollectionMetrics(snapshot, collectionScope));
-                }
+        for (Map.Entry<String, Collector> entry : working.entrySet()) {
+            OperationSnapshot snapshot = entry.getValue().snapshot(entry.getKey());
+            if (snapshot != null) {
+                metrics.add(filterCollectionMetrics(snapshot, collectionScope));
             }
         }
```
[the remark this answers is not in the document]
yhmo: Fixed: the window boundary is now captured inside the swap critical section, so the fresh bucket's next window starts at the swap instant rather than after the per-bucket sort loop. Operations recorded into the fresh map while the snapshot is being built therefore fall within the next window that begins when they actually occurred.

sdk-core/src/main/java/io/milvus/telemetry/ClientTelemetryManager.java
```diff
@@ -748,13 +749,20 @@ private void createSnapshot() {
...
         CollectionScope collectionScope = snapshotCollectionScope();
+        // Swap the accumulated collectors out in O(1) so the per-bucket snapshot work
+        // below (copy + sort of up to 1000 samples per collection) runs without holding
+        // the lock. Concurrent recordOperation calls meanwhile start a fresh next-window
+        // bucket, so heartbeat work no longer blocks the hot path with many collections.
+        Map<String, Collector> working;
+        synchronized (collectorsLock) {
+            working = collectors;
+            collectors = new HashMap<>();
+        }
```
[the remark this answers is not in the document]
yhmo: Fixed: added `sdk-core/src/test/java/io/milvus/telemetry/TelemetrySnapshotSwapTest` with two unit tests:

- `sequentialSnapshotsReportSeparateWindows`: records split across two snapshots, and the window boundary is contiguous (first `end_time` == next `timestamp`).
- `recordsDuringSnapshotSwapStartFreshWindow`: a deterministic concurrency test that observes the collectors-map swap (under the collectors lock) while `createSnapshot` is in flight, records into the fresh bucket, then asserts the record is absent from the in-flight snapshot and present in the next one.

sdk-core/src/main/java/io/milvus/telemetry/ClientTelemetryManager.java
yhmo: Benchmark (Java 8, 12 CPU, in-process manager, telemetry enabled, sampling 1.0). The key metric is the maximum time a concurrent `recordOperation` is blocked while `createSnapshot` runs, measured by a spinning record thread during an in-flight snapshot:

| scenario | before | after |
|---|---|---|
| 1 op x 1000 collections, 100 samples/bucket | 2.9-7.0 ms | 0.04-0.07 ms |
| 1 op x 1000 collections, 1000 samples/bucket | 7.5-12.8 ms | 0.04-0.05 ms |
| 7 ops x 1000 collections, 1000 samples/bucket | 28-68 ms | ~0.11 ms |
| 7 ops x 3000 collections, 1000 samples/bucket | ~115 ms | ~0.1 ms |

The snapshot sort work itself still takes the same wall time (tens of ms at 1000 collections), but it now runs on the heartbeat thread without holding the `collectors` lock, so business operations are no longer blocked. ~250-600x reduction in hot-path blocking.

sdk-core/src/main/java/io/milvus/telemetry/ClientTelemetryManager.java
```diff
@@ -532,14 +544,20 @@ public synchronized void processCommands(List<ClientCommand> commands) {
                 continue;
             }
             synchronized (executedCommands) {
-                if (executedCommands.containsKey(command.getCommandId())) {
-                    queueReply(successReply(command.getCommandId(), ByteString.EMPTY));
+                ExecutedCommand executed = executedCommands.get(command.getCommandId());
+                if (executed != null) {
+                    // Re-queue the command's actual reply so a redelivery cannot turn an
```
yhmo, line 549: sdk-core/src/main/java/io/milvus/telemetry/ClientTelemetryManager.java line:549
Medium ---- The redelivery path now replays the previously stored CommandReply instead of a success ACK. TelemetryHardeningTest now covers the show_errors truncation, the truncatedFieldValue branches and the handoff backoff reset, but the reply-replay behavior at this hunk is still not directly tested: a command whose first execution produced a failure should, on redelivery, re-queue that stored failure reply rather than a success ACK. The commandId canonicalization in handleCommand also has no direct test. The codecov/patch check is still failing on this PR (67% patch coverage). Could you add a test that executes a command once, then processes the same command ID again, and asserts the stored reply (e.g., a failure) is re-queued exactly once more?

sdk-core/src/main/java/io/milvus/telemetry/ClientTelemetryManager.java
```diff
@@ -748,18 +766,32 @@ private void createSnapshot() {
             return;
         }
         CollectionScope collectionScope = snapshotCollectionScope();
+        // Swap the accumulated collectors out in O(1) so the per-bucket snapshot work
+        // below (copy + sort of up to 1000 samples per collection) runs without holding
+        // the lock. Concurrent recordOperation calls meanwhile start a fresh next-window
```
yhmo, line 771: sdk-core/src/main/java/io/milvus/telemetry/ClientTelemetryManager.java line:771
Low ---- This PR title and description only describe the snapshot-outside-the-lock performance change, but the diff also bundles the milvus-sdk-cpp#601 telemetry-hardening changes (executedCommands reply replay, ACK commandId canonicalization, show_errors multi-field truncation, unsupportedStreak reset, window-start clamp) plus unrelated minio image bumps in four docker-compose files. The description is also stale: it states "no test cases were written for this change yet" although TelemetrySnapshotSwapTest and TelemetryHardeningTest are added, and the benchmark result is only posted in a review thread rather than in the description. Per CODE_REVIEW.md the code should "do exactly the same as title and commit message describe". Consider moving the docker-compose changes to a separate PR and expanding the description so the reviewable scope matches the diff.

sdk-core/src/main/java/io/milvus/telemetry/ClientTelemetryManager.java
```diff
@@ -532,14 +544,20 @@ public synchronized void processCommands(List<ClientCommand> commands) {
...
+                    queueReply(executed.reply == null
+                            ? successReply(command.getCommandId(), ByteString.EMPTY)
+                            : executed.reply);
                     continue;
                 }
             }
             CommandReply reply = handleCommand(command);
             synchronized (executedCommands) {
-                executedCommands.put(command.getCommandId(), command.getCreateTime());
+                executedCommands.put(command.getCommandId(),
```
yhmo, line 559: sdk-core/src/main/java/io/milvus/telemetry/ClientTelemetryManager.java line:559
Low ---- The dedupe map now stores the complete CommandReply (previously just a Long timestamp) for every executed command at the current cursor, until a batch with a newer createTime arrives. Replies can carry large payloads (e.g., show_errors / latency-history responses up to the 1 MB reply limit), and entries are only purged once a later timestamp is seen; if command timestamps stall, distinct command IDs at the cursor accumulate with their full replies retained. Consider storing only the reply outcome (success/error flag + message, or a small serialized summary) in ExecutedCommand instead of the whole protobuf, so redelivery replay does not pin potentially large payloads in memory.

sdk-core/src/main/java/io/milvus/telemetry/ClientTelemetryManager.java
```diff
@@ -748,18 +766,32 @@ private void createSnapshot() {
...
+            now = System.currentTimeMillis();
+        }
         List<OperationSnapshot> metrics = new ArrayList<>();
-        synchronized (collectors) {
-            for (Map.Entry<String, Collector> entry : collectors.entrySet()) {
-                OperationSnapshot snapshot = entry.getValue().snapshot(entry.getKey());
-                if (snapshot != null) {
-                    metrics.add(filterCollectionMetrics(snapshot, collectionScope));
-                }
+        for (Map.Entry<String, Collector> entry : working.entrySet()) {
```
yhmo, line 784: sdk-core/src/main/java/io/milvus/telemetry/ClientTelemetryManager.java line:784
Low ---- With the swap-before-process design, any exception thrown inside the per-bucket loop (snapshot() copy/sort or filterCollectionMetrics) now silently discards the entire window: the working map was already detached under collectorsLock, so the accumulated collectors are unreachable and will be GC’d, and lastSnapshotEnd is not advanced. Previously the loop ran under the collectors monitor on the live map, so collectors that had not yet been processed survived for the next heartbeat. Consider wrapping the loop so a failure re-attaches or at least logs the lost window.

sdk-core/src/test/resources/docker/docker-compose.yml
```diff
@@ -18,7 +18,7 @@ services:
 
   minio:
     container_name: milvus-javasdk-minio
-    image: minio/minio:RELEASE.2024-12-18T13-15-44Z
+    image: quay.io/minio/minio:RELEASE.2025-09-07T16-13-09Z
```
yhmo, line 21: sdk-core/src/test/resources/docker/docker-compose.yml line:21
Low ---- These MinIO image bumps (Docker Hub minio/minio to quay.io/minio/minio, plus a large version jump across 4 compose files, including a 2022-03-17 → 2025-09-07 change in tests/milvustest) are unrelated to the telemetry snapshot change and are not mentioned in the PR description. If they are required for CI (e.g. the pinned Milvus image needs a newer MinIO), please call that out or split them into a separate PR so the perf change is easy to bisect.

sdk-core/src/main/java/io/milvus/telemetry/ClientTelemetryManager.java
```diff
@@ -828,7 +860,11 @@ private synchronized CommandReply handleCommand(ClientCommand command) {
             return failedReply(command.getCommandId(), "unknown command type: " + command.getCommandType());
         }
         try {
-            return handler.handle(command);
+            CommandReply reply = handler.handle(command);
+            // The incoming command ID is the protocol correlation key. Extension handlers
+            // control the result, but cannot redirect or drop its ACK.
+            return reply == null ? null
+                    : reply.toBuilder().setCommandId(command.getCommandId()).build();
```
yhmo: sdk-core/src/main/java/io/milvus/telemetry/ClientTelemetryManager.java line:867
Low ---- This now overwrites the reply’s commandId with the incoming command’s ID for every handler, including extension handlers registered through registerCommandHandler. Default handlers already set it, so they are unaffected, but a custom CommandHandler that deliberately correlated its reply to a different command ID would have it silently rewritten. The intent is documented, but it changes the extension-handler contract without being mentioned in the PR description — worth noting for anyone with a custom handler.

sdk-core/src/main/java/io/milvus/telemetry/ClientTelemetryManager.java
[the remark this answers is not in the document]
yhmo: Fixed: added two redelivery tests to TelemetryHardeningTest that exercise the reply-replay path directly:

- duplicateCommandRedeliveryRequeuesOriginalFailedReply: an unknown command type produces a failed reply; redelivering the same ID re-queues the original failed reply (same commandId + error message), not a success ACK.
- duplicateNullReplyRedeliveryFallsBackToSuccessAck: a custom handler returning null stores no reply; on redelivery the fallback queues a success empty ACK.

sdk-core/src/main/java/io/milvus/telemetry/ClientTelemetryManager.java
yhmo: Fixed: added two redelivery tests to TelemetryHardeningTest that exercise the reply-replay path directly:

- duplicateCommandRedeliveryRequeuesOriginalFailedReply: an unknown command type produces a failed reply; redelivering the same ID re-queues the original failed reply (same commandId + error message), not a success ACK.
- duplicateNullReplyRedeliveryFallsBackToSuccessAck: a custom handler returning null stores no reply; on redelivery the fallback queues a success empty ACK.

sdk-core/src/main/java/io/milvus/telemetry/ClientTelemetryManager.java
yhmo: Fixed: the PR description now covers the full scope - the snapshot-outside-the-lock performance change with benchmark, the milvus-sdk-cpp#601 hardening changes, the new tests, and the MinIO test-infra bump.

sdk-core/src/main/java/io/milvus/telemetry/ClientTelemetryManager.java
yhmo: The retention is bounded by the cursor window: processCommands keeps only entries whose createTime equals the current batch's maxTimestamp and removes older ones at the end of each batch, so the map holds only the latest batch's commands. Per-heartbeat control-plane traffic is small, and each stored reply is already bounded by MAX_REPLY_BYTES (1 MB). This matches the milvus-sdk-cpp#601 approach (it also stores the full reply for replay). Keeping the full reply is what lets a redelivery replay the exact original failure instead of a generic ACK; only a pathological server pushing a huge batch at the same cursor would grow it, bounded to that window.

sdk-core/src/main/java/io/milvus/telemetry/ClientTelemetryManager.java
yhmo: Fixed: the per-bucket loop is now wrapped in a try/catch per operation (logs at debug and continues), so one failing collector no longer discards the whole window, and an exception can no longer escape createSnapshot and take down the heartbeat scheduling loop.

sdk-core/src/test/resources/docker/docker-compose.yml
yhmo: The MinIO bump is required for CI and documented in the PR description: the Docker Hub minio/minio tags used by the test compose files - including tests/milvustestv2, which the 'Deploy milvus server,build and test' job uses - return 'pull access denied' for anonymous pulls, which broke that job. The bump to quay.io/minio/minio:RELEASE.2025-09-07T16-13-09Z was verified pullable and was requested to be included in this commit.

sdk-core/src/main/java/io/milvus/telemetry/ClientTelemetryManager.java
yhmo: Intentional, matching milvus-sdk-cpp#601: the incoming command ID is the protocol correlation key, so extension handlers control the reply's result but cannot redirect or drop its ACK. Built-in handlers already set the correct ID, so this is a no-op for them.

sdk-core/src/main/java/io/milvus/telemetry/ClientTelemetryManager.java
```diff
@@ -173,6 +177,10 @@ public ClientTelemetryManager(
 
     public void setStub(ClientTelemetryServiceGrpc.ClientTelemetryServiceBlockingStub stub) {
         this.stub = stub;
+        // A newly attached transport must be probed immediately. Carrying an
+        // UNIMPLEMENTED backoff from the retired endpoint can otherwise keep a
+        // supporting endpoint silent for up to thirty minutes.
+        unsupportedStreak = 0;
```
yhmo, line 183: sdk-core/src/main/java/io/milvus/telemetry/ClientTelemetryManager.java line:183
Low ---- The immediate-probe reset of `unsupportedStreak` here is the other half of the UNIMPLEMENTED-backoff hardening (besides `handoffRuntimeStateTo`, which has a test). `setStub` itself is untested, and it is the only thing that neutralizes the streak carried in by `restoreRuntimeState` on the V2 reconnect path (MilvusClientV2.connect restores state, then calls setStub). A small unit test that sets `unsupportedStreak`, calls `setStub`, and asserts `isSupported()` would pin this behavior and guard against a future reorder of the connect sequence.

sdk-core/src/main/java/io/milvus/telemetry/ClientTelemetryManager.java
```diff
@@ -748,18 +766,38 @@ private void createSnapshot() {
...
+                // window, and must not take down the heartbeat scheduling loop.
+                logger.debug("Client telemetry snapshot skipped an operation", exception);
             }
         }
-        long now = System.currentTimeMillis();
         long start = lastSnapshotEnd == 0 || lastSnapshotEnd > now
                 ? now - config.getHeartbeatIntervalMs() : lastSnapshotEnd;
+        // Cap the reported window start at the retained-history range so a maximum
+        // pushed interval cannot surface an extreme (underflowed) timestamp.
+        start = Math.max(start, now - SNAPSHOT_HISTORY_TTL_MS);
```
yhmo, line 800: sdk-core/src/main/java/io/milvus/telemetry/ClientTelemetryManager.java line:800
Low ---- This clamp is a new, user-visible behavior (for heartbeat intervals longer than one hour, the reported latency-history window start is capped at the 1h retention range), but it is not directly tested. The new snapshot tests cover window contiguity but not the huge-interval path. A regression test such as: set a multi-hour `heartbeatIntervalMs` (or push `lastSnapshotEnd` backwards), invoke `createSnapshot`, then assert the snapshot timestamp is within `now - SNAPSHOT_HISTORY_TTL_MS` would protect the hardening described in the PR body.

sdk-core/src/main/java/io/milvus/telemetry/ClientTelemetryManager.java
yhmo: Fixed: added setStubClearsUnsupportedBackoff to TelemetryHardeningTest - it sets unsupportedStreak=3, calls setStub, and asserts isSupported() becomes true, covering the V2 reconnect path (restoreRuntimeState carries the streak, setStub neutralizes it).

sdk-core/src/main/java/io/milvus/telemetry/ClientTelemetryManager.java
yhmo: Fixed: added snapshotWindowStartIsCappedAtRetentionRange to TelemetryHardeningTest - it configures a 3-hour heartbeat interval, invokes createSnapshot, and asserts the reported window start is capped at end_time - 1h instead of underflowing.

sdk-core/src/test/java/io/milvus/system/v1/client/MultiClientAsyncTest.java
```diff
@@ -184,6 +187,7 @@ void testAsyncMethods() {
 
         // query async
         String expr = field1Name + " in " + distinctQueryIDs;
+        System.out.println("query expr: " + expr);
```
yhmo: sdk-core/src/test/java/io/milvus/system/v1/client/MultiClientAsyncTest.java line:190
Low ---- These two System.out.println calls added by this PR (line 190 and the "field data size" print at lines 230-231) are leftover debug output and only add noise to CI logs. Please remove them (or move the diagnostics behind a logger) before merge.

yhmo: Fixed: removed both debug printlns. The retained assertion (distinctQueryIDs.size() == futureResponses.size()) plus the pre-existing row-count/distinct-ID/full-ID prints already cover the diagnostics.

PR #2090, opened 2026-09-17
feat: add scalable Volume upload policies and credential recovery

lentitude2tk:
## Summary

Volume uploads currently overwrite every local file and retain the entire file list and scheduled tasks in memory. This change adds LIST-based upload policies and bounded planning/scheduling for large directories.

- Add `OVERWRITE`, `SKIP_IF_EXISTS`, `SKIP_IF_SAME_SIZE` (default), and `SIZE_AND_MTIME`. Matching uses the exact target key and nullable LIST metadata; no object HEAD/GET permission is required. List the requested target prefix in pages of up to 1,000 objects and stop when all local keys have been found.
- Keep small manifests in memory and migrate to a temporary SQLite index above 10,000 files, 10,000 queued directories, or an 8 MiB memory estimate. Add `sqlite-jdbc`; stream remaining entries through a bounded upload executor and clean up request-local resources on completion, failure, or cancellation.
- Add preparation logs, throttle progress callbacks and invoke them outside the tracker lock, validate local file stability, and refresh expired credentials for `ExpiredToken` / `SecurityTokenExpired` using the existing bounded retry loop. Page retries preserve the continuation token; PUT retries honor the selected policy.
- Add opt-in live tests and demonstrate the default upload policy while preserving the existing example settings. No credentials, private endpoints, generated reports, or unrelated workspace changes are included.

### Behavior changes

The default now skips existing objects with the same size; callers wanting the previous overwrite behavior must choose `OVERWRITE`. `VolumeBulkWriter` explicitly uses `OVERWRITE` for its UUID-scoped chunks, preserving its previous behavior and avoiding per-chunk LIST calls. Size/time comparisons are heuristics, not content checksums. Symbolic links to regular files and directories retain their logical upload paths. Directory cycles, broken links, and special files fail during planning; preparation errors surface through the returned future. Preparation progress is logged; `ProgressListener` remains upload-only.

## Validation

- All 189 BulkWriter tests passed, including callbacks outside the tracker lock, BulkWriter overwrite without LIST, symbolic file/directory uploads, alias preservation in memory and SQLite, cycle detection, credential-expiration recovery, policy matching, pagination, cancellation, and cleanup.
- The million-record scale test passed in a separate JVM with `-Xmx64m`.
- Compiled the updated upload example with Java 8 source/target settings.
- Prior live Volume validation passed 16 scenarios using 0–16 byte files and local sets up to 10,001 files. Repeating the full 10,001-file upload made 11 LIST calls and zero PUT calls; the 10,000/10,001 memory/SQLite boundary was verified. This did not exercise large-file multipart throughput or real credential expiration.

- CI regression verification: `BoundedUploadExecutorTest` passed on JDK 11 with coverage on all three branches. The corrected cancellation test also passed 5,000 consecutive runs with JaCoCo enabled; its cleanup covers interruption during the initial barrier.


sdk-bulkwriter/src/main/java/io/milvus/bulkwriter/VolumeFileManager.java
```diff
@@ -559,59 +595,38 @@ synchronized void resetFile(String filePath) {
...
-                emitProgress(progress);
-            }
+        synchronized void updateFile(String filePath, long fileSize, long chunkBytes) {
+            if (chunkBytes <= 0) { return; }
+            long previous = fileProgress.getOrDefault(filePath, 0L);
+            long current = Math.min(fileSize, previous + chunkBytes);
+            if (current <= previous) { return; }
+            fileProgress.put(filePath, current);
+            uploadedBytes += current - previous;
+            emitIfNeeded(filePath, current, fileSize);
```
yhmo: sdk-bulkwriter/src/main/java/io/milvus/bulkwriter/VolumeFileManager.java line:605
Medium ---- The user-supplied ProgressListener (and the INFO log inside emitProgress) is now invoked while holding the UploadProgressTracker monitor: updateFile/finishFile/finishUpload are all synchronized and call emitIfNeeded -> emitProgress inside the lock. The previous code deliberately emitted progress outside the lock. A listener that re-enters the tracker would deadlock, and any blocking listener now serializes all concurrent upload workers on this monitor. Consider building the snapshot under the lock and invoking the listener (and logging) outside of it.

sdk-bulkwriter/src/main/java/io/milvus/bulkwriter/VolumeFileManager.java
```diff
@@ -89,87 +89,92 @@ public VolumeFileManager(VolumeFileManagerParam volumeFileManagerParam) {
...
+                                         int concurrency, int maxRetries, long retryInterval, long partSize,
+                                         UploadPolicy uploadPolicy, UploadFilesRequest.ProgressListener listener) throws Exception {
+        logger.info("Planning volume upload: sourcePath:{}, volumePath:{}, concurrency:{}, uploadPolicy:{}",
+                source, volumePath, concurrency, uploadPolicy);
+        try (UploadManifest manifest = new UploadManifest(temporaryDirectory)) {
+            boolean singleFile = manifest.scan(source);
+            UploadManifest.checkInterrupted();
+            UploadContext context = new UploadContext(refreshVolumeAndClient(volumePath));
+            try {
+                String targetPrefix = context.currentSession().applyVolumeResponse.getVolumePrefix() + volumePath;
```
yhmo, line 140: sdk-bulkwriter/src/main/java/io/milvus/bulkwriter/VolumeFileManager.java line:141
Low ---- targetPrefix (and the listing prefix derived from it for filterExistingFiles) is computed once from the session captured before the existence-check/upload loop, while the bucket is read from the current (possibly refreshed) session on every page and PUT attempt. If a credential refresh ever rotates the volume prefix or bucket, the listing prefix, key matching, and PUT object keys would be based on a stale prefix against the new bucket. Consider re-deriving the prefix from the current session per page/attempt, or documenting why the prefix is guaranteed to be stable across refreshes.

sdk-bulkwriter/src/main/java/io/milvus/bulkwriter/request/volume/UploadFilesRequest.java
```diff
@@ -73,6 +89,17 @@ public class UploadFilesRequest {
...
+    /**
+     * Upload selection policy. Defaults to SKIP_IF_SAME_SIZE. All comparisons use the exact
+     * target key, including its relative directory path. Non-overwrite policies use paginated
+     * LIST metadata, without HEAD/GET object access. SIZE_AND_MTIME compares local modification
+     * time with the server object's LastModified time, not a preserved source timestamp.
+     * Failed PUT attempts are rechecked using the same policy. Skipped files are excluded from
+     * planned upload limits and progress totals. None of these policies guarantees content
+     * equality or atomic create-if-absent behavior against concurrent writers.
+     */
+    private UploadPolicy uploadPolicy = UploadPolicy.SKIP_IF_SAME_SIZE;
```
yhmo, line 103: sdk-bulkwriter/src/main/java/io/milvus/bulkwriter/request/volume/UploadFilesRequest.java line:101
Low ---- VolumeBulkWriter builds its internal UploadFilesRequest without an uploadPolicy and therefore silently inherits this new default (SKIP_IF_SAME_SIZE), whereas the previous behavior was unconditional overwrite. VolumeBulkWriter also deletes local chunk files after a successful upload; if a naming change ever reuses a chunk key with identical size but different content, the update would be silently skipped and the stale remote object retained, with no way for VolumeBulkWriter callers to opt into OVERWRITE. Today chunk names are UUID-scoped so this is unlikely, but consider exposing an override on VolumeBulkWriter or at least documenting the inherited default explicitly.

sdk-bulkwriter/src/main/java/io/milvus/bulkwriter/UploadManifest.java
```diff
@@ -0,0 +1,424 @@
...
+                            && Files.isSameFile(child, directory)) { continue; }
+                    BasicFileAttributes attrs = Files.readAttributes(child, BasicFileAttributes.class,
+                            LinkOption.NOFOLLOW_LINKS);
+                    String relative = root.relativize(child).toString().replace(child.getFileSystem().getSeparator(), "/");
+                    if (attrs.isRegularFile()) {
+                        add(relative, attrs.size(), attrs.lastModifiedTime().toMillis());
+                    } else if (attrs.isDirectory()) {
+                        enqueueDirectory(relative);
+                    } else {
+                        throw new IOException("Symbolic links and special files are not supported: " + child);
```
yhmo: sdk-bulkwriter/src/main/java/io/milvus/bulkwriter/UploadManifest.java line:217
Low ---- The new scan now hard-fails the entire upload when the source tree contains any symbolic link or special file. The previous implementation (FileUtils.processLocalPath/findFilesRecursively) followed symlinks silently. This is a documented behavior change, but existing callers with symlinked subtrees will now get a complete failure instead of an upload. Consider whether a pre-scan listing of unsupported entries or an explicit opt-out is worth adding so a single stray symlink deep in a large tree does not abort everything.

sdk-bulkwriter/pom.xml
```diff
@@ -17,9 +17,15 @@
 
     <properties>
         <skip.maven.deploy>false</skip.maven.deploy>
+        <sqlite-jdbc.version>3.53.4.0</sqlite-jdbc.version>
     </properties>
 
     <dependencies>
+        <dependency>
```
yhmo, line 24: sdk-bulkwriter/pom.xml line:24
Low ---- sqlite-jdbc (a ~12 MB jar that ships native libraries) is added as a mandatory compile-scope dependency to the published milvus-sdk-java-bulkwriter artifact, although it is only exercised when the request-local index spills past the in-memory thresholds (10,000 files/directories or 8 MiB). Consider declaring it optional or runtime scope so consumers who never exceed those thresholds do not inherit a native library dependency they do not need.

sdk-bulkwriter/src/main/java/io/milvus/bulkwriter/VolumeFileManager.java
yhmo: The user-supplied `ProgressListener` callback now runs while the tracker monitor is held. Both `updateFile` (line 605) and `finishFile` (line 612) are `synchronized` and call `emitIfNeeded` -> `emitProgress` -> `progressListener.onProgress(...)` from inside the lock. In the previous implementation the progress snapshot was built under the lock but the listener was invoked outside of it (`if (progress != null) { emitProgress(progress); }`). Since every concurrent upload worker contends on this tracker for each chunk (`updateFile` is called from `FileUploadProgress.onProgress` on the MinIO transfer threads) and on every file completion, a slow or blocking listener now stalls all in-flight uploads of the request, and the coordinators `finishUpload()` also blocks behind it. Consider collecting the snapshot under the lock and invoking the listener after releasing it, as before.

sdk-bulkwriter/src/main/java/io/milvus/bulkwriter/request/volume/UploadFilesRequest.java
yhmo, line 103: This default flips the existing behavior for every current `VolumeFileManager`/`VolumeBulkWriter` caller: uploads that used to always overwrite now silently skip objects whose size matches, and a same-size but different-content file will never be pushed. The PR documents this, but note that `VolumeBulkWriter` has no escape hatch: `VolumeBulkWriterParam` does not expose `uploadPolicy`, and `VolumeBulkWriter.uploadObject()` builds the `UploadFilesRequest` without setting one (defaults to `SKIP_IF_SAME_SIZE`). A user re-running a BulkWriter job (e.g. re-import after a partial failure) can therefore silently keep stale remote chunks with no way to force `OVERWRITE` through the writer API. Consider adding an `uploadPolicy` passthrough to `VolumeBulkWriterParam`, or at least document the BulkWriter implication in the writer Javadoc.

PR #2097, opened 2026-09-18
Close redundant auto-backport PRs whose content is already on 3.0

yhmo:
Prevents the duplicate-backport edge case where a change is delivered to `3.0` both by the auto-backport workflow and by a manual cherry-pick. Today the auto PR would be squash-merged by Mergify as an **empty commit** (seen on PR #2096). This change closes the redundant PR and deletes its throwaway branch before that happens.

## Changes

All in **`.github/workflows/backport-3.0.yml`** (the sweep is added as a second job in the same file; there is no separate workflow file):

- **Duplicate-prevention at creation time** ("Check for duplicate content on 3.0" step): after cherry-picking, runs `git cherry -v origin/3.0 HEAD` (patch-id comparison, so an equivalent manual cherry-pick matches) and sets `skip=true` when every branch commit is already present on 3.0, gating the sign-off/push/create-PR steps.

- **`sweep` job** (new, same workflow): runs on every push to `3.0`, on a 15-minute schedule, and via `workflow_dispatch`. It lists open PRs into 3.0 carrying the `backport-3.0` label, checks each against `origin/3.0` with `git cherry`, and for redundant ones closes the PR with a comment and deletes the `cherry-pick-*` branch using `CHERRY_PICK_PAT`. PRs carrying the `do-not-merge` escape hatch are skipped. Manual `cherry-pick.yml` PRs (no `backport-3.0` label) are never touched.

- The `backport` job is now gated with `github.event_name == 'pull_request_target'`; the `sweep` job runs for the non-`pull_request_target` triggers. Concurrency stays keyed per merge commit; sweep runs share the empty group and cancel each other, which is fine because a sweep is idempotent and fast.


.github/workflows/backport-3.0.yml
```diff
@@ -238,3 +274,63 @@ jobs:
...
+            NUM=$(printf '%s' "$row" | cut -f1)
+            HEAD_REF=$(printf '%s' "$row" | cut -f2)
+            echo "Checking PR #${NUM} (${HEAD_REF})"
+            if ! git fetch origin "$HEAD_REF" >/dev/null 2>&1; then
+              echo "skip: cannot fetch ${HEAD_REF}"
+              continue
+            fi
+            # '+' lines mean changes not yet on 3.0; if there are none, the backport is
+            # redundant. grep reads all input (no early-exit pipe).
+            if ! git cherry -v origin/3.0 FETCH_HEAD | grep -q '^+'; then
```
yhmo: The inline comment says "grep reads all input (no early-exit pipe)", but `grep -q` does the opposite: it exits as soon as the first match is found and closes the read end of the pipe. Under `set -euo pipefail`, if `git cherry` still has output to write (output larger than the pipe buffer), it dies with SIGPIPE (exit 141) and the pipeline is reported as a failure.

That failure then flows into the surrounding logic in the wrong direction:
- Here: `if ! git cherry ... | grep -q "^+"; then` -> a SIGPIPE'd pipeline (a `+` match that SHOULD keep the PR open) makes `! <non-zero>` true, so the sweep closes a backport PR that still has unreported changes and deletes its branch.
- In the `dupcheck` step (same pattern), a SIGPIPE'd pipeline makes the `if` false, so the step sets `skip=true` and silently skips a backport that is still needed.

This is exactly the failure the workflow avoids everywhere else in the same file (validate step: "a `grep -q` early-exit would be killed by SIGPIPE under pipefail and misreport a match as failure"; cherry-pick step: "capture instead of `git ls-files -u | grep -q .`"). The new steps should follow the same convention, e.g. capture the output and test it (`[ -n "$(git cherry -v origin/3.0 FETCH_HEAD | grep "^+")" ]`), or use `git rev-list --cherry-pick` style comparison. Please also drop the incorrect "no early-exit pipe" comment.

.github/workflows/backport-3.0.yml
```diff
@@ -141,8 +160,25 @@ jobs:
...
+        id: dupcheck
         if: steps.validate.outputs.skip != 'true' && steps.cherrypick.outputs.skip != 'true'
+        run: |
+          set -euo pipefail
+          git fetch origin 3.0
+          # git cherry compares patch-ids, so an equivalent change already on 3.0 (e.g. a
+          # manual cherry-pick merged after this run started) makes every commit of the
+          # branch show as '-' and there is nothing new to backport. Skip instead of opening
+          # a duplicate PR that Mergify would merge as an empty commit.
+          if git cherry -v origin/3.0 HEAD | grep -q '^+'; then
```
yhmo: Same `grep -q`/pipefail hazard as the sweep step below: `grep -q` exits at the first `^+` match and closes the pipe; if `git cherry` is still writing (output larger than the pipe buffer), it dies with SIGPIPE under `set -o pipefail`, the pipeline is reported as non-zero, and this `if` takes the `else` branch -> `skip=true` -> a backport that is still needed is silently skipped. This contradicts the workflow's own convention documented in the validate and cherry-pick steps (avoid `grep -q` under pipefail; capture output instead). Consider `[ -n "$(git cherry -v origin/3.0 HEAD | grep '^+')" ]` so the whole output is consumed before testing.

.github/workflows/backport-3.0.yml
```diff
@@ -238,3 +274,63 @@ jobs:
...
+          GH_TOKEN: ${{ secrets.CHERRY_PICK_PAT }}
+        run: |
+          set -euo pipefail
+          if [ -z "$GH_TOKEN" ]; then
+            echo "::error::secrets.CHERRY_PICK_PAT is not configured; add a PAT with 'repo' scope and store it as CHERRY_PICK_PAT"
+            exit 1
+          fi
+
+          mapfile -t PRS < <(gh pr list --repo "${GITHUB_REPOSITORY}" \
+            --base 3.0 --state open --label backport-3.0 --limit 100 \
```
yhmo, line 312: The sweep selects PRs solely by the `backport-3.0` label and closes them (and deletes their branch) without respecting the `do-not-merge` label, which `.github/mergify.yml` documents as the escape hatch to halt a specific backport. A maintainer who applies `do-not-merge` to a backport PR to inspect or dispute it before the redundant-content logic should run would have that PR force-closed and its throwaway branch deleted by this sweep, with no opportunity to intervene. Since the action is destructive and fully automatic, consider filtering it out of the selection (e.g. `--label backport-3.0 --label "!do-not-merge"`, or `gh pr list --json labels` and skip rows carrying `do-not-merge`) so the documented halt mechanism stays authoritative.

.github/workflows/backport-3.0.yml
```diff
@@ -238,3 +274,63 @@ jobs:
...
+        env:
+          GH_TOKEN: ${{ secrets.CHERRY_PICK_PAT }}
+        run: |
+          set -euo pipefail
+          if [ -z "$GH_TOKEN" ]; then
+            echo "::error::secrets.CHERRY_PICK_PAT is not configured; add a PAT with 'repo' scope and store it as CHERRY_PICK_PAT"
+            exit 1
+          fi
+
+          mapfile -t PRS < <(gh pr list --repo "${GITHUB_REPOSITORY}" \
```
yhmo: .github/workflows/backport-3.0.yml line:300
Medium ---- `mapfile -t PRS < <(gh pr list ...)` swallows the exit status of `gh pr list`: if the command fails (expired CHERRY_PICK_PAT, rate limit, transient network error), the process-substitution status is not propagated to `mapfile`, so `PRS` stays empty and the step prints "No open backport PRs to check." and exits 0, reporting success while the sweep did nothing. Under `set -euo pipefail` this still passes because `mapfile` returns 0. If `gh` fails, the redundant PR is never closed and Mergify squash-merges it as an empty commit — the exact failure this PR is meant to prevent — and the run is green. Consider capturing the listing into a variable and checking the exit code (or `gh pr list ... > /tmp/prs` then test `$?`) so a broken token surfaces as a failed step instead of silently disabling the sweep.

.github/workflows/backport-3.0.yml
```diff
@@ -238,3 +274,63 @@ jobs:
             fi
             echo "Pull request created: $PR_URL"
           fi
+
+  sweep:
```
yhmo, line 284: .github/workflows/backport-3.0.yml line:278
Low ---- The PR description says a new workflow file `.github/workflows/close-redundant-backports.yml` is created for the sweep, but the diff adds the sweep as a second job inside `backport-3.0.yml` (this file). There is no `close-redundant-backports.yml` in the diff. Please align the description (or move the job into its own file) so the change and the docs match.

.github/workflows/backport-3.0.yml
```diff
@@ -24,13 +31,24 @@ on:
   # never checks out or executes the PR head.
   pull_request_target:
     types: [closed]
+  # run the sweep whenever 3.0 moves (e.g. a manual cherry-pick landed) and as a periodic
+  # safety net for pushes that race Mergify's auto-merge
+  push:
```
yhmo, line 39: .github/workflows/backport-3.0.yml line:36
Low ---- For `push` events GitHub runs the workflow file as it exists on the pushed branch, so this `push` trigger only fires once this updated `backport-3.0.yml` (with the `sweep` job) is itself present on `3.0`. Until that backport lands, pushes to 3.0 still execute the old 3.0-branch copy of this file, which has no `push` trigger and no `sweep` job, so the sweep runs only via the 15-minute `schedule` (default-branch version). The schedule covers the interim, but the PR description saying the sweep "runs on every push to 3.0" is only true after this change is backported to 3.0. Consider noting this bootstrap dependency.

.github/workflows/backport-3.0.yml
yhmo: The issue is confirmed fixed. Resolving this conversation.

.github/workflows/backport-3.0.yml
yhmo: The issue is confirmed fixed. Resolving this conversation.

.github/workflows/backport-3.0.yml
yhmo: The issue is confirmed fixed. Resolving this conversation.

.github/workflows/backport-3.0.yml
yhmo: The issue is confirmed fixed. Resolving this conversation.

.github/workflows/backport-3.0.yml
yhmo: The issue is confirmed fixed. Resolving this conversation.

.github/workflows/backport-3.0.yml
yhmo: The issue is confirmed fixed. Resolving this conversation.

.github/workflows/backport-3.0.yml
```diff
@@ -238,3 +280,75 @@ jobs:
...
+          # do-not-merge escape hatch (documented in .github/mergify.yml) so a maintainer
+          # pausing a backport for review is never force-closed by this sweep.
+          PR_LIST=$(gh pr list --repo "${GITHUB_REPOSITORY}" \
+            --base 3.0 --state open --label backport-3.0 --limit 100 \
+            --json number,headRefName,labels \
+            --jq '.[] | select(all(.labels[]?; .name != "do-not-merge")) | [.number, .headRefName] | @tsv') || {
+            echo "::error::gh pr list failed (check CHERRY_PICK_PAT / rate limit)"
+            exit 1
+          }
+          mapfile -t PRS <<< "$PR_LIST"
```
yhmo: .github/workflows/backport-3.0.yml line:318
Low ---- When `gh pr list` returns no matching PRs (the steady state for this queue), `PR_LIST` is empty, but `mapfile -t PRS <<< "$PR_LIST"` still yields one empty array element because a here-string always appends a trailing newline. As a result `[ ${#PRS[@]} -eq 0 ]` never fires, the intended "No open backport PRs to check." early-exit is dead code, and every scheduled run iterates once with `NUM=""`/`HEAD_REF=""`: it executes a spurious `git fetch origin ""` (which fetches the remote default branch into FETCH_HEAD rather than failing) and then logs misleading `Checking PR # ()` / `PR # () still has changes not on 3.0; keeping it open` lines. Consider guarding on the string instead, e.g. `if [ -z "$PR_LIST" ]; then echo "No open backport PRs to check."; exit 0; fi` before the `mapfile` call.

.github/workflows/backport-3.0.yml
```diff
@@ -238,3 +280,75 @@ jobs:
...
+              echo "::notice::PR #${NUM} changes are already on 3.0; closing and deleting ${HEAD_REF}"
+              if gh pr close "$NUM" --repo "${GITHUB_REPOSITORY}" \
+                  --comment "This backport is already present on 3.0 (delivered by another cherry-pick). Closing and deleting its branch."; then
+                closed=$((closed+1))
+              else
+                echo "::warning::failed to close PR #${NUM}"
+                continue
+              fi
+              # push with the tokenized URL inline so the PAT is never written into .git/config
+              git push "https://x-access-token:${GH_TOKEN}@github.com/${GITHUB_REPOSITORY}.git" --delete "$HEAD_REF" \
```
yhmo: .github/workflows/backport-3.0.yml line:348
Low ---- The sweep force-deletes the PR head branch with `git push ... --delete "$HEAD_REF"` based only on the `backport-3.0` label, without validating that `HEAD_REF` matches the throwaway `cherry-pick-*to-3.0-*` pattern the workflow itself creates. In the normal flow the label is only applied by the backport job to its own branches, but the label is also manually assignable and the same branch can back a second open PR, so a close here would silently delete a branch the author did not intend to be removed. Consider gating the deletion on the branch-name pattern (e.g. `case "$HEAD_REF" in cherry-pick-*) ... ;; *) warn and skip ;; esac`); the close itself would still complete for labeled PRs regardless.

.github/workflows/backport-3.0.yml
yhmo: The issue is confirmed fixed. Resolving this conversation.

.github/workflows/backport-3.0.yml
yhmo: The issue is confirmed fixed. Resolving this conversation.

PR #2100, opened 2026-09-21
doc: tag Zilliz Cloud only APIs with @zillizCloudOnly javadoc tag

yhmo:
Add a custom @zillizCloudOnly javadoc tag (registered in the release profile of the root pom) and apply it to all Zilliz Cloud specific classes, methods and constants so generated docs can distinguish cloud-only surface from the general-purpose API.

Marked in sdk-core: globalcluster package, MilvusClientV2Session, MilvusClientV2.session() and CLOUD_SERVERLESS_URI_REGEX. Marked in sdk-bulkwriter: volume manager/file manager/bulk writer and their params, cloud volume/import request DTOs, responses, models and the DataVolumeUtils REST helper.


pom.xml
```diff
@@ -203,6 +203,13 @@
                         <version>${maven.javadoc.plugin.version}</version>
                         <configuration>
                             <additionalJOption>-Xdoclint:none</additionalJOption>
+                            <tags>
```
yhmo, line 206: pom.xml line:206
Medium ---- The @zillizCloudOnly tag is registered only inside the `release` profile of the root pom, but the docs site published to milvus-io.github.io is generated by `.github/workflows/generate_javadoc.yml` via `mvn javadoc:javadoc` without `-Prelease`. In that build the tag is not registered, so javadoc emits "unknown tag: zillizCloudOnly" warnings and drops the "Zilliz Cloud only" marker from the published site docs, defeating the PR's goal for the website docs (only the Maven Central javadoc jar built with `-Prelease` in publish-release.yaml will render it). Consider registering the tag in the default (non-profile) javadoc plugin configuration, or activating the release profile in the generate_javadoc workflow.

yhmo: Accepting as-is. The tag is registered in the `release` profile, which is the profile used by the Maven Central release build (`publish-release.yaml` runs `mvn -Prelease clean deploy`), so the shipped javadoc jar renders the "Zilliz Cloud only" section correctly. The github.io site javadoc (generated without `-Prelease`) only drops the structured tag section; the leading "Zilliz Cloud only." sentence in each javadoc description remains rendered as plain text, and the tag stays fully present in the source comments for future doc tooling. The unknown-tag output is a warning, so the site build remains green.

